PR #224, opened 2021-05-17
initial library for wasm host env integration testing + wasm testing

tzemanovic:
This PR adds a root-level crate "tests" that is used for:
- integration testing of the host env functions between wasm imports and node's host environment
- testing library for tests in wasm

The updated dependency diagram: https://github.com/heliaxdev/anoma-prototype/blob/tomas/host-env-tests/tech-specs/src/explore/design/overview.md

I had to change a few things for this:
- Removed .cargo/config from wasms, because it was also being applied to test builds. The tests have to be built for native target. Instead, the build target is set in wasms' Makefiles
- Renamed all the functions imported to wasms to globally unique names (this is due to a limitation in a tests macro, that's used to setup testing environment, which requires that unique symbols are applied to it)
- Added `"testing"` feature to the main anoma lib, which is used to expose testing helpers and to make some code public to the outside for integration testing
- Refactored Makefile to also run fmt/clippy/tests in wasm directories
- Separate testing helper code from anoma crate, because depending directly on it from wasms bloats the build a lot
- switch to `resolver = "2"` in all the crates (this is to avoid feature unification in resolver v1 between dependencies and dev-dependencies, which use different features)
- Decouple the matchmaker from the VM host_env - the matchmaker is now being called via an implementer of `MmHost` trait
- Move the whole VM module into the shared crate, the wasm runtime is optional (can be switched on using a cargo feature)
- Move the storage trait into the shared crate so it can be used for testing. I had to parametrize the storage hasher used in the merkle tree, because blake2b has C deps, so it cannot be compiled to wasm
- and some cosmetic renames and moves of modules

Left for follow-ups, because this PR is too big already:
- implement VP `eval` runner for the tests
- refactor matchmaker's VM host reference
- write more host env integration and VP/tx tests

Let's add a similar library code for matchmaker and filter after we write some integration tests for host env and VP/tx tests to see how well this works.


tzemanovic:
Sorry for the large PR, it's been in works for a while and it kinda snowballed changes over time and, unfortunately, some module moves don't show :/

@sribst pls review the changes relevant to the matchmaker

@yito88 if you could pls review everything else - the main change is the new `tests` crate

apps/src/lib/node/gossip/intent_gossiper/matchmaker.rs
```diff
@@ -106,9 +135,22 @@ impl Matchmaker {
...
     pub async fn handle_mm_message(&mut self, mm_message: MatchmakerMessage) {
         match mm_message {
-            MatchmakerMessage::InjectTx(tx) => {
+            MatchmakerMessage::InjectTx(tx_data) => {
+                let tx_code = self.tx_code.clone();
+                let keypair = wallet::matchmaker_keypair();
+                let signed = SignedTxData::new(&keypair, tx_data, &tx_code);
+                let signed_bytes = signed
+                    .try_to_vec()
+                    .expect("Couldn't encoded signed matchmaker tx data");
```
sribst, line 144:
```suggestion
                    .expect("Couldn't encode signed matchmaker tx data");
```

apps/src/lib/types/mod.rs
```diff
@@ -1,10 +1,8 @@
 use std::collections::HashSet;
 
-use crate::proto::types;
-
 #[derive(Debug)]
 pub enum MatchmakerMessage {
-    InjectTx(types::Tx),
+    InjectTx(Vec<u8>),
```
sribst, line 5: Maybe this message could be rename to `CraftAndInjectTx`, but it might be a bit too long ?

sribst (commented):
I've only reviewed matchmaker change, and LGTM

yito88 (approved):
Great tests! LGTM 👍

apps/src/lib/types/mod.rs
tzemanovic: Yeah, I think that's a bit wordy and I'm not sure if it actually makes it clearer. Let's give a more detailed description in docstrings (#258).

PR #252, opened 2021-06-01
Protobuf validation

yito88:
adding validation layer for protobuf types

### DIFF
--- a/apps/src/lib/proto/types.rs
+++ b/apps/src/lib/proto/types.rs
@@ -1,0 +1,329 @@
+use std::collections::hash_map::DefaultHasher;
+use std::hash::{Hash, Hasher};
+
+use prost::Message;
+use thiserror::Error;
+
+use super::generated::{services, types};
+
+#[derive(Error, Debug)]
+pub enum Error {
+    #[error("Error decoding a transaction from bytes: {0}")]
+    TxDecodingError(prost::DecodeError),
+    #[error("Error decoding an IntentGossipMessage from bytes: {0}")]
+    IntentDecodingError(prost::DecodeError),
+    #[error("Error decoding an DkgGossipMessage from bytes: {0}")]
+    DkgDecodingError(prost::DecodeError),
+    #[error("Intent is empty")]
+    NoIntentError,
+    #[error("Dkg is empty")]
+    NoDkgError,
+}
+
+pub type Result<T> = std::result::Result<T, Error>;
+
+#[derive(Debug)]
+pub struct Tx {
+    inner: types::Tx,
+}
+
+impl Tx {
+    pub fn from(tx_bytes: impl AsRef<[u8]>) -> Result<Self> {
+        let inner = types::Tx::decode(tx_bytes.as_ref())
+            .map_err(Error::TxDecodingError)?;
+        Ok(Tx { inner })
+    }
+
+    pub fn new(code: Vec<u8>, data: Option<Vec<u8>>) -> Self {
+        let inner = types::Tx {
+            code,
+            data,
+            timestamp: Some(std::time::SystemTime::now().into()),
+        };
+        Tx { inner }
+    }
+
+    pub fn to_bytes(&self) -> Vec<u8> {
+        let mut bytes = vec![];
+        self.inner
+            .encode(&mut bytes)
+            .expect("encoding a transaction failed");
+        bytes
+    }
+
+    pub fn code(&self) -> Vec<u8> {
+        self.inner.code.clone()
+    }
+
+    pub fn data(&self) -> Vec<u8> {
+        match &self.inner.data {
+            Some(d) => d.clone(),
+            None => Vec::new(),
+        }
+    }
+
+    pub fn timestamp(&self) -> std::time::SystemTime {
+        match &self.inner.timestamp {
+            Some(t) => t.clone().into(),
+            None => std::time::SystemTime::now(),
+        }
+    }
+}
+
+pub struct IntentGossipMessage {
+    inner: types::IntentGossipMessage,
+}
+
+impl IntentGossipMessage {
+    pub fn from(intent_bytes: impl AsRef<[u8]>) -> Result<Self> {
+        let inner = types::IntentGossipMessage::decode(intent_bytes.as_ref())
+            .map_err(Error::IntentDecodingError)?;
+        match &inner.msg {
+            Some(_) => Ok(IntentGossipMessage { inner }),
+            None => Err(Error::NoIntentError),
+        }
+    }
+
+    pub fn new(intent: &Intent) -> Self {
+        let inner = types::IntentGossipMessage {
+            msg: Some(types::intent_gossip_message::Msg::Intent(
+                intent.convert(),
+            )),
+        };
+        IntentGossipMessage { inner }
+    }
+
+    pub fn to_bytes(&self) -> Vec<u8> {
+        let mut bytes = vec![];
+        self.inner
+            .encode(&mut bytes)
+            .expect("encoding an intent failed");
+        bytes
+    }
+
+    pub fn intent(&self) -> Intent {
+        match &self.inner.msg {
+            Some(types::intent_gossip_message::Msg::Intent(i)) => {
+                Intent::from(i.clone())
+            }
+            _ => unreachable!(),
+        }
+    }
+}
+
+#[allow(dead_code)]
+pub struct DkgGossipMessage {
+    inner: types::DkgGossipMessage,
+}
+
+#[allow(dead_code)]
+impl DkgGossipMessage {
+    pub fn from(dkg_bytes: impl AsRef<[u8]>) -> Result<Self> {
+        let inner = types::DkgGossipMessage::decode(dkg_bytes.as_ref())
+            .map_err(Error::DkgDecodingError)?;
+        match &inner.dkg_message {
+            Some(_) => Ok(DkgGossipMessage { inner }),
+            None => Err(Error::NoDkgError),
+        }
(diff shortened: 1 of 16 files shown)


apps/src/lib/proto/types.rs
```diff
@@ -0,0 +1,329 @@
...
+    pub fn code(&self) -> Vec<u8> {
+        self.inner.code.clone()
+    }
+
+    pub fn data(&self) -> Vec<u8> {
+        match &self.inner.data {
+            Some(d) => d.clone(),
+            None => Vec::new(),
+        }
+    }
```
sribst, line 63: I think it's a bit strange to returns an empty vec when there is no data, it might be better instead to return `Option<Vec<u8>>`

When using this function it should be checkable if the tx has data or not

apps/src/lib/proto/types.rs
```diff
@@ -0,0 +1,329 @@
...
+pub struct Tx {
+    inner: types::Tx,
+}
+
+impl Tx {
+    pub fn from(tx_bytes: impl AsRef<[u8]>) -> Result<Self> {
+        let inner = types::Tx::decode(tx_bytes.as_ref())
+            .map_err(Error::TxDecodingError)?;
+        Ok(Tx { inner })
+    }
```
sribst, line 35: I'm actually not sure, but I think the classic way of implementing this trait is : 

```rust
impl From<Into<Vec<u8>>> for Result<Tx> {
      fn from(tx_bytes: impl AsRef<[u8]>) ->Result<Self> {
              let inner = types::Tx::decode(tx_bytes.as_ref())
                  .map_err(Error::TxDecodingError)?;
              Ok(Tx { inner })
}
impl Tx {
...
```

apps/src/lib/proto/types.rs
```diff
@@ -0,0 +1,329 @@
...
+        Tx { inner }
+    }
+
+    pub fn to_bytes(&self) -> Vec<u8> {
+        let mut bytes = vec![];
+        self.inner
+            .encode(&mut bytes)
+            .expect("encoding a transaction failed");
+        bytes
+    }
```
sribst, line 52: Same idea for these one :
```rust
impl ToBytes for Tx {
     fn to_bytes(self) -> Bytes {
        let mut bytes = vec![];
        self.inner
            .encode(&mut bytes)
            .expect("encoding a transaction failed");
        bytes
        }
}
```

apps/src/lib/proto/types.rs
```diff
@@ -0,0 +1,329 @@
...
+
+impl IntentGossipMessage {
+    pub fn from(intent_bytes: impl AsRef<[u8]>) -> Result<Self> {
+        let inner = types::IntentGossipMessage::decode(intent_bytes.as_ref())
+            .map_err(Error::IntentDecodingError)?;
+        match &inner.msg {
+            Some(_) => Ok(IntentGossipMessage { inner }),
+            None => Err(Error::NoIntentError),
+        }
+    }
```
sribst, line 86: Please refer to previous comment

apps/src/lib/proto/types.rs
```diff
@@ -0,0 +1,329 @@
...
+        IntentGossipMessage { inner }
+    }
+
+    pub fn to_bytes(&self) -> Vec<u8> {
+        let mut bytes = vec![];
+        self.inner
+            .encode(&mut bytes)
+            .expect("encoding an intent failed");
+        bytes
+    }
```
sribst, line 103: Please refer to previous comment

apps/src/lib/proto/types.rs
```diff
@@ -0,0 +1,329 @@
...
+#[allow(dead_code)]
+impl DkgGossipMessage {
+    pub fn from(dkg_bytes: impl AsRef<[u8]>) -> Result<Self> {
+        let inner = types::DkgGossipMessage::decode(dkg_bytes.as_ref())
+            .map_err(Error::DkgDecodingError)?;
+        match &inner.dkg_message {
+            Some(_) => Ok(DkgGossipMessage { inner }),
+            None => Err(Error::NoDkgError),
+        }
+    }
```
sribst, line 130: Please refer to previous comment

apps/src/lib/proto/types.rs
```diff
@@ -0,0 +1,329 @@
...
+        DkgGossipMessage { inner }
+    }
+
+    pub fn to_bytes(&self) -> Vec<u8> {
+        let mut bytes = vec![];
+        self.inner
+            .encode(&mut bytes)
+            .expect("encoding a DKG failed");
+        bytes
+    }
```
sribst, line 146: Please refer to previous comment

apps/src/lib/proto/types.rs
```diff
@@ -0,0 +1,329 @@
...
+            message: Some(services::rpc_message::Message::Topic(
+                message.convert(),
+            )),
+        };
+        RpcMessage { inner }
+    }
+
+    pub fn convert(&self) -> services::RpcMessage {
+        self.inner.clone()
+    }
```
sribst, line 185: Please refer to previous comment

I think this correspond to the `Into` trait

apps/src/lib/proto/types.rs
```diff
@@ -0,0 +1,329 @@
...
+        Intent::from(self.inner.intent.clone().expect("no intent"))
+    }
+
+    pub fn topic(&self) -> String {
+        self.inner.topic.clone()
+    }
+
+    pub fn convert(&self) -> services::IntentMessage {
+        self.inner.clone()
+    }
```
sribst, line 220: Please refer to previous comment

apps/src/lib/proto/types.rs
```diff
@@ -0,0 +1,329 @@
...
+    #[error("Dkg is empty")]
+    NoDkgError,
+}
+
+pub type Result<T> = std::result::Result<T, Error>;
+
+#[derive(Debug)]
+pub struct Tx {
+    inner: types::Tx,
+}
```
sribst, line 28: As it's only a wrapper you could use instead : 
```rust
pub struct Tx (types::Tx);
```
I'm not sure if it's preferable, both are acceptable I feel

sribst (changes requested):
It's nice to add a wrapper, test are good

nit picking code, I'm actually not sure if my suggestion are the way to go

apps/src/lib/proto/types.rs
tzemanovic: I know there's a some of repetition involved in this, but I'd be in favor of actually writing the types here by hand, because proto3 puts optional on many fields that are actually not optional. E.g. for the `Tx`:

```rust
pub struct Tx {
    pub code: Vec<u8>,
    pub data: Option<Vec<u8>>,
    pub timestamp: Timestamp,
}
```

The `data` is declared as optional, but `timestamp` should be required in a valid `Tx`. The validation layer would then make the data nicer for consumers of these types

apps/src/lib/proto/types.rs
tzemanovic: It should be `TryFrom<&[u8]>` as `From` conversions must not fail

apps/src/lib/proto/types.rs
sribst: I aggree with @tzemanovic, It's better to hide protobuf option

apps/src/lib/proto/types.rs
tzemanovic: I don't think there is such a trait?

sribst: oh, you're right It's part of the `bytes` crates that we don't use. I looked online a bit too quickly

apps/src/lib/proto/types.rs
yito88: That makes sense! Thank you!

apps/src/lib/proto/types.rs
yito88: Thank you for your suggestion! I used `TryFrom`.

apps/src/lib/proto/types.rs
yito88: Thank you! I set fields for each type.

apps/src/lib/proto/types.rs
```diff
@@ -0,0 +1,457 @@
...
+
+impl TryFrom<&[u8]> for Tx {
+    type Error = Error;
+
+    fn try_from(tx_bytes: &[u8]) -> Result<Self> {
+        let tx = types::Tx::decode(tx_bytes).map_err(Error::TxDecodingError)?;
+        let timestamp = match tx.timestamp {
+            Some(t) => t,
+            None => std::time::SystemTime::now().into(),
```
tzemanovic, line 41: Because the `timestamp` is required in a valid transaction, I think we should return error when the `timestamp` is missing instead

apps/src/lib/proto/types.rs
```diff
@@ -0,0 +1,457 @@
...
+pub struct Intent {
+    pub data: Vec<u8>,
+    pub timestamp: Timestamp,
+}
+
+impl From<types::Intent> for Intent {
+    fn from(intent: types::Intent) -> Self {
+        let timestamp = match intent.timestamp {
+            Some(t) => t,
+            None => std::time::SystemTime::now().into(),
```
tzemanovic, line 284: I think same in here as for `Tx`, the timestamp should be required. Is that right, @sribst ?

tzemanovic:
thank you for making the changes @yito88! Besides the two comments on `timestamp` handling, this LGTM!

tzemanovic (approved):

sribst (approved):
All good for me

Issue #258, opened 2021-06-02
add missing rust docstrings


tzemanovic:
set `#![warn(missing_docs)]` and add any missing doc strings in crates:

* [x] shared
* [ ] vm_env + vm_macro
* [ ] apps
* [ ] tests

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201757896515792/1201763001026769) by [Unito](https://www.unito.io)

PR #260, opened 2021-06-02
shared crate update docstrings

tzemanovic:
the first point for #258 

depends on #256  (currently included here)


yito88 (approved):
Looks good! Thank you!

PR #263, opened 2021-06-02
spec updates

tzemanovic:
Besides the spec changes, these small fixes are included:
- gas is using `PARALLEL_GAS_DIVIDER` instead of `PARALLEL_GAS_MULTIPLIER` to avoid decimal
- gas `add_base_transaction_fee` only charges a constant `BASE_TRANSACTION_FEE` without `bytes length * TX_GAS_PER_BYTE`. This is because changes in the size of `Tx`'s sub-fields will affect gas differently. We charge `COMPILE_GAS_PER_BYTE` for `tx.code` size and memory gas costs for `tx.data` which is written to the WASM memory as input.
- couple additionally nice to have fields we were missing in `tech-specs/book.toml` config

The ledger spec page (and the spec in general) still have more detail to be added, but let's pls review and merge as is and continue in follow-ups.


tech-specs/src/specs/encoding.md
```diff
@@ -0,0 +1,15 @@
+# Encoding
+
+All the data fields are REQUIRED, unless specified otherwise.
+
+## The ledger
+
+### Transactions
+
+Transactions MUST be encoded using [proto3](https://developers.google.com/protocol-buffers/docs/reference/proto3-spec) in the format as defined for [message Tx](https://github.com/anomanetwork/anoma/blob/master/proto/types.proto).
```
cwgoes, line 9: Can we inline the `.proto` syntax, perhaps?

tech-specs/src/specs/README.md
```diff
@@ -1,3 +1,7 @@
 # Specifications
 
-TODO
+Anoma is a sovereign, proof-of-stake blockchain protocol that enables private, asset-agnostic cash and private bartering among any number of parties.
```
cwgoes, line 3: Link to the whitepaper, perhaps?

tech-specs/src/specs/ledger.md
```diff
@@ -0,0 +1,190 @@
...
+### Transactions
+
+A transaction [encoded with proto3](./encoding.md#transactions) received from ABCI `DeliverTx` method is executed in two main steps:
+
+1. [Transaction execution](#transaction-execution)
+1. [Validity predicates check](#validity-predicates-check)
+
+#### Transaction execution
+
+For any error encountered in any of the following steps of transaction execution, the protocol MUST charge the gas used by the transaction and discard any storage changes that the transaction attempted to perform.
```
cwgoes, line 16: except for gas charged and nonce incremented (of some sort); these storage changes must always happen

cwgoes: we need to increment the nonce so the transaction cannot be replayed

tech-specs/src/specs/ledger.md
```diff
@@ -0,0 +1,190 @@
...
+
+1. Charge a base transaction [gas](#gas): \\[ \verb|BASE_TRANSACTION_FEE| \\]
+1. Decode the transaction bytes and validate the data.
```
cwgoes, line 19: What kind of data validation is performed?

tech-specs/src/specs/ledger.md
```diff
@@ -0,0 +1,190 @@
...
+1. Charge WASM compilation gas, proportional to the bytes `length` of the `code` field of the transaction (this is because the WASM code is compiled with a single-pass compiler): \\[ \verb|length| * \verb|COMPILE_GAS_PER_BYTE| \\].
+1. [Validate the WASM code](#wasm-validation) from the `code` field of the transaction.
+1. Inject a [gas counter](#gas) into the `code`.
+1. Inject a [stack height](#stack-height-limiter) limiter into the `code`.
+1. Compile the transaction `code` with single-pass compiler. The compilation computational complexity MUST be linear in proportion to the `code` size.
```
cwgoes, line 24: Can we link to something describing the single-pass compiler?

tech-specs/src/specs/ledger.md
```diff
@@ -0,0 +1,190 @@
...
+1. Initialize the WASM linear memory with descriptor having the initial memory size equal to [`TX_MEMORY_INIT_PAGES`](#wasm-constants) and maximum memory size to [`TX_MEMORY_MAX_PAGES`](#wasm-constants).
+1. Instantiate the WASM module with imported [transaction host environment functions](#transaction-host-environment-functions) and the instantiated WASM memory.
+1. Write the transaction's `data` into the memory exported from the WASM module instance.
+1. Attempt to call the module's entrypoint function. The entrypoint MUST have signature:
```
cwgoes, line 28:
```suggestion
1. Attempt to call the newly compiled module's entrypoint function. The entrypoint MUST have signature:
```

tech-specs/src/specs/ledger.md
```diff
@@ -0,0 +1,190 @@
...
+   ```wat
+   func (param i64 i64)
+   ```
+   The first argument is the offset to the `data` input written into the memory and the second argument is its bytes length.
+
+If the transaction executed successfully, it is followed [Validity predicates check](#validity-predicates-check).
```
cwgoes, line 34:
```suggestion
If the transaction executed successfully, it is followed by the [validity predicates check](#validity-predicates-check).
```

cwgoes: worth noting that this phase can write state changes

tech-specs/src/specs/ledger.md
```diff
@@ -0,0 +1,190 @@
...
+
+#### Validity predicates check
+
+For the transaction to be valid, all the triggered validity predicates must accept it.
+
+First, the addresses whose validity predicates should be triggered by the transaction are determined. In this process, the addresses get associated with a set of modified storage keys that are relevant to the address:
+1. The addresses set by the transaction (see `insert_verifier` in [transaction host environment functions](#transaction-host-environment-functions)) are associated with *all* the modified storage keys.
+
+   TODO - <https://github.com/anomanetwork/anoma/issues/292>
+1. The storage keys that were modified by the transaction are associated with the addresses included in the storage keys. Note that a storage key may contain more than one address, in which case all its addresses are associated with this key.
```
cwgoes, line 44: did we decide to allow a storage key to be associated with more than one address? I thought we decided to make it possible for the executing code to trigger additional validity predicates

tech-specs/src/specs/ledger.md
```diff
@@ -0,0 +1,190 @@
...
+
+   TODO - <https://github.com/anomanetwork/anoma/issues/292>
+1. The storage keys that were modified by the transaction are associated with the addresses included in the storage keys. Note that a storage key may contain more than one address, in which case all its addresses are associated with this key. 
+1. All these addresses are additionally associated with the storage key to the validity predicates of any newly initialized accounts' by the transaction (see `init_account` in [transaction host environment functions](#transaction-host-environment-functions)).
+
+For all these addresses, attempt to read their validity predicate WASM code from the storage. For each validity predicate look-up, charge storage read gas and WASM compilation gas, proportional to the bytes length of the validity predicate. If any of the validity predicates look-ups fails, or any validity rejects the transaction or fails anywhere in the execution, the whole transaction is rejected. If the transaction is rejected, the protocol MUST charge the gas used by the transaction and discard any storage changes that the transaction attempted to perform.
```
cwgoes, line 47: Should this be a step-by-step list of operations?

tech-specs/src/specs/ledger.md
```diff
@@ -0,0 +1,190 @@
...
+| `VP_MEMORY_INIT_PAGES`               | number of `PAGE`s |   100 |
+| `VP_MEMORY_MAX_PAGES`                | number of `PAGE`s |   200 |
+| `WASM_STACK_LIMIT`                   | stack depth       | 65535 |
+
+
+The WASM instantiation, the types, instructions, validation and execution of WASM modules MUST conform to the [WebAssembly specification](https://webassembly.github.io/spec/core/intro/index.html).
+
+##### WASM validation
+
+The WebAssembly code is REQUIRED to only use deterministic instructions. Furthermore, it MUST NOT use features from any of the following WebAssembly proposals:
```
cwgoes, line 102: and presumably we validate this?

tech-specs/src/specs/ledger.md
```diff
@@ -0,0 +1,190 @@
...
+cargo test test_vp_stack_limiter
+-->
+
+##### WASM memory
+
+- TODO memory read/write gas costs
+
+##### Transaction host environment functions
+
+The following functions from the host ledger are made available in transaction's WASM code. They MAY be imported in the WASM module as shown bellow and MUST be provided by the ledger's WASM runtime:
```
cwgoes, line 130: Can we explain a bit about what each of these functions do?

tech-specs/src/specs/ledger.md
```diff
@@ -0,0 +1,190 @@
...
+```wat
+(export "memory" (memory 0))
+```
+
+- `anoma_tx_init_account` TODO newly created accounts' validity predicates aren't used until the block is committed (i.e. only the transaction that created the account may write into its storage in the block in which its being applied).
+- TODO describe functions in detail
+
+##### Validity predicate host environment functions
+
+The following functions from the host ledger are made available in validity predicate's WASM code. They MAY be imported in the WASM module as shown bellow and MUST be provided by the ledger's WASM runtime.
```
cwgoes, line 161: Can we explain a bit about what each of these functions do?

tech-specs/src/specs/overview.md
```diff
@@ -0,0 +1,20 @@
+# Overview
+
+At high level, Anoma is composed of two main components: distributed ledger and intent gossip with matching system. While they are designed to complement each other, they can be operated separately.
```
cwgoes, line 3:
```suggestion
At a high level, Anoma is composed of two main components: the distributed ledger and the intent gossip / matchmaking system. While they are designed to complement each other, they can be operated separately.
```

tech-specs/src/specs/overview.md
```diff
@@ -0,0 +1,20 @@
...
+
+## The ledger
+
+The ledger is a distributed state machine, relying on functionality provided by [Tendermint](https://docs.tendermint.com/master/spec/) such as its BFT consensus algorithm with instant finality, P2P networking capabilities, transaction mempool and more. The ledger state machine is built on top the [ABCI](https://docs.tendermint.com/master/spec/abci/).
+
+For proposer selection, the ledger employs a proof-of-stake system.
```
cwgoes, line 9: well, yes, but mostly just for validator voting power assignment

cwgoes (approved):
ACK but see comments

tech-specs/src/specs/encoding.md
tzemanovic: good idea! I've included the `.proto` file in the spec from source

tech-specs/src/specs/README.md
tzemanovic: added

tech-specs/src/specs/ledger.md
tzemanovic: Currently, we only require that the `timestamp` is specified. Updated

tech-specs/src/specs/ledger.md
tzemanovic: The best I could find was this blog post describing it https://medium.com/wasmer/a-webassembly-compiler-tale-9ef37aa3b537
I've added the link

tech-specs/src/specs/ledger.md
tzemanovic: yup, the tx and VP sections link to this with `1. [Validate the WASM code](#wasm-validation) from the `code` field of the transaction.` and `1. [Validate the WASM code](#wasm-validation) of the validity predicate.`, resp.

tech-specs/src/specs/ledger.md
tzemanovic: Sure, I'll describe these in detail in a follow up. (left a `TODO` placeholder)

tech-specs/src/specs/overview.md
tzemanovic: Good point. I've corrected this, thank you!

tech-specs/src/specs/ledger.md
tzemanovic: same as below, left a TODO placeholder

tech-specs/src/specs/ledger.md
tzemanovic: I agree, but I'm not sure what would be best fitting approach for us, so I'd like to discuss a bit more about the possible ways we could add replay protection first. At a high level, we can either implement it in the ledger or in VPs. 

If it's in the ledger, then we need to decide how is the nonce/counter associated with an account as our txs don't have a single source. Maybe this could be in the DKG wrapper, which will need a source source for gas anyway, but then we also need to link this in the decrypted tx somehow.

tzemanovic: As discussed last week, we will include replay protection in the DKG wrapper. I left a note in the spec to describe this further.

PR #269, opened 2021-06-04
VM host env integration tests

tzemanovic:
Fixes applied for issues found in tests:
- `tx::init_account` result wasn't memory safe: The returned `Address` has a fixed size, so we can use that to pre-allocate the result
- tx and vp `read` and `prefix_iter` results weren't memory safe: In this case we don't know the byte size up-front, so we need to read them in 2-steps. The first step reads the value into a read cache and returns the size back to wasm, the second step reads the value from cache into a pre-allocated buffer with the obtained size.
- `tx::delete` has an unused generic type parameter

And in test helpers:
- The mock DB prefix iterator wasn't matching the logic of rocksdb iterator
- TestStorage chain ID was an of an incorrect size, which caused the the host fn to fill the missing bytes with zeroes

Other changes:
- moved `get_verifiers` from protocol mod to write_log
- renamed `KeySeg::to_string` to `KeySeg::raw` to avoid ambiguity with `Display::to_string`
- allow to use `&str` with `Key::parse`
- added `init_vp_env_from_tx` test helper to run tests on a VP after transaction execution


shared/src/vm/host_env.rs
```diff
@@ -44,6 +44,9 @@ where
     /// The verifiers whose validity predicates should be triggered.
     /// Not thread-safe, assuming single-threaded Tx runner
     pub verifiers: MutEnvHostWrapper<'a, &'a HashSet<Address>>,
+    /// Cache for 2-step reads from host environment.
+    /// Not thread-safe, assuming single-threaded Tx runner
+    pub read_cache: MutEnvHostWrapper<'a, &'a Option<Vec<u8>>>,
```
yito88, line 49: I associate the name `read_cache` with a general "read cache". And, it looks like a buffer.
How about `result_buffer`?

yito88 (commented):
Overall, looks good!
I left a comment about `read_cache`.

shared/src/vm/host_env.rs
tzemanovic: I agree and I wasn't happy with the name either. `result_buffer` sounds right, I'll update this

tzemanovic: updated in c8495f5, thank you!

yito88 (approved):
LGTM!

PR #272, opened 2021-06-07
check in wasm builds

### LINKED ISSUE #154: wasm deterministic builds
With @sribst, we noticed some difference between wasm built on mac vs. Arch linux. We need to ensure that the wasm builds are deterministic (especially for transactions and VPs).

One reason could be different rust version, so we should lock the project to specific version in the `rust-toolchain` file. There has also been a bug in LLVM described in detail in https://dev.to/gnunicorn/hunting-down-a-non-determinism-bug-in-our-rust-wasm-build-4fk1, but that should have already been fixed. We could use a similar technique describe in the blog to check for deterministic build in the CI (compare hash of two build outputs: https://github.com/paritytech/substrate/pull/6597/files).

tzemanovic:
closes #154 

- add built wasm in git
- removed sccache
- moved all wasms into a wasm sub-directory and add a rust-toolchain with minimal components

@Fraccaman Unfortunately, incremental wasm builds are still quite slow (~1:40 minutes on machine). It looks like it's the `Updating crates.io index` step that takes longest, so it would be nice if we could cache this between the runs somehow. There's a request for `cargo build --dependencies-only` (https://github.com/rust-lang/cargo/issues/2644) that would make this easy, but it's not yet available. There are however some workarounds: https://blog.mgattozzi.dev/caching-rust-docker-builds/ and https://github.com/rust-lang/cargo/issues/2644#issuecomment-830872790. Alternatively, we could change the instructions to keep the docker image running and attach to it for rebuilds. What do you think?


Fraccaman:
Multistage docker build are cool, but I need to experiment a little bit with them (the article is a great start). Avoid deleting the docker container is another option (but I would like to avoid it). I found another project called `cargo-chef` which may come handy. Here a blog post: https://www.lpalmieri.com/posts/fast-rust-docker-builds/ . I think `cargo-chef` could offer a better solution than the previous two.

tzemanovic:
> Multistage docker build are cool, but I need to experiment a little bit with them (the article is a great start). Avoid deleting the docker container is another option (but I would like to avoid it). I found another project called `cargo-chef` which may come handy. Here a blog post: https://www.lpalmieri.com/posts/fast-rust-docker-builds/ . I think `cargo-chef` could offer a better solution than the previous two.

Nice, `cargo-chef` looks ideal for this!

Fraccaman (approved):

Fraccaman:
Merge and follow up the work with #275

PR #274, opened 2021-06-08
[spec] use a temporary mermaid patch

tzemanovic:
The spec is still failing the deploy, it's a bug in the mermaid plugin which is incompatible with a newer version of mdbook described here: https://github.com/badboy/mdbook-mermaid/pull/11

I'm changing the installation command to get the plugin from this commit until it's merged and released.

Failing job: https://github.com/anomanetwork/anoma/runs/2771610663?check_suite_focus=true#step:5:711

### DIFF
--- a/tech-specs/Makefile
+++ b/tech-specs/Makefile
@@ -8,6 +8,8 @@ serve:
 
 dev-deps:
 	$(cargo) install mdbook
-	$(cargo) install mdbook-mermaid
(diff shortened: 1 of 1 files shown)


Fraccaman (approved):

Fraccaman:
I think you should run the CI script otherwise drone is going to fail (Makefile has been modified).

PR #278, opened 2021-06-08
add rustfmt to docs CI

tzemanovic:


.github/workflows/gh-pages.yml
```diff
@@ -17,6 +17,8 @@ jobs:
         with:
           toolchain: stable
       - uses: Swatinem/rust-cache@v1
+      - name: Install rustfmt
+        run: rustup component add rustfmt
```
Fraccaman, line 21: I think its better if we let the `actions-rs/toolchain@v1` gh action take care of installing rustfmt. Something like 
```
uses: actions-rs/toolchain@v1
with:
    toolchain: stable
    components: rustfmt
```

tzemanovic: I see, thanks! I've applied a fixup

Fraccaman (approved):

PR #281, opened 2021-06-09
keep invalid tx in cache

### LINKED ISSUE #236: tendermint `DeliverTx.code=1` issue
We've noticed this issue in the playnet.

There were 4 validator nodes running on cloud (ubuntu). Once I connected to them with a local non-validator node (on mac - not sure if that's relevant), any transaction that was invalid (we set `code=1` in ABCI deliverTx response) wasn't getting evicted from my local node's mempool and so it kept gossiping it after it was applied in a block and the transaction was getting indefinitely applied in all following blocks. This would happen no matter to which RPC the transaction was originally submitted to.

I was able to replicate this again on a fresh state in the same setup, but I'm not able to replicate with a local-only network with validator nodes and a non-validator node.

As a workaround for playnet, we've commented out the code that sets `code=1`. I will try to debug this further once we're done with playnet.

tzemanovic:
closes #236 - first commit. I had to update the config in tendermint-rs, hence the dependencies branch switch. This is being upstreamed in https://github.com/informalsystems/tendermint-rs/pull/897.

In the second commit, I've improved the error handling from the Tendermint module. I noticed that when `tendermint::run` failed, the Anoma node was hanging. This is now fixed.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4,11 +4,11 @@ version = 3
 
 [[package]]
 name = "addr2line"
-version = "0.14.1"
+version = "0.15.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "a55f82cfe485775d02112886f4169bde0c5894d75e79ead7eafe7e40a25e45f7"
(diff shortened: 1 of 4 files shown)


yito88 (approved):
Looks good!

PR #283, opened 2021-06-10
IBC integration design doc

yito88:
closes anoma/namada#10

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201757896515792/1201762935748348) by [Unito](https://www.unito.io)


tech-specs/src/explore/design/ibc.md
```diff
@@ -0,0 +1,18 @@
...
+
+TODO The behaviour of the destination ledger
+
+## IBC storage
+IBC storage is an IBC module to make a packet and send it to the destination. A transaction stores an update for another ledger to the IBC storage instead of the storage into which other updates are written. The IBC storage doesn't persist the update into the DB, it makes a packet and send it to the destination ledger.
+
+We have the write log to store updates or deletes before commit. A transaction should be able to store an update for another ledger into the write log. The log of the update should be a different modification type from other updates to the same ledger, e.g. `StorageModification::IbcWrite`. At the end of the transaction execution, when the write log has an update for another ledger, IBC storage can check and send it as a packet.
+
+## IBC validation
+A validation code validates an update to another ledger has been received on the destination ledger. For the performance, the code should be Rust native code. The validation is executed after all transaction executions. We can execute the validation and validity predicates concurrently.
```
yito88, line 16: I'm not sure what the validation code should do.
Is it enough to receive and check the acknowledgment message?

tech-specs/src/explore/design/ibc.md
```diff
@@ -0,0 +1,18 @@
+# IBC integration
+
+IBC allows the ledgers to interact with each other. IBC is a protocol to agree the consensus state and to send/receive packets between ledgers.
+
+## Update to another ledger
+When a ledger executes a transaction including an update to another ledger, it sends a packet and executes the associated validation code like validity predicates to check if the packet has been received. The transaction writes the update to an internal storage subspace called IBC storage. IBC storage as an IBC module makes a packet and send it. After the transaction execution, not only validity predicates but also the IBC validation code associated with the update are executed. If the validation fails, e.g. the destination ledger fails to do the update, the transaction is aborted.
+
+TODO The behaviour of the destination ledger
```
yito88, line 8: Which module on a ledger receives the packet?

yito88: How does the module execute the update? Does it submit a transaction?

tech-specs/src/explore/design/ibc.md
tzemanovic: I think this would call the the verification functions provided in https://github.com/informalsystems/ibc-rs/blob/master/modules/src/ics04_channel/handler/verify.rs

tech-specs/src/explore/design/ibc.md
```diff
@@ -0,0 +1,18 @@
+# IBC integration
+
+IBC allows the ledgers to interact with each other. IBC is a protocol to agree the consensus state and to send/receive packets between ledgers.
```
cwgoes, line 3: Link to the spec here.

tech-specs/src/explore/design/ibc.md
```diff
@@ -0,0 +1,18 @@
...
+
+## Update to another ledger
+When a ledger executes a transaction including an update to another ledger, it sends a packet and executes the associated validation code like validity predicates to check if the packet has been received. The transaction writes the update to an internal storage subspace called IBC storage. IBC storage as an IBC module makes a packet and send it. After the transaction execution, not only validity predicates but also the IBC validation code associated with the update are executed. If the validation fails, e.g. the destination ledger fails to receive the update due to time out, the transaction is aborted.
```
cwgoes, line 6: Sending and receiving packets are different operations; the former writes a proof of packet to the store so it can be validated; the latter verifies the proof of a packet having been sent on the other ledger.

tech-specs/src/explore/design/ibc.md
cwgoes: We need to implement the IBC specification exactly, since that's what other ledgers will implement.

tech-specs/src/explore/design/ibc.md
```diff
@@ -0,0 +1,18 @@
...
+
+TODO The behaviour of the destination ledger
+
+## IBC storage
+IBC storage is an IBC module to make a packet and send it to the destination. A transaction stores an update for another ledger to the IBC storage instead of the storage into which other updates are written. The IBC storage doesn't persist the update into the DB, it makes a packet and send it to the destination ledger.
```
cwgoes, line 11: It's not quite _storage_ updates for the other ledger, but rather messages which are handled in a particular way.

cwgoes (changes requested):
Good start - it's not so necessary to specify IBC itself here; I would focus on specifying the interface exposed by and semantics implemented by the Anoma IBC "validity predicate", which should interface to Informal's handlers in `ibc-rs`.

cwgoes (changes requested):
Take a look at @ndbunner's initial proof of stake validity predicate at https://hackmd.io/1lASgSGlQWahK1CaZOuCbA for an idea of the kind of structure we want here. Let's discuss this next week.

tech-specs/src/explore/design/ibc.md
```diff
@@ -0,0 +1,127 @@
+# IBC integration
+
+[IBC](https://arxiv.org/pdf/2006.15918.pdf) allows a ledger to track another ledger's consensus state using a light client. IBC is a protocol to agree the consensus state and to send/receive packets between ledgers.
+
+## Transaction for IBC
+A requester (IBC relayer or user) who wants to execute IBC operations on a ledger sets IBC packet or message like `MsgCreateAnyClient` as transaction data and submit the following transaction. IBC validity predicate is invoked after this transaction execution to check the IBC operation is validated. The trigger to invoke IBC validity predicate is `insert_verifier()` in IBC-related operation to add an address for IBC `InternalAddress::Ibc`.
```
cwgoes, line 6: Well, as long as state owned by the IBC validity predicate is written, the VP should be triggered, right?

tech-specs/src/explore/design/ibc.md
```diff
@@ -0,0 +1,127 @@
...
+
+The transaction is given an IBC packet or message which specifies what to do. Because the packet or message has been encoded and stored in transaction data and there are some types of messages, it has to check what packet or message is given first and then decodes the message or packet. After the process according to the packet or message, it emits an IBC event.
```
cwgoes, line 8: Hmm not necessarily, the state writes may just happen, right? The tx may need to include extra data (proofs)

tech-specs/src/explore/design/ibc.md
```diff
@@ -0,0 +1,127 @@
...
+          }
+      }
+  }
+  ```
+
+### Make a packet
+IBC-related transaction for some messages or packets makes a packet. For example, when a transaction wants to transfer a token between ledgers, it should make a packet including `FungibleTokenPacketData` to specify the sender, receiver, token, and amount.
+
+### Store a proof and state
+A proof can be stored to the storage on the ledger. Its storage key should be prefixed (e.g. `ibc/`) to protect them from other storage operations. The paths(keys) for Tendermint client are defined by [ICS 24](https://github.com/cosmos/ibc/blob/master/spec/core/ics-024-host-requirements/README.md#path-space).
```
cwgoes, line 60: We don't need to store proofs, we just need to validate them.

cwgoes (changes requested):
Still needs to be rewritten so it's a relation over the prior state and the posterior state.

tech-specs/src/explore/design/ibc.md
yito88: That's correct. `changed_keys` should have IBC-related keys after IBC-related transactions are executed.
I modified that.

tech-specs/src/explore/design/ibc.md
yito88: Thanks, right. I modified that.

tech-specs/src/explore/design/ibc.md
yito88: Thanks, I misunderstood that. I've modified that.

tech-specs/src/explore/design/ibc.md
```diff
@@ -0,0 +1,400 @@
...
+                } else {
+                    // Verify the proof to check the counterpart ledger's state is expected
+                    //   - The state can be inferred from the own channel state change
+                    //   - Use `ibc-rs::ics04_connection::handler::verify::verify_channel_proofs()`
+                }
+            }
+
+            // send a packet
+            if key.segments.contains(&StringSeg("nextSequenceSend".to_owned())) {
+                match get_packet(tx_data) {
```
yito88, line 222: I'm fixing this part.
In this case, a packet shouldn't be in the `tx_data` because it is made in a transaction. And, the packet doesn't have to be persistent.
I think it needs to be stored in a temporary space or a mechanism to pass from the transaction to the VP.

tzemanovic: @yito88 and I talked about this briefly and one way would be to allow a tx to write temporary arbitrary bytes, which could then be passed to the triggered VPs as an additional argument. After the VPs call, we can drop this structure. Perhaps there will be other scenarios in which we could use this.

@cwgoes what do you think?

tech-specs/src/explore/design/ibc.md
```diff
@@ -0,0 +1,408 @@
+# IBC integration
+
+[IBC](https://arxiv.org/pdf/2006.15918.pdf) allows a ledger to track another ledger's consensus state using a light client. IBC is a protocol to agree the consensus state and to send/receive packets between ledgers.
+
+## Transaction for IBC
+A requester (IBC relayer or user) who wants to execute IBC operations on a ledger sets IBC packet or message like `MsgCreateAnyClient` as transaction data and submit the following transaction. IBC validity predicate is invoked after this transaction execution to check the IBC operation is validated. The trigger to invoke IBC validity predicate is changing IBC-related keys (e.g. prefixed with `ibc/`).
+
+The transaction is given an IBC packet or message which specifies what to do. Because the packet or message has been encoded and stored in transaction data and there are some types of messages, it has to check what packet or message is given first and then decodes the message or packet. Then, it can modify the ledger state by writing not only data specified in the transaction but also IBC-related data on the storage sub-space. After the process according to the packet or message, it emits an IBC event.
```
cwgoes, line 8: what does "some types of messages" mean?

cwgoes: also, IBC-related data _is_ data specified in the transaction, no?

tech-specs/src/explore/design/ibc.md
```diff
@@ -0,0 +1,408 @@
...
+
+- Transaction given a message
+  ```rust
+  #[transaction]
+  fn apply_tx(tx_data: Vec<u8>) {
```
cwgoes, line 13: we don't really need messages at all, the WASM can just take the action directly (`create_client`)

tech-specs/src/explore/design/ibc.md
```diff
@@ -0,0 +1,408 @@
...
+                            //   - Use `ibc-rs::ics04_connection::handler::verify::verify_channel_proofs()`
+                        }
+                        _ => return Err(Error::InvalidStateChange("Invalid state change happened")),
+                    }
+                }
+
+                "nextSequenceSend" => {
+                    // "SendPacket"
+                    let packet = get_packet(key, data_from_tx)?;
+
```
cwgoes, line 241: is there a function in `ibc-rs` we can call here?

tech-specs/src/explore/design/ibc.md
```diff
@@ -0,0 +1,408 @@
...
+                    // Assert that the timeout height and timestamp have not passed on the destination ledger
+
+                    // Assert that the commitment has been stored
+                }
+
+                "nextSequenceRecv" => {
+                    // "RecvPacket"
+                    let packet = get_packet(key, data_from_tx)?;
+                    let proofs = get_proofs(key, data_from_tx)?;
+
```
cwgoes, line 261: is there a function in `ibc-rs` we can call here?

tech-specs/src/explore/design/ibc.md
```diff
@@ -0,0 +1,408 @@
...
+                    // Verify the proofs that the counterpart ledger has stored the commitment
+                    //   - Use `ibc-rs::ics04_connection::handler::verify::verify_packet_recv_proofs()`
+                }
+
+                "nextSequenceAck" => {
+                    // "Acknowledgement"
+                    let packet = get_packet(key, data_from_tx)?;
+                    let proofs = get_proofs(key, data_from_tx)?;
+
```
cwgoes, line 280: is there a function in `ibc-rs` we can call here?

tech-specs/src/explore/design/ibc.md
```diff
@@ -0,0 +1,408 @@
...
+                    // check if the commitment is deleted
+                    match check_commitment_state(key) {
+                        StateChange::Deleted => {
+                            // Assert that the packet was actually sent on this channel
+                            //   - Get the stored commitment and compare it with a commitment made from the packet
+
+                            // Check the channel state change
+                            match check_channel_state(channel_id) {
+                                StateChange::Closed => {
+                                    // "Timeout"
```
cwgoes, line 307: is there a function in `ibc-rs` we can call here?

tech-specs/src/explore/design/ibc.md
```diff
@@ -0,0 +1,408 @@
...
+                        }
+                        StateChange::Created | StateChange::Updated => {
+                            // They are checked in other cases
+                        }
+                        _ => return Err(Error::InvalidStateChange("Invalid state change happened")),
+                    }
+                }
+
+                "ports" | "receipts" | "acks" => {
+                    // They are checked in other cases
```
cwgoes, line 335: really? how do we assert that the correct writes are made in conjunction?

cwgoes (changes requested):
Thanks; much clearer, curious on points of the `ibc-rs` integration.

tech-specs/src/explore/design/ibc.md
cwgoes: Hmm, that's alright I suppose, but why can't the VP just check that a commitment to the packet with expected fields was written?

yito88: Thanks, I reconsidered that.
The VP can check the commitment.
I thought that timeout height and timestamp in the packet have to be checked in the VP. But we can get them from `tx_data` because the data should be set in `tx_data`.
I'll fix them.

tech-specs/src/explore/design/ibc.md
yito88: I've fixed that without massages. yes, the IBC-related data is specified by the transaction.

tech-specs/src/explore/design/ibc.md
yito88: I've modified that without messages

tech-specs/src/explore/design/ibc.md
yito88: `ChannelReader`'s functions are called here.

tech-specs/src/explore/design/ibc.md
yito88: `ChannelReader`'s functions and [`verify_packet_recv_proofs()`](https://github.com/informalsystems/ibc-rs/blob/5cf3b6790c45539c5aaadeef6e1af1f51a5f437f/modules/src/ics04_channel/handler/verify.rs#L56) are called

tech-specs/src/explore/design/ibc.md
yito88: `ChannelReader`'s functions and [`verify_packet_acknowledgement_proofs()`](https://github.com/informalsystems/ibc-rs/blob/5cf3b6790c45539c5aaadeef6e1af1f51a5f437f/modules/src/ics04_channel/handler/verify.rs#L101) are called

tech-specs/src/explore/design/ibc.md
yito88: In "Timeout" case. `ChannelReader`'s functions and [`verify_next_sequence_recv()`](https://github.com/informalsystems/ibc-rs/blob/5cf3b6790c45539c5aaadeef6e1af1f51a5f437f/modules/src/ics04_channel/handler/verify.rs#L134) or [`verify_packet_receipt_absence()`](https://github.com/informalsystems/ibc-rs/blob/5cf3b6790c45539c5aaadeef6e1af1f51a5f437f/modules/src/ics04_channel/handler/verify.rs#L165) are called.

tech-specs/src/explore/design/ibc.md
yito88: Ah, they should be validated. Thanks!

tech-specs/src/explore/design/ibc.md
```diff
@@ -0,0 +1,443 @@
...
+                                    //   - For ordering channels, use `ibc-rs::ics04_connection::handler::verify::verify_next_sequence_recv()`
+                                    //   - For not-ordering channels, use `ibc-rs::ics04_connection::handler::verify::verify_packet_receipt_absence()`
+                                }
+                                _ => return Err(Error::InvalidChannel("Invalid channel state")),
+                            }
+                        }
+                        StateChange::Created => {
+                            // "SendPacket"
+                            // Assert that the packet metadata matches the channel and connection information
+                            //   - the port is owend
```
tzemanovic, line 315:
```suggestion
                            //   - the port is owned
```

tzemanovic:
moved to https://github.com/anoma/namada/pull/295

PR #291, opened 2021-06-11
sign the whole transaction

### LINKED ISSUE #266: add `timestamp` to `SignedTxData`
There's a missing field `timestamp` in the signature create from `SignedTxData`.

We should move the generation of proto files with Tonic in `build.rs` from `apps` crate to the `shared` crate to be able to use it there and sign the whole transaction without missing any fields. Furthermore, maybe instead of returning this type directly, we could have a function e.g. `sign_tx(Tx, &Keypair) -> Tx` that would create `SignedTxData` and replace the `tx.data` with it.

yito88:
closes #266


apps/src/lib/proto/generated/types.rs
```diff
@@ -1,47 +1 @@
...
-}
-/// Nested message and enum types in `DkgGossipMessage`.
-pub mod dkg_gossip_message {
-    #[derive(Clone, PartialEq, ::prost::Oneof)]
-    pub enum DkgMessage {
-        #[prost(message, tag = "1")]
-        Dkg(super::Dkg),
-    }
-}
+
```
tzemanovic, line 1: I think this file can be deleted from here

proto/types.proto
```diff
@@ -8,10 +8,14 @@ package types;
 
 message Tx {
   bytes code = 1;
-  optional bytes data = 2;
+  TxData data = 2;
```
tzemanovic, line 11: Is this change necessary? I think it would be good to keep this explicitly `optional`

yito88: I tried building this with `optional`, but `protoc` couldn't build that. In my investigation, Protobuf v3.15 supports `optional`, `prost` which supports it hasn't released yet. I'm not sure why we couldn't build it directly, even though we could do it with `services.proto`.
https://github.com/danburkert/prost/issues/351

yito88:
@tzemanovic Thank you! The changes look good!

shared/src/types/key/ed25519.rs
```diff
@@ -105,34 +106,36 @@ pub fn verify_signature_raw(
...
-    }
+/// Verify that the transaction has been signed by the secret key
+/// counterpart of the given public key.
+pub fn verify_tx_sig(
+    pk: &PublicKey,
+    tx: &Tx,
+    sig: &Signature,
+) -> Result<(), VerifySigError> {
+    let data = tx.to_bytes();
+    verify_signature_raw(pk, &data, sig)
```
yito88, line 138: I tried this verification, but it failed. The verification failed because the Tx had `SignedTxData`, but the signature had been made without `SignedTxData`.

tzemanovic: ah, yes, thank you very much catching and fixing this!

proto/types.proto
tzemanovic: It seems that this might be possible by setting `.protoc_arg("--experimental_allow_proto3_optional");` on the tonic build config. Would you mind to try this pls?

yito88: That works well! Thank you! https://github.com/anomanetwork/anoma/pull/291/commits/2b4100e676ef22df69f213ce0c47ab97c0718ca2

tzemanovic: thank you too!

tzemanovic (approved):

PR #294, opened 2021-06-14
wasm mem limits

### LINKED ISSUE #84: wasm memory limits
Unfortunately, the current memory bounds set for wasm memory only work for the input that is written to memory before we call the wasm module. It doesn't limit memory usage from wasm allocations and wasm host call allocations. I think we need to implement something like this for that <https://github.com/wasmerio/wasmer/blob/af59fcbf221c1792da41c2f0946f6e42b7bda817/examples/tunables_limit_memory.rs>. I'll turn this into an issue

_Originally posted by @tzemanovic in https://github.com/heliaxdev/anoma-prototype/pull/61#discussion_r599774517_

To test the limits, one can e.g. allocate some bytes `let bytes = [0_u8; 2_usize.pow(21)];` inside wasm and inside host calls that write data to wasm memory.

tzemanovic:
closes #84 

I've added tests (in `shared/src/vm/wasm/runner.rs`) for hitting memory limits in tx, VPs and `eval` for which I've started a new root dir: `wasm_for_tests` (previously they were included in `tests/res/wasm`, but unfortunately, I wasn't able to set the root Cargo workspace to ignore this folder, as it's inside workspace member folder `tests`).

As a side note, because I needed a few different transactions and VPs, I found that we can build them from one crate with conditional compilation, which speeds-up the build time a lot (as all the deps are shared). I will apply the same change in `wasm` folder in another PR.

Updated crates diagram at: https://github.com/anomanetwork/anoma/blob/tomas/wasm-mem-limits/tech-specs/src/explore/design/overview.md

Other changes:
- I've update root Makefile to include `make check`
- added a note on how to enable logging in tests in the readme and added its deps it in the shared crate
- updated the stack limiter tests to also check that 1 below the limit passes
- import `use std::collections::HashSet;` in VP prelude, as it's used in its arguments


yito88 (approved):
Looks good!

PR #299, opened 2021-06-22
system tests

### LINKED ISSUE #132: refactor CLI arguments
Having cli parser functions is an improvement, but I we could improve the cli further. Maybe we could have some helper function for building CLI arguments together with their parses (e.g. we known that when we set the arg as required or give it a default value, we can parse with `_req`). I'm not sure if maybe something like this already exists in clap.

_Originally posted by @tzemanovic in https://github.com/heliaxdev/anoma-prototype/pull/129#r613353543_

tzemanovic:
This PR adds some more e2e for the ledger and transactions. 

fixes:
- 36db68433dbd8be54573dc237e23cf42be36a411: Termination signals weren't handled properly, because `std::process:Child::kill` uses `SIGKILL` which cannot be caught with signal hooks. This changes to use `SIGTERM` instead
- fixed missing `--base-dir` arg in anoma
- CLI refactor: #132 also with a nicer structure for sub-commands and typed arguments
- switch off unused `clap_derive` feature from `clap`
- 471dd538affe5e21628091f70579f1b00b5175ed: reduced dependencies logging in gossip 
- 1d1869271874b49fea9e4f2ad1a0f8edb57b501e: increase RPC max body request size, which was failing `--dry-run`s with larger body requests

closes #132


Fraccaman:
I think we can close this on as me merged #305

tzemanovic:
> I think we can close this on since me merged #305

There's some new tests and bug fixes for the ledger. I'm just finishing with additional transaction tests.

apps/src/bin/anoma-node/main.rs
```diff
@@ -5,23 +5,44 @@ use std::env;
...
-        _ => filter.add_directive(abci_level),
-    };
+    let log_env_var = env::var(logging::ENV_KEY);
+
+    // Ledger dependencies
+    let filter = reduce_log_level("tendermint_abci", filter, &log_env_var)?;
+    // Gossip dependencies
+    let filter = reduce_log_level("regalloc", filter, &log_env_var)?;
+    let filter =
+        reduce_log_level("wasmer_compiler_cranelift", filter, &log_env_var)?;
```
Fraccaman, line 23: I think we should also add libp2p which is pretty verboose

Fraccaman (approved):
I think that after adding libp2p custom log we can merge this

PR #300, opened 2021-06-22
desc internal addresses with native VPs

tzemanovic:
related to #47, #283 and #38

PR #303, opened 2021-06-22
native VPs

tzemanovic:
this PR provides an interface for native VPs as described in #300


shared/src/ledger/native_vp.rs
```diff
@@ -0,0 +1,242 @@
...
+where
+    DB: storage::DB + for<'iter> storage::DBIter<'iter>,
+    H: StorageHasher,
+{
+    /// Storage prefix iterators.
+    pub iterators: PrefixIterators<'a, DB>,
+    /// VP gas meter.
+    pub gas_meter: VpGasMeter,
+    /// Read-only access to the storage.
+    pub storage: &'a Storage<DB, H>,
```
yito88, line 67: I think a native VP stores the data for the functionality to the storage.
Should we provide `write` functions?

yito88 (commented):
Thank you! Looks great.
I have a question.

shared/src/ledger/native_vp.rs
tzemanovic: Thank you for reviewing! It is my understanding that any storage modifications should be done via transactions (via the write log) and then the VP checks that these are correct (same as with wasm VPs). For the native VPs integrations, we can add functions to write storage in the `vm_env` crate using the tx host env functions. These can then be called from transactions.

yito88: Sorry, I misunderstood that. As we discussed, the VP doesn't update the storage. So, write access isn't needed.

yito88 (approved):

PR #304, opened 2021-06-23
Chain ID subdir and load storage just after open

### LINKED ISSUE #268: chain ID in storage
We should write the DB into a subdir (has a `TODO` in code in `ledger::Storage::commit()`) and update the loading of state. The chain ID has to be known before we try to load the state. We should set the chain ID in `genesis` module and override the one in tendermint with it (like we do with the validator for "dev" feature with `write_validator_key`).

Also, we shouldn't return this from `DB::read_last_block` in the `BlockState`. Furthermore, we should write the `BlockState` type in `DB::write_block`.

Perhaps we can also load the state from storage earlier, before we receive `GetInfo` from ABCI as we currently do, so if there's a problem with storage, we don't even start-up a tendermint node.

yito88:
closes #268


apps/src/lib/node/ledger/tendermint.rs
```diff
@@ -505,3 +509,28 @@ fn write_validator_key(
...
+    home_dir: impl AsRef<Path>,
+    chain_id: String,
+) -> io::Result<()> {
+    let home_dir = home_dir.as_ref();
+    let path = home_dir.join("config").join("genesis.json");
+
+    // Don't use `TendermintConfig::load_genesis_file()` because tendermint
+    // 0.34.x requires `block.TimeIotaMs`. It doesn't exist in tendermint-rs.
+    // In the master of Tendermint (go), it was also removed at
+    // [#5987](https://github.com/tendermint/tendermint/pull/5987))
```
tzemanovic, line 524: It looks like some other fields are outdated too, but `block.TimeIotaMs` has not been released in Tendermint(go) 0.34.x. I will probably be in 0.35. I'll open an issue for this

tzemanovic: I've pushed a commit to use an updated branch. The other changes we previously upstreamed to tendermint-rs have been released so I made a new branch from their master.

yito88: Thank you! That's pretty cool.

tzemanovic (approved):

PR #312, opened 2021-06-30
host env runtime error handling

### LINKED ISSUE #311: move host env calls failure conditions handling
The fatal failures in the `shared/src/vm/host_env.rs` cannot shut down execution with panic. Instead, the wasmer runtime allows to return `Result`s from the host functions, which should be used for fatal runtime errors only. Specifically, these are:
- tx/VP running out of gas
- tx storage write to a non-existing address
- update a VP of an account with an invalid VP
- initializing an account with an invalid VP
- wasm memory errors
- parsing/serialization and any unsafe conversions

tzemanovic:
closes #311


yito88 (approved):
Looks good! Thank you!

PR #314, opened 2021-07-01
VP gas improve error handling

### LINKED ISSUE #313: nicer parallelized VP out-of-gas error short-circuiting
Following the changes done for #312, it is possible to match on the runtime error that occurred in the WASM execution, if any. We should use that to perform short-circuiting by returning an error inside the `.par_iter().try_fold(...` if the runtime error is a gas error.

Currently, we using a somewhat hacky workaround in which we store the error in `VpGasMeter`, which can then be removed.

Fraccaman:
Close #313


tzemanovic (approved):
LGTM, thanks!

PR #317, opened 2021-07-01
code review refactors

tzemanovic:


shared/src/vm/prefix_iter.rs
```diff
@@ -54,14 +51,14 @@ where
 {
     fn default() -> Self {
         Self {
-            index: PrefixIteratorId::new(0),
+            index: PrefixIteratorId::default(),
             iterators: HashMap::new(),
```
batconjurer, line 55: Very pedantic, but it would look more consistent to also use `HashMap::default()`

batconjurer (approved):
Tiny comment. Take it or leave it

shared/src/vm/prefix_iter.rs
tzemanovic: done, thanks!

Issue #230, opened 2021-05-20
ledger genesis initialization


tzemanovic:
The genesis initialization is currently performed in `Shell::new`, but it should only happen in `Shell::init_chain`

batconjurer:
There are several functions a `Shell` can *theoretically* run on an uninitialized chain, but it is not clear if we *should* allow it. For example, `mempool_validate` or `begin_block`.

tzemanovic:
> There are several functions a `Shell` can _theoretically_ run on an uninitialized chain, but it is not clear if we _should_ allow it. For example, `mempool_validate` or `begin_block`.

I don't think we need to worry about that for now as the initialization is specified and handled by Tendermint ABCI. The handshake first invokes `GetInfo` request and if no previous state it will issue `InitChain`. We could potentially check that we're on the expected block height in `begin_block`, but there's no danger in `mempool_validate`.

batconjurer:
Clear, that makes this much simpler.

tzemanovic:
done in #319

PR #319, opened 2021-07-01
Issue #230 Changed genesis initialization in Shell

batconjurer:
Changed the genesis initialization in the shell to occur in `Shell::init_chain`. Also fixed some linting


apps/src/lib/node/ledger/mod.rs
```diff
@@ -296,6 +208,88 @@ impl Shell {
                 current_chain_id, chain_id
             )));
         }
+        if self.last_state().is_none() {
```
tzemanovic, line 211: We could either skip this check or make it a fatal error as this shouldn't be called if there is a state

tzemanovic (approved):
besides the minor comment, LGTM, thanks!

batconjurer:
General question, is the shell intended to be run only by ABCI? If so, we can remove the public visibility on the internal methods and just expose run. Then we could just rely on the correctness of the ABCI and don't need these checks.

tzemanovic:
> General question, is the shell intended to be run only by ABCI? If so, we can remove the public visibility on the internal methods and just expose run. Then we could just rely on the correctness of the ABCI and don't need these checks.

Yup, that sounds good!

tzemanovic (approved):

Issue #57, opened 2021-03-15
What to do/add in the CI


sribst:
It would be nice to have a CI ! 
Lets use this issue to report all stuff we want the CI to check

batconjurer:
I would propose running clippy with the --all-targets flags. I like nicely formatted unit tests too :smile: I ran it on master and caught only a few minor things (overzealous cloning!). So it should be a simple add.

Fraccaman:
I think it would be a good idea!

PR #320, opened 2021-07-02
Added the --all-targets flag to clippy linting

batconjurer:
In order to get good clippy linting on our unit tests, the --all-targets flag has been added to the clippy linting. This satisfies the request I made on Issue #57 (specifically [this](https://github.com/anomanetwork/anoma/issues/57#issuecomment-872435713))

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -4,7 +4,7 @@ kind: pipeline
 name: anoma-ci-build-pr
 node: {project: anoma}
 steps:
-- commands: [echo "5f49930241fc317b8c1efb460b2411b39fb6e9fbf5bb526c230b3b848bd76686  Makefile"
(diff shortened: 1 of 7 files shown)


batconjurer:
And as a great first check, clippy failed the build because of linting in some unit tests

tzemanovic:
> And as a great first check, clippy failed the build because of linting in some unit tests

Good stuff, thank you! I'm thinking maybe we should just replace the `make clippy` recipe with the one for `make clippy-check` that denies warnings - I don't think we need both. If you could change this here, I'd be grateful!

batconjurer:
> > And as a great first check, clippy failed the build because of linting in some unit tests
> 
> Good stuff, thank you! I'm thinking maybe we should just replace the `make clippy` recipe with the one for `make clippy-check` that denies warnings - I don't think we need both. If you could change this here, I'd be grateful!

So shall we delete the clippy-check recipe altogether?

tzemanovic (approved):
LGTM, thank you!

Issue #328, opened 2021-07-06
IBC validity predicate for client validation


yito88:
depends on #327

PR #329, opened 2021-07-06
validate CreateClient in IBC validity predicate

yito88:
a part of #328

- Add a validation for `CreateClient` case
- Check the IBC prefix of the key
- Fix the native VP context to use `RefCell`
  - `ClientReader` of `ibc-rs` requires that the context is immutable, but the gas meter should be mutable to add the gas cost
- Reserve a prefix character for IBC-related data


shared/src/ledger/native_vp.rs
```diff
@@ -61,9 +73,9 @@ where
     H: StorageHasher,
 {
     /// Storage prefix iterators.
-    pub iterators: PrefixIterators<'a, DB>,
+    pub iterators: Rc<RefCell<PrefixIterators<'a, DB>>>,
```
tzemanovic, line 76: I don't think `Rc` is needed here and for the `gas_meter` as each native VP instance should have their own instances

shared/src/types/storage.rs
```diff
@@ -236,6 +240,19 @@ impl Key {
             _ => false,
         }
     }
+
+    /// Check if the given key is a key to IBC-related data
+    pub fn is_ibc_key(&self) -> bool {
+        self.segments[0] == DbKeySeg::StringSeg(RESERVED_IBC_KEY.to_owned())
```
tzemanovic, line 246: the access by index will panic for an empty vec - it would be safer to use `get`

shared/src/ledger/native_vp.rs
```diff
@@ -241,3 +268,78 @@ where
...
+    H: 'static + StorageHasher,
+{
+    fn client_type(&self, client_id: &ClientId) -> Option<ClientType> {
+        let path = Path::ClientType(client_id.clone()).to_string();
+        let key = Key::ibc_key(path)
+            .expect("Creating a key for a client type shouldn't fail");
+        match self.read_post(&key) {
+            Ok(Some(value)) => {
+                let s: String = storage::types::decode(&value)
+                    .expect("decoding shouldn't fail");
```
tzemanovic, line 284: A malicious tx could write a value that would fail decoding here and panic. We can use `.ok()` on the `Result` instead. `Option` type also supports the `?` syntax, so this could be:
```suggestion
                    .ok()?;
```

Same for all the `expect`s below

shared/src/ledger/ibc.rs
```diff
@@ -22,15 +27,272 @@ impl NativeVp for Ibc {
...
+enum IbcPrefix {
+    Client,
+    Connection,
+    Channel,
+    Packet,
+    Unknown,
+}
+
+fn get_ibc_prefix(key: &Key) -> IbcPrefix {
+    match &*key.segments[1].raw() {
```
tzemanovic, line 94: The index access can panic here too

shared/src/ledger/ibc.rs
```diff
@@ -22,15 +27,272 @@ impl NativeVp for Ibc {
...
+        "clients" => IbcPrefix::Client,
+        "connections" => IbcPrefix::Connection,
+        "channelEnds" => IbcPrefix::Channel,
+        "packets" => IbcPrefix::Packet,
+        _ => IbcPrefix::Unknown,
+    }
+}
+
+fn get_client_id(key: &Key) -> Result<ClientId> {
+    ClientId::from_str(&key.segments[2].raw())
```
tzemanovic, line 104: and here

shared/src/ledger/native_vp.rs
```diff
@@ -241,3 +268,78 @@ where
         }
     }
 }
+
+impl<'a, DB, H> ClientReader for Ctx<'a, DB, H>
```
tzemanovic, line 272: I think we could move this impl to the ibc module, so that none of the ibc imports are here.

shared/src/ledger/vp_env.rs
```diff
@@ -27,6 +27,8 @@ pub enum RuntimeError {
     NumConversionError(TryFromIntError),
     #[error("Memory error: {0}")]
     MemoryError(Box<dyn std::error::Error + Sync + Send + 'static>),
+    #[error("IBC key error: {0}")]
+    IbcKeyError(String),
```
tzemanovic, line 31: It's not specific just to this PR, but I think maybe we should add an associated `type Error` to the `trait NativeVp`, so that each native VP impl can provide its own error type. Currently, we're just using this `Error` type imported from here and it's breaks the IBC module's encapsulation.

shared/Cargo.toml
```diff
@@ -17,9 +17,12 @@ testing = [
   "rand_core",
 ]
 wasm-runtime = [
+  "ibc",
```
tzemanovic, line 20: instead of grouping these with `wasm-runtime` I think it maybe worth separating them into a new feature. Then only enable this in `apps/Cargo.toml` for `anoma_shared`

shared/src/ledger/mod.rs
```diff
@@ -1,8 +1,11 @@
 //! The ledger modules
 
 pub mod gas;
+#[cfg(feature = "wasm-runtime")]
```
tzemanovic, line 4: If we move all the ibc impl for `native_vp::Ctx` into the `ibc` module (https://github.com/anomanetwork/anoma/pull/329/files#r665105370), we can then only put the condition on the `ibc` module and in `native_vp` on the ibc module import for genesis initialization and `Ibc::init_genesis_storage`

shared/Cargo.toml
yito88: That looks better. Thank you! I added `ibc-vp` feature.

shared/src/ledger/native_vp.rs
yito88: That makes sense! Thanks!

shared/src/ledger/vp_env.rs
yito88: Thank you for your suggestion! I've added `Error` for `NativeVp`.

tzemanovic (approved):
LGTM, thanks!

shared/src/ledger/ibc.rs
```diff
@@ -1,36 +1,419 @@
...
+        for key in keys_changed {
+            if !key.is_ibc_key() {
+                continue;
+            }
+
+            let accepted = match Self::get_ibc_prefix(key) {
+                IbcPrefix::Client => {
+                    let client_id = Self::get_client_id(key)?;
+                    if !clients.insert(client_id.clone()) {
+                        // this client has been checked
```
cwgoes, line 84: why would we be iterating over it twice, then? what if a different (not necessarily valid) value is written?

shared/src/ledger/ibc.rs
```diff
@@ -1,36 +1,419 @@
...
+                Ok(StateChange::Deleted)
+            }
+        } else if self.ctx.has_key_post(&key)? {
+            Ok(StateChange::Created)
+        } else {
+            Ok(StateChange::NotExists)
+        }
+    }
+
+    fn validate_created_client(&self, client_id: &ClientId) -> Result<bool> {
```
cwgoes, line 190: we should actually call `ibc-rs` here to check that the client state and consensus state are semantically valid, yes?

cwgoes (dismissed):
See comments; there should be client-type-specific validation, I think.

shared/src/ledger/ibc.rs
yito88: Do you mean that the function is [`Initialize()`](https://github.com/cosmos/ibc-go/blob/ca7f92d0a173a1a4d3fbb565cdba7aab015a5d83/modules/core/02-client/keeper/client.go#L35) in `ibc-go`?

shared/src/ledger/ibc.rs
yito88: Thank you for your comment.
It avoids the repeated validations because `clients/{identifier}/clientType`, `clients/{identifier}/clientState` and `clients/{identifier}/consensusStates/{height}` are checked at once.
However, it looks better to check the suffix and validate each state.

cwgoes: ah I see, never mind

shared/src/ledger/ibc.rs
cwgoes: yes

cwgoes: do we call that somewhere?

yito88: The `Initialize()` checks if the client type is consistent between the client state and the consensus state.
`ibc-rs` wraps these states as enum [`AnyClientState`](https://github.com/informalsystems/ibc-rs/blob/47899196c271f523a1db395202d77072beaa0593/modules/src/ics02_client/client_state.rs#L44) and [`AnyConsensusState`](https://github.com/informalsystems/ibc-rs/blob/47899196c271f523a1db395202d77072beaa0593/modules/src/ics02_client/client_consensus.rs#L46), and we can check the consistency as [here](https://github.com/anomanetwork/anoma/blob/44e9ffc8f6c10961769460a301a00fc16ccdcbb2/shared/src/ledger/ibc.rs#L222).

PR #330, opened 2021-07-07
wasm single crate

### LINKED ISSUE #302: build default wasm from single crate with conditional compilator
as done for `wasm_for_tests` #294, we can build `wasm` from one crate with conditional compilation, which speeds-up the build time a lot (as all the deps are shared)

However, we should keep the `*_template`s wasms separated as they are to be used as templates for custom implementations.

tzemanovic:
closes #302 

All the provided wasm scripts are built from a single crate, which makes the CI about twice as fast.

I've also moved the logic implemented in `matchmaker_template` and `filter_template` into the single source's `mm_token_exch` (matchmaker token exchange) and `mm_filter_token_exch` features and removed the logic from the templates as they should provide an clean slate for custom implementations.


wasm/wasm_source/Makefile
```diff
@@ -0,0 +1,65 @@
...
+wasms += vp_token
+wasms += vp_user
+wasms += mm_filter_token_exch
+wasms += mm_token_exch
+
+# Build all wasms
+all: $(wasms)
+
+# `cargo check` all wasms
+check:
```
yito88, line 20: just a trivial stuff
```suggestion
check:
```

wasm_for_tests/wasm_source/Makefile
```diff
@@ -13,17 +15,44 @@ wasms += vp_memory_limit
 wasms += vp_read_storage_key
 
 
+# Build all wasms
 all: $(wasms)
 
+# `cargo check` all wasms
+check:
```
yito88, line 22:
```suggestion
check:
```

wasm_for_tests/wasm_source/Makefile
```diff
@@ -13,17 +15,44 @@ wasms += vp_memory_limit
...
 
+# Build all wasms
 all: $(wasms)
 
+# `cargo check` all wasms
+check: 
+	$(foreach wasm,$(wasms),make check_$(wasm) && ) true
+
+# `cargo check` all wasms
+check:
```
yito88, line 26:
```suggestion
check:
```

tzemanovic:
I've updated the `assert_cmd` dependency as the older version had unmaintained transitive dependency failing the `cargo audit` CI job

yito88 (approved):
LGTM!

PR #336, opened 2021-07-12
epochs

### LINKED ISSUE #331: Protocol parameters - epochs
Various components (e.g. PoS and DKG) rely on a notion of epochs. We should decide if we want to measure these in a number of blocks or time. Epochs should be uniquely identified by consecutive integers.

In the base ledger, the duration should be set in a protocol parameter that is written to and read from storage to allow for changes. Any changes to protocol parameters could be applied by validators and switched on 2/3 voting power consensus. This could be implemented as a native validity predicate.

tzemanovic:
closes #331. It adds a native VP for protocol paramters, which currently include just the epoch duration. The parameters updates are currently prohibited (follow-up #351, #352). I've also added some wrapper types in `shared/src/types/time.rs` with added borsh serialization.

spec updates:
https://github.com/anoma/anoma/blob/tomas/epochs/tech-specs/src/explore/design/ledger/parameters.md
https://github.com/anoma/anoma/blob/tomas/epochs/tech-specs/src/explore/design/ledger/epochs.md

I partialy fixed the cargo audit warning `RUSTSEC-2021-0076` by updating `prost = "0.8.0"`, but it's still triggered from  transitive deps of libp2p, which I didn't update because it contains some breaking changes: follow-up #340


apps/src/lib/genesis.rs
```diff
@@ -12,6 +13,7 @@ use sha2::{Digest, Sha256};
 #[derive(Debug)]
```
batconjurer, line 13:
```rust 
#[derive(Debug)]
pub struct Genesis {
    #[cfg(not(feature="dev"))]
    pub validators: Vec<Validator>,
    #[cfg(feature="dev")]
    pub validators: Validator,
    pub parameters: Parameters,
}
```
is terser than duplicating the struct twice (which also looks more surprising at first)

shared/src/ledger/storage/mod.rs
```diff
@@ -413,18 +516,158 @@ pub mod testing {
...
+            min_duration in 1..100_i64,
+        )
+        (
+            min_num_of_blocks in Just(min_num_of_blocks),
+            min_duration in Just(min_duration),
+            start_height in Just(start_height),
+            start_time in Just(start_time),
+            block_height in start_height + 1..(start_height + 2 * min_num_of_blocks as u64),
+            block_time in start_time + 1..(start_time + 2 * min_duration),
+            min_blocks_delta in -(min_num_of_blocks as i64 - 1)..5,
```
batconjurer, line 568: I assume this `-`  in front of `(min_num_of_blocks...)` is not a mistake, but it is not clear what it should mean

batconjurer (commented):
Looks pretty good. I had a couple of small comments, but you can take them or leave them

apps/src/lib/genesis.rs
tzemanovic: that's is better, applied, thanks!

shared/src/ledger/storage/mod.rs
tzemanovic: I've added some comments here: `Delta will be applied on the min_num_of_blocks parameter`. The function generates a random number from this range, the left bound is set to a value so that after the delta is applied, the `min_num_of_blocks` doesn't go below 1 (`min_num_of_blocks - (min_num_of_blocks - 1) == 1`

batconjurer: :+1:

batconjurer (changes requested):
It looks like you recommitted the old tendermint.rs file in the ledger. In master it is now called tendermint_node.rs (at some point the file name conflicted with something in my cargo.toml so I changed the name).

batconjurer (approved):

PR #341, opened 2021-07-15
IBC validity predicate for updating client

### LINKED ISSUE #328: IBC validity predicate for client validation

yito88:
closes #328 

- Using non-released `ibc-rs` for fixing the mock bugs

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -105,9 +105,9 @@ dependencies = [
  "signal-hook",
  "sparse-merkle-tree",
  "tempfile",
- "tendermint 0.20.0",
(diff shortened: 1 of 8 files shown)


cwgoes (commented):
Generally this looks fine. Will multiple updates to the same client within the same transaction be accepted? As long as there is a valid chain of updates they should be.

yito88:
Thank you for your comment. I didn't consider that case. I'll fix it.

tzemanovic (approved):

cwgoes:
> Thank you for your comment. I didn't consider that case. I'll fix it.

Thank you; your changes look sensible to me. We should always consider this case, unless for some reason it is very hard.

PR #344, opened 2021-07-16
Use tower-abci in the ledger. Compatible with Tendermint v0.34.x

### LINKED ISSUE #240: integrate tower-rs for ABCI
Integrate https://github.com/hdevalence/tower-abci. It's built on top of tendermint-rs but with new domain types (this will hopefully be upstreamed in https://github.com/informalsystems/tendermint-rs/pull/862)

batconjurer:
[Tower-abci](https://github.com/hdevalence/tower-abci/tree/16bbe02c3237222ef2767303cdd532cd492c905a) provides a nice way to write an ABCI server that is compatible with the tower crate and is asynchronous. This replaces the previous ABCI server.

closes #240


apps/src/lib/node/ledger/mod.rs
```diff
@@ -1,425 +1,214 @@
...
-                            .map_err(|e| format!("{}", e));
-
-                        reply.send(result).map_err(|e| {
-                            Error::AbciChannelSendError(format!(
-                                "ApplyTx {}",
-                                e
-                            ))
-                        })?
+    fn call(&mut self, req: Request) -> Self::Future {
+        tracing::info!(?req);
```
tzemanovic, line 38: These are very verbose - for txs they print out the wasm code in bytes. Because info is on by default, we should only use it for things that are useful for users. You could maybe still print the request kind in debug

apps/src/lib/node/ledger/mod.rs
```diff
@@ -1,425 +1,214 @@
...
+                Ok(Response::OfferSnapshot(Default::default()))
+            }
+            Request::LoadSnapshotChunk(_) => {
+                Ok(Response::LoadSnapshotChunk(Default::default()))
+            }
+            Request::ApplySnapshotChunk(_) => {
+                Ok(Response::ApplySnapshotChunk(Default::default()))
+            }
+        };
+        tracing::info!(?rsp);
```
tzemanovic, line 123: this could be debug too

apps/src/lib/node/ledger/mod.rs
```diff
@@ -1,425 +1,214 @@
...
+#[tokio::main]
+async fn run_shell(
+    config: config::Ledger,
+    abort_registration: AbortRegistration,
+) {
+    // Construct our ABCI application.
+    let service = Shell::new(&config.db, config::DEFAULT_CHAIN_ID.to_owned());
+
+    // Split it into components.
+    let (consensus, mempool, snapshot, info) = split::service(service, 1);
```
tzemanovic, line 141: the `bound` arg can probably be greater to allow serving more concurrent requests, iiuc

apps/src/lib/node/ledger/mod.rs
```diff
@@ -1,425 +1,214 @@
...
+        .unwrap();
 
-    /// Begin a new block.
-    fn begin_block(&mut self, hash: BlockHash, height: BlockHeight) {
-        self.gas_meter.reset();
-        self.storage.begin_block(hash, height).unwrap();
-    }
+    // Run the server with the shell
+    let future =
+        Abortable::new(server.listen(config.address), abort_registration);
```
tzemanovic, line 166: I'm not sure how to do this with the async inside the `Abortable`, but previously in `tendermint.rs:149`, we sent a special message to shut down Anoma node when Tendermint node goes down with `sender.send(AbciMsg::Terminate)`. I think it would be nice to do something similar

tzemanovic (changes requested):
Other than the couple comments, looks good!

apps/src/lib/node/ledger/mod.rs
batconjurer: :+1:

apps/src/lib/node/ledger/mod.rs
batconjurer: Ok, trying to pick a sensible value here. According to `tower`:
```
When Buffer’s implementation of poll_ready returns Poll::Ready, it reserves a slot in the channel 
for the forthcoming call. However, if this call doesn’t arrive, this reserved slot may be held up for a 
long time. As a result, it’s advisable to set bound to be at least the maximum number of concurrent 
requests the Buffer will see. If you do not, all the slots in the buffer may be held up by futures that
have just called poll_ready but will not issue a call, which prevents other senders from issuing new
requests.
```
I think our max number if 4 in this case, so I'm going to tentatively suggest 5. Apparently wildly overshooting it can hide backpressure according to the tower-abci doc string
```
However, large buffers hide backpressure from propagating to the caller.
```

apps/src/lib/node/ledger/mod.rs
batconjurer: I actually re-architected this whole function to ensure graceful shutdowns on panics in child theads.

tzemanovic: created #349

tzemanovic (approved):

Issue #349, opened 2021-07-20
Handle Tendermint node shut down


tzemanovic:
When the Tendermint Go node shuts down (try with e.g. `killall tendermint`) while Anoma ledger node is running, we should shut down Anoma node too.

_Originally posted by @tzemanovic in https://github.com/anoma/anoma/pull/344#discussion_r672517737_

tzemanovic:
done in #350

PR #350, opened 2021-07-20
Handle tendermint Shutdown [Fixes Issue #349]

batconjurer:
Anoma will now shut down if the Tendermint process ends. Fixes issue #349


batconjurer:
I added an e2e test for this an it is pretty quick. Should not be a problem to keep it.

tzemanovic (approved):
LGTM, thanks!

PR #359, opened 2021-07-21
Cargo audit as cron job

Fraccaman:
Removing `cargo audit` from the PR pipeline and moving it to a cron job.
The cron job is executed once a day and the output of `cargo audit` is posted on github as an issue.


tzemanovic (approved):
LGTM, thanks!

PR #360, opened 2021-07-21
WIP: Technical design of DKG in Anoma

batconjurer:
This specifies how we use Ferveo and ABCI++ to implement DKG in Anoma as well as how encrypted transactions will be included and the protocol for decrypting them.

**EDIT: The  relevant docs have been significantly reworked after the comments, conversations, etc. Good to re-review this with fresh eyes.**

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201757896515792/1201762890435940) by [Unito](https://www.unito.io)


tech-specs/src/explore/design/dkg.md
```diff
@@ -1 +1,154 @@
...
+
+Some parts of the DKG protocol may be done through the gossip layer or through
+off-chain communication. However, the entire DKG cannot exist solely in the 
+gossip layer as consensus on the final shared secrets must be achieved. This 
+will require posting information on chain.
+
+## Starting the Protocol
+
+To initializing the protocol to generate a new key, the  current block 
+proposer proposes a new instance. This would be posted on the blockchain and
```
cwgoes, line 16: clarify what you mean by "new instance" (new instance of the DKG) h/t @Acentelles

tech-specs/src/explore/design/dkg.md
```diff
@@ -1 +1,154 @@
...
+notify all the participating validators to begin the announcement phase. 
+However, when should the proposer make this decision? In all likelihood it will be a combination of blocks transpired since
```
cwgoes, line 18: bad phrasing, the proposer doesn't make this decision, this is determined by the protocol (prior to the start of a new epoch)

cc @joebebel

tech-specs/src/explore/design/dkg.md
```diff
@@ -1 +1,154 @@
...
+
+When deciding to aggregate the PVSS instances into a single one, we shall 
+consider the instances ranked in terms of the weights of their dealers (in 
+descending order). The topmost dealers constituting enough weight are then 
+chosen and aggregated in their specified order.
+
+We note that enough weight may be gossiped to the current proposer and thus
+they can put the PVSS transactions and aggregation all in the same proposed 
+block.
+
```
ggkitsas, line 72: Note that any number of PVSS transcripts can be aggregated. So even if we don't get enough weight we can still propose one transaction which will be the partial aggregation the current PVSS transcripts.

tech-specs/src/explore/design/dkg.md
```diff
@@ -1 +1,154 @@
...
+   proposer adds a `PartitionShares` message to the block after the block starting
+   the DKG instance. It is based off all the `AnnounceStake` transactions
+   which will also be included in the next block (for verification).
+ - `PartitionShares`: Validators verify the partitions and weights and
+   should extract their weight shares and store them in the DKG state machine.
+   Then they should send out `Deal` transactions.
+ - `Deal`: These need to be verified. The current block proposer checks if 
+   enough PVSS instances have been reached to meet the security threshold.
+   If so, it adds a `Aggregate` transaction to the block. Otherwise, the
+   next block proposer will take on the job.
```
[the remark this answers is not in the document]
ggkitsas: If I understand correctly, this bit suggests that **only** an `Aggregate` transaction will be posted to the blockchain. 
That seems contradicting to (or I'm missing something)  https://github.com/anoma/anoma/pull/360/files#diff-0154dfbaff793ae84655cf88ba919b9136d89a979d9065a50e95456516a7283dR49

tech-specs/src/explore/design/dkg.md
joebebel: I think it's maybe better to have the DKG process start at a specific block height; for example, once block 400000*i is reached, then everyone knows that a new DKG phase begins synchronously (without any message or proposal)

tech-specs/src/explore/design/dkg.md
joebebel: Right, it's best if the protocol explicitly specifies the height, then there's no need for an explicit announcement

tech-specs/src/explore/design/dkg.md
```diff
@@ -1 +1,154 @@
...
+of the total weight, it is necessary for validators to announce their stake.
+
+Furthermore, the total weight will be scaled down to a fixed parameter 
+\\( W\\) (and fractional weights rounded appropriately to acheive integral
+values). After this procedure, each validator has a corresponding number of
+weight shares. 
+
+Lastly, the DKG protocol requires a canonical ordering of validators with
+shares must be agreed upon. This will be done via the consensus layer.
+The current block proposer will apportion out weight shares as well as the
```
joebebel, line 36: I think the partitioning should also happen implicitly. Assuming that, when the DKG start block height is reached, everyone agrees on the staking weights of the validator set, then everyone can locally compute the same partitioning - it doesn't have to be posted on the chain.

tech-specs/src/explore/design/dkg.md
```diff
@@ -1 +1,154 @@
...
+In a single DKG instance, a subset of validators holding at least
+\\( \frac{2}{3} \\) of the weight shares must each act as dealers. It
+is the responsibility of a dealer to post a signed message to the blockchain
+which includes the session id (\\( \tau \\)) as well the data of the
+PVSS (we do not concern ourselves with the details here). The creation of this
+message is handled by the Ferveo library.
+
+This means that the signed message is a transaction that must be gossiped and
+eventually included on the blockchain. For this particular transaction, there
+is a special verification that must be done, otherwise the transaction should
```
joebebel, line 51: This is going to be a little messy: we want to do signature verification on the "deal" transaction, but otherwise defer all verification until a later stage. The reason for this is that in the "optimistic" case all of the PVSS messages can be batch-verified. This will save a lot of compute, but we'll have to add a complaint message as well to handle the pessimistic case.

tech-specs/src/explore/design/dkg.md
```diff
@@ -1 +1,154 @@
...
+Ferveo library.
+
+## Aggregation
+
+At some point, enough validators (meaning their combined weight  shares
+exceeds \\( \frac{2}{3}\\) of the total) have their PVSS on the chain 
+for the session with id \\( \tau \\). At this point, the next block proposer can
+aggregate the messages into a single instance using Ferveo. This is then 
+included in the next block. Once this DKG is on the blockchain, it can be 
+used for encryption of transactions until the next DKG instance is run.
```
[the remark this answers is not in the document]
joebebel: Yes, it's better to have PVSS transcripts posted to the chain, because the cost (extra on-chain storage) is outweighed by the benefits (simplicity, lower overall compute time) of synchronization - and the PVSS transcripts should be pruned from the blockchain archive once the relevant epoch is over.

Right now I don't think any off-chain gossiping is needed. Other than the PVSS transactions being gossiped to the block proposers like other txs.

tech-specs/src/explore/design/dkg.md
joebebel: it's not clear to me if there is a benefit to partial aggregation (unlike the asynchronous case, where partial aggregation is a critical part of the protocol)

Mostly because in the on-chain aggregation case, only one final aggregation is needed, so I think it's hard to do better than that by splitting up the aggregation.

tech-specs/src/explore/design/dkg.md
```diff
@@ -1 +1,154 @@
...
+We note that enough weight may be gossiped to the current proposer and thus
+they can put the PVSS transactions and aggregation all in the same proposed 
+block.
+
+## Notes on pipelining
+
+Note that in the decription above, at least two blocks must be submitted to the blockchain in order to
+produce the next public encryption key. The first block determines the 
+partition of weight shares of the validators and the last aggregates the 
+PVSS instances from the dealers leading to the final key.
```
joebebel, line 78: if the partitioning happens implicitly, then it doesn't need it's own block, but we still need multiple blocks (to hold the PVSS instances, which are rather large, and to hold the aggregation)

tech-specs/src/explore/design/dkg.md
```diff
@@ -1 +1,154 @@
...
+
+In practice, it may take even longer as the protocol waits for enough PVSS 
+instances to get onto the blockchain. Depending on the chosen refresh rate
+of the keys, another instance of the DKG protocol may need to be started 
+before the current one finishes. This is possible so long as all data
```
joebebel, line 83: I don't think we want DKG instances to overlap, though? (At least, I sincerely hope the DKG doesn't take a full day to run)

I think the problem is that the public key for epoch i must be created by running the DKG during epoch i-1, which presents the following issues:

1. The validator set during epoch i-1 might be very different than epoch i, so in the worst case scenario maybe the old validator set censors all the DKG messages sent by the new validator set. So for safety there must be a fallback for the DKG to keep going into the new epoch (if necessary).
2. The staking weights for epoch i must be finalized before the DKG for epoch i begins

tech-specs/src/explore/design/dkg.md
```diff
@@ -1 +1,154 @@
...
+posted for a DKG instance includes the session id \\( \tau \\).
+
+This means that the storage of a node must be adapted. Each instance of a
+DKG protocol from Ferveo is a state machine that keeps all necessary data
+from the protocol instance. We thus need to store the following data
+ - A list of sesssion ids from the last completed DKG instance up until the
+   latest started (we may prune ids for failed instances however)
```
joebebel, line 90: for simplicity, maybe it's easiest to use an epoch number (or something easily derivable from the block heights) as the session id?

tech-specs/src/explore/design/dkg.md
```diff
@@ -1 +1,154 @@
...
+before the current one finishes. This is possible so long as all data
+posted for a DKG instance includes the session id \\( \tau \\).
+
+This means that the storage of a node must be adapted. Each instance of a
+DKG protocol from Ferveo is a state machine that keeps all necessary data
+from the protocol instance. We thus need to store the following data
+ - A list of sesssion ids from the last completed DKG instance up until the
+   latest started (we may prune ids for failed instances however) 
+ - The DKG state machine associated to each of these session ids.
+
```
joebebel, line 92: One thing I want to add to the persistent state are the validators encryption keys (used to send PVSS messages). That is to say, the keys which were previously part of the "Announce" message should instead be stored outside of the DKG state machine.

tech-specs/src/explore/design/dkg.md
```diff
@@ -1 +1,154 @@
...
+##  The DKG state machine
+
+We first describe each new message type needed for the protocol.
+- `StartDKG`: A message from the current block proposer to start the protocol
+- `AnnounceStake`: Validators participating in this instance of the DKG
+  broadcast their stake
+- `PartitionShares`: Once the block containing all of the `AnnounceStake`
+  messages are finalized, the set of participants is fixed and the
+  partition of the weight shares is added to the block.
```
joebebel, line 101: I think the "announce" message should be moved out of the DKG state machine (each validator, either at genesis or when first joining the network, announces their encryption key, and occasionally updates with a new key - but completely outside of the DKG protocol)

Then the "StartDKG" and "PartitionShares" messages, I think, should be implicit (local API triggers) instead of network messages (since presumably each validator already has all the information available to determine when the DKG starts and how to partition)

tech-specs/src/explore/design/dkg.md
```diff
@@ -1 +1,154 @@
...
+- `Deal`: This message contains the PVSS transcript from a validator
+  acting as a dealer
+- `Aggregate`: Combine the PVSS instances into a single one, thereby
+  producing the final key.
+
```
joebebel, line 106: One more minor message is needed (unfortunately): "Complaint" which indicates that a "Deal" message was invalid.

Hopefully this means that the messages (and state machine) will be much simplified from what it was before: three messages (Deal, Aggregate, and Complaint) and two states (Sharing and Complete)

tech-specs/src/explore/design/dkg.md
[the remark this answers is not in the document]
joebebel: The PVSS instances must be posted to the blockchain; otherwise there is no guarantee that a validator will receive enough information to properly verify the "Aggregate" transaction

tech-specs/src/explore/design/dkg.md
```diff
@@ -1 +1,154 @@
...
+   next block proposer will take on the job. 
+ - `Aggregate`: This simply needs to be verified and the resulting key added
+   to the DKG instance. 
+
+Most of the above transactions occur as a reaction to other messages
+and/or data in the proposed block header. The three exceptions are the 
+`StartDKG`, `PartitionShares`, and `Aggregate` transactions. These must be 
+made by the current block proposer if their current DKG state machine is in the 
+appropriate state and/or the appropriate conditions are met. Thus block 
+proposers must check the state machines of their active DKG instances.
```
joebebel, line 134: So in the simplified model I think it looks like:

if 2/3 weight of Deal transactions has been posted, AND I am the next block producer, then I must include in the next block an Aggregate message (aggregating all previous Deal transactions) OR include a Complaint message (if one or more Deal transactions was invalid)

tech-specs/src/explore/design/dkg.md
```diff
@@ -1 +1,154 @@
...
+
+We list the phase of ABCI++ in which each of the above actions takes place.
+ - `StartDKG`: This is added to the block by the current proposer as part of 
+    the Prepare Proposal phase.
+ - `AnnounceStake`: These are sent out by validators during the Vote
+    Extension phase.
+ - `PartitionShares`: This also happens in the Vote Extension phase and the
+    result will be included in the next block's header.
+ - `Deal`: Validators should verify the weight shares included in the block 
+    header. If validation fails, the DKG instance fails.  Otherwise, `Deal`
```
joebebel, line 150: Since partition shares should happen implicitly, I think we don't have to worry about its failure

tech-specs/src/explore/design/dkg.md
```diff
@@ -1 +1,154 @@
...
+ - `AnnounceStake`: These are sent out by validators during the Vote
+    Extension phase.
+ - `PartitionShares`: This also happens in the Vote Extension phase and the
+    result will be included in the next block's header.
+ - `Deal`: Validators should verify the weight shares included in the block 
+    header. If validation fails, the DKG instance fails.  Otherwise, `Deal` 
+    happens during the Vote Extension.
+ - `Aggregate`: This is done during the Vote Extension and will be included
+    in the header of the next block. The final result should be verified
+    by validators in the next round or else the protocol has failed.
```
joebebel, line 154: OK, this is an interesting question; when a "Deal" or "Aggregate" message fails to verify, we can either ignore the offending message and resume the protocol as if it never existed (except for some slashing penalty) or restart the protocol completely from the beginning. 

While I don't think there's much major difference between the two, I think it's probably better to just ignore any message with a valid Complaint against it and resume.

tech-specs/src/explore/design/dkg.md
batconjurer: That was a

> I think it's maybe better to have the DKG process start at a specific block height; for example, once block 400000*i is reached, then everyone knows that a new DKG phase begins synchronously (without any message or proposal)

That actually was a question that I had, namely if it had to be initiated by a validator or it would be scheduled at regular intervals.

batconjurer: But I will take the suggestion from @joebebel for now

tech-specs/src/explore/design/dkg.md
batconjurer: :+1:

batconjurer: I assume that we will initialize the protocol at fixed blockheights with an exception being that if the protocol fails, we immediately start again?

tech-specs/src/explore/design/dkg.md
batconjurer: I actually to to update this bit as the proposer should not receive these PVSS via gossiping but rather as part of the `VoteExtension` phase. They will aggregate as part of the `VerifyVoteExtension` phase and this will be put in the block header. Then we *should* get enough PVSS transcripts as otherwise we would not have received enough precommits for the block to be finalized anyway. The only way we would not (barring problems with the rounding of the weight shares) is that some validator sent a vote extension without a PVSS transcript. Do we want to punish such behavior?

joebebel: Ok, that sounds very reasonable (as long as the amount of data is not too big)
I don't know if punishing failure to attach a PVSS matters. I would suggest just treating it like a missing vote (which I guess counts as punishment)

tech-specs/src/explore/design/dkg.md
```diff
@@ -1 +1,154 @@
...
+However, when should the proposer make this decision? In all likelihood it will be a combination of blocks transpired since
+the last instance completed as well as special cases (such as an instance
+failing to complete).
+
+## Announcing and distributing weight shares
+
+In order for the DKG protocol to work, it requires the contribution of 
+ \\(\frac{2}{3} \\) of the total staked weight among live validators. In 
+order to know when a subset of validators own at least \\(\frac{2}{3}\\)
+of the total weight, it is necessary for validators to announce their stake.
```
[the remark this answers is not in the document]
batconjurer: This means that we could give out weight shares to non-live validators. This probably not a big concern, but in the current setup, since announcements must be proactive, we get a handle on who is live and who isn't. But if we assume liveness, we can make this change

tech-specs/src/explore/design/dkg.md
batconjurer: We just need to specify a canonical ordering. Descending in terms of stake almost gets us there but we need a tie break on equally staked weights. Maybe validator addresses sorted lexicographically?

tech-specs/src/explore/design/dkg.md
batconjurer: So this part is a bit out of date. We will get the `Deal` messages as vote extensions, not as transactions. So all of the verification can happen in `VerifyVoteExtension` in bulk. Since that doesn't trigger until 2/3 validators have responded, the optimistic case should be good.  If something goes wrong, complaints  (instead of an aggregation) can be added to header instead I suppose.

tech-specs/src/explore/design/dkg.md
joebebel: As a practical matter, I think the most important thing is to align the DKG's generated key lifetime with the staking epoch, so the protocol should enforce that the DKG is on the same schedule as staking changes

tech-specs/src/explore/design/dkg.md
batconjurer: I was thinking something like semver where we do E.I where E is the epoch and I is an instance number starting at 0 for each epoch and incremented as necessary if we have to do multiple runs in an epoch due to failures and whatnot.

tech-specs/src/explore/design/dkg.md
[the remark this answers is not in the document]
joebebel: My concern is that liveness (or lack of liveness) should not propagate past the actual unlive period. For example, if a validator is unlucky enough to go offline during some of the DKG, they should not be locked out of validating the entire epoch - but if they fail to announce during the DKG, then they don't get access to their key shares and they're effectively jailed the entire time.

On one hand, sure we want to incentivize liveness, but I'm concerned this creates a denial-of-service attack window  where a short attack can affect the network for a longish period

tech-specs/src/explore/design/dkg.md
joebebel: Hmm, yeah. We want the complaint case (either a bad `Deal` or bad `Aggregate`) to be punished, so hopefully it never happens (and certainly not many times in a row) but we do have to handle it somehow in a not-too-expensive way (I am a bit concerned that generating the `Complaint` might be rather slow)

tech-specs/src/explore/design/dkg.md
batconjurer: I included this section before it was clear how often we would generate keys. But now that once per epoch should be the target, this is much less critical. However, I agree we need a fallback in the case of radical validator changes. I suppose we just use the old key a bit longer until a new one can be created?

tech-specs/src/explore/design/dkg.md
batconjurer: I'm realizing that I haven't spent much time thinking about punishments at all. I will add a section to the spec about it but it is probably worth some brainstorming as to how we want to achieve it.

tech-specs/src/explore/design/dkg.md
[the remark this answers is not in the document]
batconjurer: This is a good point. I agree with you. And generally we don't expect a lack of liveness to be a huge issue I think.

tech-specs/src/explore/design/dkg.md
batconjurer: I am a bit concerned about the size of the data. Especially as you said elsewhere these transcripts are rather large. From the abci++ perspective, it is pretty clear they want these kind of extra protocols done during vote extension phases but it puts a lot of weight on our headers.

tech-specs/src/explore/design/dkg.md
batconjurer: Actually, there is no guarantee that we can decrypt on old key either. I'm not sure what we should do in this case. The only two options I see is 
1. Accept unencrypted transactions until a new key is made
2. Accept no transactions until a new key is made.

tech-specs/src/explore/design/dkg.md
joebebel: I had been imagining that we could do fewer transcripts per block (over more blocks) but if that is not really compatible with vote extensions, then we could split transcripts in the opposite direction: everyone posts a tiny bit of their transcript per block. Although it's slightly more complex to keep track of which validators missed posting some part of their transcripts and need to continue to provide more

tech-specs/src/explore/design/dkg.md
joebebel: I think this is related to the broader question of what happens to old transactions in the mempool that don't make it on chain in time before the key expires.

Unfortunately I think there are not many options other than not accepting new transactions - hopefully only a few blocks have to be empty before the new DKG finishes, and it only happens because 2/3 of validators in the previous epoch chose to behave maliciously (in which case they can do a lot more bad stuff)

tech-specs/src/explore/design/dkg.md
batconjurer: What we could do is to decide to split up the posting of the PVSS transcripts across `n` blocks. Since these are ordered, we say at block `i` that the transcripts corresponding to weight shares between `n*i` and `n*(i+1)` should be submitted.

tech-specs/src/explore/design/dkg.md
batconjurer: :+1:

tech-specs/src/explore/design/dkg.md
```diff
@@ -1 +1,157 @@
...
+  producing the final key.
+- `Complain`: A message stating that a given validator's PVSS transcript or
+  a block proposer's aggregation failed verification. 
+
+In Ferveo, the DKG protocol is modelled as a state machine. This allows a
+validator to initialize an instance and drive it forward by reacting to
+each of the above message in the appropriate way. All necessary state
+is persisted by this state machine. We describe the  appropriate action for
+each of the above transactions.
+
```
Acentelles, line 117: I think I understand this a bit differently:
Each validator sends a `Deal` message with their PVSS. The DKG listens to these messages until a threshold has been met (66%). Then it (internally) triggers the shares generation step. Once shares have been distributed, a node sends an `Aggregate` message which is added to the block header, or a `Complain` if it can't aggregate the transcripts succesfully, which in turn sets the DKG state machine to an Invalid state.
After this `Aggregate` message, we still need consensus from validators that `Check` this aggregation. They can also issue a `Complain` message if they don't succeed.

Acentelles: Screenshot 2021-08-19 at 11 41 23

batconjurer: We could do that, but I think it would be better to aggregate in as many validators as possible. That way we don't limit which 2/3 set can be used later for decryption (in case of validators going down etc).

batconjurer: This diagram just indicates that at least 2/3 should deal, not that we only aggregate the fist 2/3 that are received. I'm more in favor of  aggregatoing all shares accrued with the `n` blocks of time designated. If that is 99% of the validators, that is even better right? It should just be at least 2/3

batconjurer: Actually, re-looking at how vote extensions work, this may indeed be how we want to do it.

batconjurer: Actually, I'm quickly becoming convinced. @joebebel What do you think? The big question is how to accomodate spreading of sharing PVSS transcripts across multiple blocks.

batconjurer: Ok, after much confusion on my part, I understand what you're saying. I will make the appropriate changes.

Acentelles: We'd also need to discuss what the contents of the `Invalid` state are, so that the next block proposer can lookup what went wrong. E.g. the `complain` message from an invalid pvss during aggregation may contain different data than the `complain` message from finding the aggregation itself invalid

Acentelles: This is a more conventional graph where messages are ellipses and blue rectangles are states. Hope it helps. I'm still not clear with the complain messages

Screenshot 2021-08-19 at 15 27 38

joebebel: Right, I think we still have to figure out how the Invalid state can recover back to to the Deal/Aggregate states (mostly just pretend the bad data doesn't exist, I guess?)

tech-specs/src/explore/design/dkg.md
```diff
@@ -1 +1,178 @@
...
+DKG scheme. It does not cover encryption, decryption, or validation of transactions.
+
+Some parts of the DKG protocol may be done through the gossip layer or through
+off-chain communication. However, the entire DKG cannot exist solely in the 
+gossip layer as consensus on the final shared secrets must be achieved. This 
+will require posting information on chain.
+
+## Starting the Protocol
+
+The protocol to generate a new key will be scheduled to begin at specific
```
cwgoes, line 15: in general, we can use present tense ("is") instead of future ("will be")

tech-specs/src/explore/design/dkg.md
```diff
@@ -1 +1,178 @@
...
+gossip layer as consensus on the final shared secrets must be achieved. This 
+will require posting information on chain.
+
+## Starting the Protocol
+
+The protocol to generate a new key will be scheduled to begin at specific 
+block heights, in pratices this will be at the beginning of every eposh.
+This way all the validators know when to begin the announcement phase. In 
+addition to starting at the pre-determined blocks heights, if the protocol 
+fails, all validators know to immediately start a new round of the protocol.
```
cwgoes, line 19: assuming that the validators can detect that the protocol has failed, we should clarify this

tech-specs/src/explore/design/dkg.md
```diff
@@ -1 +1,178 @@
...
+
+It is important to note that the key that is being generated is to be used
+encrypting transactions in the epoch following the current one. However,
+this relies on the assumption that after the validators sets change between
+epochs that there is still validators whose combined weight exceeds 
+\\( \frac{2}{3} \\) of \\W\\. Otherwise, it will be impossible to decrypt
```
cwgoes, line 25: why? clarify please

tech-specs/src/explore/design/dkg.md
```diff
@@ -1 +1,178 @@
...
+
+__TBD__: Do we accept unencrypted transactions while the new key is generated
+or do we stop accepting transactions until the new key is available? Or 
+something else (that is hopefully smarter)?
+
+##  Distributing weight shares
+
+In order for the DKG protocol to work, it requires the contribution of 
+ \\(\frac{2}{3} \\) of the total staked weight among validators. At the
+beginning of the DKG protocol, Tendermint can be queried each validator
```
cwgoes, line 36: queried by each validator?

tech-specs/src/explore/design/dkg.md
```diff
@@ -1 +1,178 @@
...
+transactions. If this does not happen a new key will need to be generated.
+
+__TBD__: Do we accept unencrypted transactions while the new key is generated
+or do we stop accepting transactions until the new key is available? Or 
+something else (that is hopefully smarter)?
+
+##  Distributing weight shares
+
+In order for the DKG protocol to work, it requires the contribution of 
+ \\(\frac{2}{3} \\) of the total staked weight among validators. At the
```
cwgoes, line 35: this is because we choose the threshold. note that we choose the threshold to be the same as proof-of-stake (2/3)

tech-specs/src/explore/design/dkg.md
```diff
@@ -1 +1,178 @@
...
+beginning of the DKG protocol, Tendermint can be queried each validator
+to get the set of validators, their associated address, and voting power
+(or weight).
+
+The total weight will be scaled down to a fixed parameter 
+\\( W\\) (and fractional weights rounded appropriately to achieve integral
+values). After this procedure, each validator has a corresponding number of
+weight shares.
```
cwgoes, line 43: corresponding weight?

tech-specs/src/explore/design/dkg.md
```diff
@@ -1 +1,178 @@
...
+beginning of the DKG protocol, Tendermint can be queried each validator
+to get the set of validators, their associated address, and voting power
+(or weight).
+
+The total weight will be scaled down to a fixed parameter 
+\\( W\\) (and fractional weights rounded appropriately to achieve integral
+values). After this procedure, each validator has a corresponding number of
+weight shares. 
+
+The DKG protocol requires a canonical ordering of validators. The ordering
```
cwgoes, line 45: why does the DKG need an ordering? clarify

tech-specs/src/explore/design/dkg.md
```diff
@@ -1 +1,178 @@
...
+will be chosen by ordering validators in terms of decreasing voting power.
+If there is a tie, it is broken by considering the lexicographic ordering
+on their associated addresses. 
+
+This allows each validator to independently compute the partition of weight
```
cwgoes, line 50: not sure exactly what this sentence means, do you mean that each validator computes how many shares they get? how many shares all of the other validators get?

tech-specs/src/explore/design/dkg.md
```diff
@@ -1 +1,178 @@
...
+
+## Dealers
+In a single DKG instance, a subset of validators holding at least
+\\( \frac{2}{3} \\) of the weight shares must each act as dealers. It
+is the responsibility of a dealer to post a signed message to the blockchain
+which includes the session id (\\( \tau \\)) as well the data of the
+PVSS (we do not concern ourselves with the details here). The creation of this
+message is handled by the Ferveo library.
+
+The signed message will be returned as part of the vote extension and there
```
cwgoes, line 62: when you mention an ABCI step (e.g. vote extension) cite the spec section explicitly

tech-specs/src/explore/design/dkg.md
```diff
@@ -1 +1,178 @@
...
+   of the aggregated key fails, a `Complain` should be issued.
+
+While every validator must send and react to `Deal` messages, `Aggregate` 
+must be made by the current block proposer if their current DKG state machine 
+is in the  appropriate state and/or the appropriate conditions are met. Thus 
+block proposers must check the state machines of their active DKG instances.
+
+## ABCI++
+
+The creation of the public key and the shared secrets requires ABCI++ as we
```
cwgoes, line 141: cite the ABCI++ spec (link)

tech-specs/src/explore/design/dkg.md
```diff
@@ -1 +1,178 @@
...
+We list the phase of ABCI++ in which each of the above actions takes place.
+
+ - `Deal`: `Deal` happens during the Vote Extension phase.
+ - `Aggregate`: This is done during the Prepare Proposal phase (it could be
+   done earlier if validators know they are likely to be the next proposer)
+   and will be  included in the header of the proposed block. The final result 
+   should be verified by validators. If validation fails, the next block 
+   proposer can try to aggregate the the PVSS transcripts correctly and 
+   include the result in the header of the subsequent block. Furthermore, 
+   `Complains` may be issued against the validator that incorrectly
```
cwgoes, line 154: when does this happen w.r.t. ABCI++?

tech-specs/src/explore/design/dkg.md
```diff
@@ -1 +1,178 @@
...
+   pass validation
+ * `Aggregate_Complaint`: The aggregated key in a proposed block is not valid
+ 
+These transactions must include a proofs of malfeasance. The necessary 
+proofs for the above complaints are as follows:
+ * `PVSS_Complaint`: The transcript that failed verification
+ * `Aggregate_Complaint`: The alleged incorrect key and the PVSS transcripts
+    that should have been used to derive it. 
+
+The above proofs can be used to verify that the protocol was not adhered to.
```
cwgoes, line 174: these are just verified by the state machine, yes?

tech-specs/src/explore/design/ledger/front-running.md
```diff
@@ -2,6 +2,210 @@
...
+outer transaction which wraps an inner transaction within to be included 
+after it is decrypted.
+
+## Wrapper transactions
+
+A wrapper transaction contains an encrypted payload that will be another
+transaction upon decryption. In addition to this payload, the wrapper 
+transaction will contain the following data:
+ * fee: the price of including the block onto the chain
+ * fee payer: the account paying the above fee
```
cwgoes, line 29: this needs to be an implicit account

tech-specs/src/explore/design/ledger/front-running.md
```diff
@@ -2,6 +2,210 @@
...
+ 
+The encryption of the wrapped transaction using the public key from the DKG
+and the derivation of the ciphertext public key are provided by the Ferveo
+library.
+
+For a wrapped transaction to be included in a block proposal, the following 
+conditions must be met:
+ * The fee payers signature must be checked
+ * The fee payer must have sufficient balance to pay the fee
+ * The fee payer must have sufficient balance to pay
```
cwgoes, line 47: duplicate condition?

tech-specs/src/explore/design/ledger/front-running.md
```diff
@@ -2,6 +2,210 @@
...
+
+However, to prevent users from setting gas limits higher than necessary, if the gas limit of a transaction is specified as 
+k * `GAS_LIMIT_RESOLUTION`, then the user will have to pay for at least 
+(k - 1) * `GAS_LIMIT_RESOLUTION` units of gas. Otherwise, they are charged only
+for the units of gas that their transaction actually consumes. And as usual, if
+their transaction exceeds the gas limit, the application of the transaction
+is halted.
+
+### Encrypting a transaction
+The key used to encrypt transactions for the current is created from the
```
cwgoes, line 76: for the current what?

tech-specs/src/explore/design/ledger/front-running.md
```diff
@@ -2,6 +2,210 @@
...
+ - No alteration to block after Prepare Proposal phase
+ - Less communication overhead
+#### Cons:
+ - Validators need to now in advance that they are likely to be the next proposer
+ - Preparing new proposals is blocked until decryption is finished (which can be slow)
+ - More than one validator will do decryption even though there will be a
+   single proposer (wasted effort)
+
+### Async-decrypt protocol
+###TODO: This section doesn't make sense yet, need to figure out the details better
```
cwgoes, line 121: not reviewing yet then

cwgoes (commented):
Read through and left some more comments, nice work so far!

tech-specs/src/explore/design/dkg.md
batconjurer: At this point, I'm not sure the protocol can fail. We just keep waiting until it finishes. I suppose it hits a natural "timeout" if somehow it lasts a whole epoch, but that shouldn't realistically ever happen.

tech-specs/src/explore/design/dkg.md
batconjurer: This is because the "keys for decryption" were only given to those who participated in the previous epoch. Tried to clarify that.

tech-specs/src/explore/design/dkg.md
batconjurer: We are computing several things offline now and need deterministic results. That way we know everyone has the same truth even though we do not use the consensus layer. Tried to clarify

tech-specs/src/explore/design/dkg.md
batconjurer: Everyone computes the shares that everyone has. And by everyone, I mean validators. Tried to clarify.

tech-specs/src/explore/design/dkg.md
batconjurer: Complaints got reworked a lot based on yesterday's conversation. They aren't an explicit thing anymore, so this is outdated.

tech-specs/src/explore/design/dkg.md
batconjurer: I think it is actually more explanatory to cite the RFC.

tech-specs/src/explore/design/dkg.md
batconjurer: Again, this is outdated as complaints have been reworked

tech-specs/src/explore/design/ledger/front-running.md
batconjurer: epoch. oops

tech-specs/src/explore/design/ledger/front-running.md
batconjurer: :+1:

tech-specs/src/explore/design/dkg.md
cwgoes: what do you mean by "if the protocol fails" then?

batconjurer: It actually doesnt say anything about failing anymore. I should have resolved this conversation when I changed that.

juped:
marked draft because it says WIP

PR #366, opened 2021-07-26
IBC connection validations

### LINKED ISSUE #338: IBC validity predicate for connection validation
depends on #327

yito88:
closes #338 

I added a new file for the connection validations.
I'm going to move client-related functions to another file in another PR.


shared/src/ledger/ibc/connection.rs
```diff
@@ -0,0 +1,423 @@
...
+{
+    pub(super) fn validate_connection(
+        &self,
+        key: &Key,
+        tx_data: &[u8],
+    ) -> Result<bool> {
+        if Self::is_counter_key(key) {
+            // the counter should be increased
+            return Ok(
+                self.connection_counter_pre() < self.connection_counter()
```
tzemanovic, line 44: should we check that it is exactly `+1` or is it enough to check that it is greater?

shared/src/ledger/ibc/connection.rs
```diff
@@ -0,0 +1,423 @@
...
+            .map(|h| TendermintConsensusState::from(h).wrap_any())
+    }
+
+    fn connection_counter(&self) -> u64 {
+        let path = COUNTER_PATH.to_owned();
+        let key = Key::ibc_key(path)
+            .expect("Creating a key for a connection counter failed");
+        match self.ctx.read_post(&key) {
+            Ok(Some(value)) => storage::types::decode(&value)
+                .expect("converting a connection counter shouldn't failed"),
```
tzemanovic, line 404: I think we have to handle the decoding error without `expect` to avoid panic on malicious state update

shared/src/ledger/ibc/connection.rs
tzemanovic: I guess it can be greater than `+1` when there is more than 1 new connection in the same tx, right?

shared/src/ledger/ibc/connection.rs
```diff
@@ -0,0 +1,423 @@
...
+use super::{Error, Ibc, Result, StateChange};
+use crate::ledger::storage::{self, StorageHasher};
+use crate::types::address::{Address, InternalAddress};
+use crate::types::ibc::{
+    ConnectionOpenAckData, ConnectionOpenConfirmData, ConnectionOpenTryData,
+    Error as IbcDataError,
+};
+use crate::types::storage::{BlockHeight, Epoch, Key, KeySeg};
+
+pub(super) const COUNTER_PATH: &str = "connections/counter";
```
tzemanovic, line 29: maybe we could add a function to the storage key with the
```
let path = COUNTER_PATH.to_owned();
Key::ibc_key(path).expect("Creating a key for a connection counter failed");
```

shared/src/types/ibc.rs
```diff
@@ -188,3 +196,420 @@ impl ClientUpgradeData {
...
+)]
+pub struct ConnectionOpenInitData {
+    /// The corresponding client ID
+    client_id: String,
+    /// The corresponding counterparty
+    counterparty: Vec<u8>,
+    /// The version
+    version: Vec<u8>,
+    /// The delay period as (secs, nanos)
+    delay_period: (u64, u32),
```
tzemanovic, line 212: we can make this a little nicer by adding a type for this in `time.rs` module, e.g. `DurationNano { secs: u64, nanos: u32 }` with `impl From<std::time::Duration> for DurationNano` and `impl From<DurationNano> for std::time::Duration`

shared/src/ledger/ibc/connection.rs
yito88: That's right. Multiple connections can be created in the same tx.

shared/src/ledger/ibc/connection.rs
yito88: Thanks! Fixed.
I'll fix that of the client counter in another PR.

shared/src/ledger/ibc/connection.rs
yito88: Thank you for your suggestion!
I added the function to the storage key.

shared/src/types/ibc.rs
yito88: That looks great! Modified

shared/src/types/ibc.rs
```diff
@@ -188,3 +191,422 @@ impl ClientUpgradeData {
...
+    pub fn delay_period(&self) -> Duration {
+        Duration::new(self.delay_period.secs, self.delay_period.nanos)
+    }
+}
+
+/// Data to try to open a connection
+#[derive(Debug, Clone, BorshSerialize, BorshDeserialize)]
+pub struct ConnectionOpenTryData {
+    /// The previous connection ID
+    prev_conn_id: Option<String>,
```
cwgoes, line 265: why do these fields need to be provided in the message data, aren't they inferable from the state writes?

cwgoes (commented):
Is it possible to use more of `ibc-rs` connection logic instead of doing it ourselves? If not, it's alright, this is simple enough, just checking. Quick question on required fields.

shared/src/types/ibc.rs
yito88: I removed `prev_conn_id`. And, I modified validations to use connection info instead of these data because we have to check the created/updated connection is valid. (The connection should be created/updated according to these data.)
I think others are used in a transaction to create/update connections.

tzemanovic (approved):

yito88:
@cwgoes I think I use `ibc-rs` logic as possible as I can for now.
If we can use some logic in the upstream after some fixes, I'm going to ask or modify them.
I've already asked the upstream to make proof verification functions public to use them in IBC VP and it has been approved.

PR #367, opened 2021-07-26
Improve intent structure

### LINKED ISSUE #176: Improve token exchange intent data type
The default token exchange intent data type is quite simple and should be improved to allow more expressive intent

maybe something similar to:
```rust 
pub struct Intent {
    addr: String,
    exchange: HashMap<(String, u64, u64) ,<HashSet<String, f64>>
    created: Time
    time_to_live: Duration
}
```
Where exchange is : `(token_sell, min_amount,max_amount) ->[(token_buy, rate)]`

Fraccaman:
Close #176.


apps/src/lib/cli.rs
```diff
@@ -697,43 +702,53 @@ pub mod args {
...
+        pub max_sell: Vec<token::Amount>,
+        /// Rate
+        pub min_rate: Vec<DecimalWrapper>,
         /// Token to buy
-        pub token_buy: Address,
-        /// Amount of token to buy
-        pub amount_buy: token::Amount,
+        pub token_buy: Vec<Address>,
+        /// Min amount of token to buy
+        pub min_buy: Vec<token::Amount>,
```
tzemanovic, line 717: I think since all these `Vec` must have the same length, it would be nicer to make the data type e.g.:

```
    pub struct CraftIntent {
        /// A key that will sign the whole intent
        pub key: Address,
        pub exchanges: Vec<TokenExchange>,
        /// Target file path
        pub file_path: String,
    }

    pub struct TokenExchange {
        /// Source address that will sign the exchange
        pub addr: Address,
        /// Token to sell
        pub token_sell: Address,
        /// Max amount of token to sell
        pub max_sell: token::Amount,
        /// Rate
        pub min_rate: DecimalWrapper,
        /// Token to buy
        pub token_buy: Address,
        /// Min amount of token to buy
        pub min_buy: token::Amount,
    }
```

shared/src/types/intent.rs
```diff
@@ -41,15 +65,113 @@ pub struct IntentTransfers {
...
+}
+
+#[allow(missing_docs)]
+#[derive(Error, Debug)]
+pub enum Error {
+    #[error("Error parsing as decimal: {0}.")]
+    DecimalParseError(String),
+}
+
+impl From<token::Amount> for Result<DecimalWrapper, Error> {
```
tzemanovic, line 104: I think you should use `TryFrom` instead

shared/src/types/token.rs
```diff
@@ -63,6 +63,11 @@ impl Amount {
             micro: amount * 1_000_000,
         }
     }
+
+    /// return the f64 rappresentation
```
tzemanovic, line 67:
```suggestion
    /// return the f64 representation
```

shared/src/types/token.rs
```diff
@@ -63,6 +63,11 @@ impl Amount {
...
+    pub fn to_f64(&self) -> f64 {
+        self.micro as f64
```
tzemanovic, line 69: It's probably better to convert from the micro units
```suggestion
        self.micro as f64 / 1_000_000
```

tests/src/e2e.rs
```diff
@@ -406,11 +411,246 @@ mod tests {
...
+
+        session_gossip
+            .exp_regex(".crafting transfer: Established: a1qq5qqqqqg4znssfsgcurjsfhgfpy2vjyxy6yg3z98pp5zvp5xgersvfjxvcnx3f4xycrzdfkak0xhx, Established: a1qq5qqqqqxsuygd2x8pq5yw2ygdryxs6xgsmrsdzx8pryxv34gfrrssfjgccyg3zpxezrqd2y2s3g5s, 100000000*")
+            .map_err(|e| eyre!(format!("{}", e)))?;
+
+        drop(session_gossip);
+        drop(session_ledger);
+        drop(session_send_intent_a);
+        drop(session_send_intent_b);
+        drop(session_send_intent_c);
```
tzemanovic, line 643: I think you can skip all the `drop`s as they will drop when they go out of scope and there's nothing happening after it

wasm/wasm_source/Cargo.toml
```diff
@@ -27,6 +27,9 @@ petgraph = {version = "0.5.1", features = ["serde-1"], optional = true}
 serde = {version = "1.0.125", optional = true}
 serde_json = {version = "1.0.64", optional = true}
 wee_alloc = "0.4.5"
+rust_decimal = { version = "1.14.3", optional = true }
+rust_decimal_macros = { version = "1.14", optional = true }
```
tzemanovic, line 31: I think this is unused, right?

wasm/wasm_source/src/mm_token_exch.rs
```diff
@@ -0,0 +1,353 @@
...
+                == current_node.exchange.data.token_buy
+                && new_node.exchange.data.max_sell
+                    >= current_node.exchange.data.min_buy
+            {
+                connect_sell.push(index);
+            }
+            if new_node.exchange.data.token_buy
+                == current_node.exchange.data.token_sell
+                && current_node.exchange.data.max_sell
+                    > new_node.exchange.data.min_buy
```
tzemanovic, line 100: Should this be
```suggestion
                    >= new_node.exchange.data.min_buy
```
?

wasm/wasm_source/src/vp_user.rs
```diff
@@ -0,0 +1,276 @@
...
+            Self::InvalidIntentSet(address)
+        } else {
+            Self::Unknown
+        }
+    }
+}
+
+// transfer - [albert, bertha, matchmaker, btc, xan]
+
+// transfer [(some exchanges), signature(matchmaker)]
```
tzemanovic, line 32:
```suggestion
```

tests/src/e2e.rs
```diff
@@ -406,11 +411,246 @@ mod tests {
...
+        //  Start gossip
+        let mut session_gossip = spawn_command(base_node_gossip, Some(40_000))
+            .map_err(|e| eyre!(format!("{}", e)))?;
+
+        //  Start ledger
+        let mut session_ledger = spawn_command(base_node_ledger, Some(40_000))
+            .map_err(|e| eyre!(format!("{}", e)))?;
+
+        session_ledger
+            .exp_string(&"No state could be found".to_string())
```
tzemanovic, line 540:
```suggestion
            .exp_string("No state could be found")
```

tests/src/e2e.rs
```diff
@@ -406,11 +411,246 @@ mod tests {
...
+        ]);
+
+        let mut session_send_intent_a =
+            spawn_command(send_intent_a, Some(20_000))
+                .map_err(|e| eyre!(format!("{}", e)))?;
+
+        // means it sent it correctly but not able to gossip it (which is
+        // correct since there is only 1 node)
+        session_send_intent_a
+            .exp_regex(".*Failed to publish_intent InsufficientPeers*")
```
tzemanovic, line 568: I think when there's no regex, you can just use
```suggestion
            .exp_string("Failed to publish_intent InsufficientPeers")
```
same for all other `exp_regex` below

tzemanovic (changes requested):
Mostly looks good! I suggested a couple smaller changes

shared/src/types/token.rs
tzemanovic: even better, I think it would be nicer to instead `impl From<Amount> for f64`. Then you can skip manually converting in places where you call `to_f64`, because the good_lp's API accepts e.g. `max<N: Into<f64>>(self, max: N)` (`Into` is automatically implemented from `From`)

tests/src/e2e.rs
Fraccaman: idk if its related to this, but if I remove this series of `drop` the CI test e2e step doesn't stop and run infinity (https://ci.heliax.dev/anoma/anoma/703)

Fraccaman:
@tzemanovic I also changed the cli to create intents. It takes a json file describing the list of exchanges

shared/src/types/token.rs
```diff
@@ -63,6 +64,11 @@ impl Amount {
             micro: amount * 1_000_000,
         }
     }
+
+    /// return the f64 rapresentation
```
tzemanovic, line 68:
```suggestion
    /// return the f64 representation
```

tests/src/e2e.rs
```diff
@@ -406,11 +413,261 @@ mod tests {
...
+            spawn_command(send_intent_c, Some(40_000))
+                .map_err(|e| eyre!(format!("{}", e)))?;
+
+        // means it sent it correctly but not able to gossip it (which is
+        // correct since there is only 1 node)
+        session_send_intent_c
+            .exp_string("Failed to publish_intent InsufficientPeers")
+            .map_err(|e| eyre!(format!("{}", e)))?;
+
+        // // check that the amount matched are correct
```
tzemanovic, line 625: uncomment this bit

tzemanovic (changes requested):
Looks mostly good now, thank you! I made a couple more small suggestions

batconjurer (changes requested):
There was one comment from Tomas that I think should be addressed. The other two things were nits.

wasm/wasm_source/src/mm_token_exch.rs
```diff
@@ -0,0 +1,350 @@
...
+    let mut intent_graph = graph.filter_map(
+        |node_index, node| {
+            if cycle_intents.contains(&node_index) {
+                let edges = graph.neighbors(node_index);
+
+                *edges
+                    .map(|target_node_index| {
+                        let target = graph[target_node_index].clone();
+
+                        let surce_var_def = variable();
```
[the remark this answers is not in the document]
Fraccaman: yeah, ill find a better variable name

batconjurer (approved):

PR #376, opened 2021-08-02
Modularize IBC client

yito88:
Moved IBC client-related functions to a new file


shared/src/ledger/ibc/client.rs
```diff
@@ -0,0 +1,401 @@
...
+        client_id: &ClientId,
+    ) -> Result<StateChange> {
+        let path = Path::ClientState(client_id.clone()).to_string();
+        let key = Key::ibc_key(path)
+            .expect("Creating a key for a client type failed");
+        self.get_state_change(&key)
+    }
+
+    fn validate_created_client(&self, client_id: &ClientId) -> Result<bool> {
+        let client_type = match self.client_type(client_id) {
```
tzemanovic, line 69: It seems that there's a bit of pattern with the methods that return `Option`s. Maybe we could use the `Result` instead and put the information that is logged in the `tracing::info` into error types. Then we can use the `?` syntax so these all will one-lines. This way, the errors will also be visible in the tx result, which is better because you might not be able to see the node logs if you're submitting a tx remotely

tzemanovic: then we should maybe define an error type for this module instead of `use super::Error` and add `impl From<client::Error> for Error` in the `ibc/mod.rs` so we get implicit conversions

shared/src/ledger/ibc/client.rs
```diff
@@ -0,0 +1,401 @@
...
+    }
+
+    pub(super) fn client_counter_pre(&self) -> u64 {
+        let key = Key::ibc_client_counter();
+        match self.ctx.read_post(&key) {
+            Ok(Some(value)) => match storage::types::decode(&value) {
+                Ok(c) => c,
+                Err(e) => {
+                    tracing::error!("decoding a client counter failed: {}", e);
+                    u64::MAX
```
tzemanovic, line 303: it might be nicer to use an error type here

tzemanovic (changes requested):
A few suggestions, other than that LGTM!

shared/src/ledger/ibc/client.rs
yito88: Thank you for the suggestion!
I've modified the error handling for the client and connection.

tzemanovic: Thank you for applying the suggestion. If I can further add to this, what I was thinking was to return the `Result` directly from `self.client_type` and all the other methods in this scope that currently return `Option`, so then we can e.g. on this line use `let client_type = self.client_type(client_id)?;`.

yito88: Yes, that looks nicer. However, the functions of the `ClientReader` trait are defined in `ibc-rs`. We need `*Reader` traits to reuse verification functions defined in `ibc-rs`. I'll propose changing the returned type at the mainstream.

shared/src/ledger/ibc/mod.rs
```diff
@@ -30,10 +20,10 @@ pub enum Error {
     NativeVpError(native_vp::Error),
     #[error("Key error: {0}")]
     KeyError(String),
-    #[error("Decoding TX data error: {0}")]
-    DecodingTxDataError(std::io::Error),
-    #[error("IBC data error: {0}")]
-    IbcDataError(crate::types::ibc::Error),
+    #[error("Connection validation error: {0}")]
```
tzemanovic, line 23:
```suggestion
    #[error("Client validation error: {0}")]
```

tzemanovic (approved):
LGTM, thanks!

PR #377, opened 2021-08-02
ABCI++ Integration

batconjurer:
This PR implements the ABCI++ interface for our shell. However, since tendermint does not use this interface yet, a shim has been added to do a crude translation between the two interfaces.

The vast majority of calls are the same, the important exceptions follow:
 - The `BeginBlock` method is simply renamed to `PrepareProposal`
 - `DeliverTx` adds the transaction to a queue (it does not apply the transation). 
 -  `EndBlock` is forwarded to `FinalizeBlock`. This applies all the transactions in queue. At present, we still drop those that are invalid but this will need to change in the future.
 
Furthermore, several of the new calls introduced in ABCI++ are never called. These include `ProcessProposal`, `ExtendVote`, `VerifyVoteExtension`, `VerifyHeader`, and `RevertProposal`. 

This also introduces custom events which the new tendermint websocket client can subscribe to. This allows the front end client to report when a transaction has been applied and data about it even though this is now done in the `EndBlock` call rather than `DeliverTx`.


apps/src/lib/node/ledger/shell.rs
```diff
@@ -258,50 +262,87 @@ impl Shell {
...
 
+    pub fn extend_vote(
+        &mut self,
+        _req: shim::request::ExtendVote,
+    ) -> shim::response::ExtendVote {
+        Default::default()
+    }
+
+    /// Validate and apply transactions.
+    /// TODO: Will we also decrypt transactions here?
```
tzemanovic, line 298: I think we might decrypt to be decrypting them in `VerifyVoteExtension` and adding them them to a queue

apps/src/lib/node/ledger/shims/abcipp_shim.rs
```diff
@@ -0,0 +1,109 @@
...
+                    prove: false,
+                };
+                Ok(Resp::DeliverTx(ResponseDeliverTx {
+                    info: self.service.query(req).info,
+                    ..Default::default()
+                }))
+            }
+            Req::EndBlock(end) => {
+                if BlockHeight::try_from(end.height).is_err() {
+                    // TODO: Should we panic?
```
tzemanovic, line 87: ah yes, I think we should

batconjurer: :+1:

apps/src/lib/node/ledger/shell.rs
batconjurer: The queue is only necessary if we decide to encrypt gas limits right?

tzemanovic: yeah, I think we still have to have the queue

apps/src/lib/node/ledger/events.rs
```diff
@@ -0,0 +1,101 @@
...
+pub struct Event {
+    event_type: EventType,
+    attributes: HashMap<String, String>,
+}
+
+/// The two types of custom events we currently use
+#[derive(Clone)]
+pub enum EventType {
+    // The transaction was accepted to be included in a block
+    Accepted,
```
tzemanovic, line 20: This case looks unused right now. Is it for later?

apps/src/lib/node/ledger/shell.rs
```diff
@@ -266,48 +270,89 @@ impl Shell {
...
+                    }
+                    tx_result["gas_used"] = result.gas_used.to_string();
+                    tx_result["info"] = result.to_string();
+                }
+                Err(msg) => {
+                    tx_result["gas_used"] = self
+                        .gas_meter
+                        .get_current_transaction_gas()
+                        .to_string();
+                    tx_result["info"] = msg.to_string();
```
tzemanovic, line 345: This branch (occurs when there's a runtime error in a tx) is missing the response code. Let's maybe use `"2"` for it
```suggestion
                    tx_result["info"] = msg.to_string();
                    tx_result["code"] = "2".into();
```

tests/src/e2e.rs
```diff
@@ -445,6 +445,114 @@ mod tests {
...
+            .map_err(|e| eyre!(format!("{}", e)))?;
+
+        session
+            .exp_string("Anoma ledger node started")
+            .map_err(|e| eyre!(format!("{}", e)))?;
+        session
+            .exp_string("Started node")
+            .map_err(|e| eyre!(format!("{}", e)))?;
+
+        // 2. Submit a an invalid transaction
```
tzemanovic, line 478:
```suggestion
        // 2. Submit a an invalid transaction (amount > source's balance)
```

tzemanovic: Thanks for adding a test case for this! You could also maybe add one that fails in `tx` (the code "2" above), e.g. use `$BERTHA` as the token address:
```rust
vec![
            "transfer",
            "--source",
            BERTHA,
            "--target",
            ALBERT,
            "--token",
            BERTHA,
            "--amount",
            "1_000.1",
        ];
```

apps/src/lib/client/tx.rs
```diff
@@ -94,8 +94,11 @@ async fn broadcast_tx(
...
-    let query = Query::from(EventType::Tx)
-        .and_eq("tx.hash", hash_tx(&tx_bytes).to_string());
+    //
+    // Note that the `applied.hash` key comes from a custom event
+    // created by the shell
+    let query = Query::from(EventType::NewBlock)
+        .and_eq("applied.hash", hash_tx(&tx_bytes).to_string());
     client.subscribe(query)?;
     println!(
         "{:?}",
```
tzemanovic, line 104:
```suggestion
        "Transaction added to mempool: {:?}",
```

tzemanovic: it's not letting me comment on that line here, but let's also maybe change line 112 to `"Transaction applied with result: {}",`

tzemanovic (changes requested):
A couple small changes, other than that looks great!

apps/src/lib/node/ledger/events.rs
batconjurer: Yes, I was thinking we would want an event for wrapped transactions getting added onto the chain and a later event when they are decrypted and applied.

tzemanovic (approved):

PR #378, opened 2021-08-03
IBC channel validations

### LINKED ISSUE #339: IBC validity predicate for channel validation
depends on #327

yito88:
closes #339 

- Need to fix `Capability` in `ibc-rs` and update the port implementation
  - Fixed: https://github.com/informalsystems/ibc-rs/commit/ebf9d31d277b0cb8a4a19d0e147ce1c9981a1f91
  - Applied to `heliaxdev/ibc-rs`

### DIFF
--- a/shared/src/ledger/ibc/channel.rs
+++ b/shared/src/ledger/ibc/channel.rs
@@ -1,0 +1,244 @@
+//! IBC validity predicate for channel module
+
+use std::str::FromStr;
+
+use borsh::{BorshDeserialize, BorshSerialize};
+use ibc::ics02_client::client_consensus::{AnyConsensusState, ConsensusState};
+use ibc::ics02_client::client_state::AnyClientState;
+use ibc::ics02_client::context::ClientReader;
+use ibc::ics02_client::height::Height;
+use ibc::ics03_connection::connection::ConnectionEnd;
+use ibc::ics03_connection::context::ConnectionReader;
+use ibc::ics04_channel::channel::ChannelEnd;
+use ibc::ics04_channel::context::ChannelReader;
+use ibc::ics04_channel::error::{Error as Ics4Error, Kind as Ics4Kind};
+use ibc::ics04_channel::packet::{Receipt, Sequence};
+use ibc::ics05_port::capabilities::Capability;
+use ibc::ics05_port::context::PortReader;
+use ibc::ics07_tendermint::consensus_state::ConsensusState as TendermintConsensusState;
+use ibc::ics23_commitment::commitment::CommitmentPrefix;
+use ibc::ics24_host::identifier::{ChannelId, ClientId, ConnectionId, PortId};
+use ibc::ics24_host::Path;
+use ibc::proofs::Proofs;
+use ibc::timestamp::Timestamp;
+use sha2::Digest;
+use tendermint_proto::Protobuf;
+
+use super::{Error, Ibc, Result, StateChange};
+use crate::ledger::storage::{self, StorageHasher};
+use crate::types::address::{Address, InternalAddress};
+use crate::types::ibc::{
+    ConnectionOpenAckData, ConnectionOpenConfirmData, ConnectionOpenTryData,
+};
+use crate::types::storage::{Key, KeySeg};
+
+impl<'a, DB, H> Ibc<'a, DB, H>
+where
+    DB: 'static + storage::DB + for<'iter> storage::DBIter<'iter>,
+    H: 'static + StorageHasher,
+{
+    pub(super) fn validate_channel(
+        &self,
+        key: &Key,
+        tx_data: &[u8],
+    ) -> Result<bool> {
+        let port_id = Self::get_port_id(key)?;
+        let channel_id = Self::get_channel_id(key)?;
+    }
+
+    /// Returns the channel ID after #IBC/channelEnds/ports/{port_id}/channels
+    fn get_channel_id(key: &Key) -> Result<ChannelId> {
+        match key.segments.get(5) {
+            Some(id) => ChannelId::from_str(&id.raw())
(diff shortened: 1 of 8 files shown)


cwgoes (commented):
Looks like this still needs tests for the rest of the channel handshake (after INIT)?

yito88:
@cwgoes Yes, we need more tests for connection and channel handshakes.
However, some functions (e.g. [root getter](https://github.com/informalsystems/ibc-rs/blob/eefa500d6b3621c824b892f0048ef1916cba4bbb/modules/src/ics02_client/client_consensus.rs#L164)) in `ibc-rs` that are required by these tests have not been implemented yet.
I'd like to implement them and add tests.

shared/src/ledger/ibc/channel.rs
```diff
@@ -0,0 +1,661 @@
...
+
+        self.authenticated_capability(&port_id).map_err(|e| {
+            Error::InvalidPort(format!(
+                "The port is not authenticated: ID {}, {}",
+                port_id, e
+            ))
+        })?;
+
+        let port_channel_id = (port_id, channel_id);
+        let channel = match self.channel_end(&port_channel_id) {
```
tzemanovic, line 94: this can use `ok_or_else`

shared/src/ledger/ibc/channel.rs
```diff
@@ -0,0 +1,661 @@
...
+            &proofs,
+        ) {
+            Ok(_) => Ok(true),
+            Err(e) => Err(Error::ProofVerificationFailure(e)),
+        }
+    }
+
+    fn get_sequence(&self, path: Path) -> Result<Sequence> {
+        let key = Key::ibc_key(path.to_string())
+            .expect("Creating akey for a sequence shouldn't fail");
```
tzemanovic, line 362:
```suggestion
            .expect("Creating a key for a sequence shouldn't fail");
```

shared/src/ledger/ibc/channel.rs
```diff
@@ -0,0 +1,661 @@
...
+    }
+
+    fn connection_channels(
+        &self,
+        conn_id: &ConnectionId,
+    ) -> Option<Vec<(PortId, ChannelId)>> {
+        let mut channels = vec![];
+        let prefix = Key::parse("channelEnds/ports")
+            .expect("Creating a key for the prefix shouldn't fail");
+        let mut iter = match self.ctx.iter_prefix(&prefix) {
```
tzemanovic, line 486: I think you can use `.ok()?` on all the matched `Result`s here and below

shared/src/types/ibc.rs
```diff
@@ -600,3 +601,549 @@ impl ConnectionOpenConfirmData {
...
+    /// Returns the data to initalize a channel
+    pub fn new(
+        port_id: PortId,
+        order: Order,
+        counterparty: ChanCounterparty,
+        connection_hops: Vec<ConnectionId>,
+        version: String,
+    ) -> Self {
+        let port_id = port_id.as_str().to_owned();
+        let order = order.as_str().to_owned();
```
tzemanovic, line 630: These can use the `Display` instances directly
```suggestion
        let port_id = port_id.to_string();
        let order = order.to_string();
```

shared/src/types/ibc.rs
```diff
@@ -600,3 +601,549 @@ impl ConnectionOpenConfirmData {
...
+        let counterparty = counterparty
+            .encode_vec()
+            .expect("Encoding a counterparty shouldn't fail");
+        let connection_hops = connection_hops
+            .iter()
+            .map(|c| c.as_str().to_owned())
```
tzemanovic, line 636:
```suggestion
            .map(|c| c.to_string())
```

shared/src/types/ibc.rs
```diff
@@ -600,3 +601,549 @@ impl ConnectionOpenConfirmData {
...
+
+    /// Returns the connection hops
+    pub fn connection_hops(&self) -> Result<Vec<ConnectionId>> {
+        let mut hops = vec![];
+        for conn_str in &self.connection_hops {
+            let id = ConnectionId::from_str(conn_str)
+                .map_err(|e| Error::DecodingError(e.to_string()))?;
+            hops.push(id);
+        }
+        Ok(hops)
```
tzemanovic, line 673: You can do this, because `Result` implements `FromIter`:
```suggestion
        self
            .connection_hops
            .iter()
            .map(|conn_str| {
                ConnectionId::from_str(conn_str)
                    .map_err(|e| Error::DecodingError(e.to_string()))
            })
            .collect()
```
But unfortunately it's actually longer than the first version :(

Alternatively, with the first version it would be better to change the first line to pre-allocate the vec for the needed size `let mut hops = Vec::with_capacity(self.connection_hops.len())`.

shared/src/types/ibc.rs
```diff
@@ -600,3 +601,549 @@ impl ConnectionOpenConfirmData {
...
+        counterparty: ChanCounterparty,
+        connection_hops: Vec<ConnectionId>,
+        version: String,
+        counterparty_version: String,
+        proof_height: Height,
+        proof_connection: CommitmentProofBytes,
+        proof_client: CommitmentProofBytes,
+        proof_consensus: CommitmentProofBytes,
+    ) -> Self {
+        let port_id = port_id.as_str().to_owned();
```
tzemanovic, line 722: these can also use `to_string()`

tzemanovic (changes requested):
A couple small changes, besides that LGTM!

shared/src/types/ibc.rs
yito88: Thanks! I've modified it to use the iteration!

tzemanovic (approved):

yito88:
I'd like to merge this PR and will add tests for each phase of the handshakes at another PR.

Issue #298, opened 2021-06-21
storage read-only RPC


tzemanovic:
The ledger should allow to query the storage via RPC. 

We can leverage tendermint's RPC interface via ABCI. We are currently using it for transaction simulation (dry-run mode). 

We could use the storage keys for query paths, i.e.:

 - ~~`chain_id`: get the chain ID~~ edit 31/1/2022: known in config
 - ~~`height`: get the last committed block height~~ edit 31/1/2022: accessible via tendermint block RPC, will add nicer client support later
 - `h`: for each block at height `h`:
   - `tree`: merkle tree
     - `root`: root hash
     - `store`: the tree's store
   - `hash`: block hash
   - `subspace`: any byte data associated with accounts
   
Additionally, if the height `h` is not specified in the query, we could default to read from the last committed block.

tzemanovic:
A basic RPC has been added in #380, which only allows to query the current block.

Next, we should extend this to allow queries on a specified block height and add a command to query any given storage key returning the raw bytes.

tzemanovic:
follow-ups in #836, #839 and #840

PR #380, opened 2021-08-03
initial storage RPC

tzemanovic:
related to #298, in this PR: 
- refactor `dry_run_tx` query to use the new RPC module
- add RPC to read a storage value from an account sub-space the last committed block height
- add RPC read a range of values with a given prefix from an account sub-space
- add client cmd to find token balance(s)


apps/src/lib/cli.rs
```diff
@@ -405,6 +410,27 @@ pub mod cmds {
         }
     }
 
+    #[derive(Debug)]
+    pub struct QueryBalance(pub args::QueryBalance);
```
batconjurer, line 414: Nit: new line after type def

apps/src/lib/cli.rs
```diff
@@ -495,6 +521,9 @@ pub mod args {
         }));
     const LEDGER_ADDRESS_OPT: ArgOpt<tendermint::net::Address> =
         LEDGER_ADDRESS.opt();
+    const LEDGER_ADDRESS_ABOUT: &str =
```
batconjurer, line 524: I don't see any definition of a `scheme` here. Although maybe it is hidden in this wall of consts. Maybe alphebetize for searching ease?

apps/src/lib/client/rpc.rs
```diff
@@ -0,0 +1,152 @@
...
+            let key = token::balance_prefix(token);
+            let balances =
+                query_storage_prefix::<token::Amount>(client, key).await;
+            let currency_code = tokens
+                .get(token)
+                .cloned()
+                .unwrap_or_else(|| token.to_string());
+            let stdout = io::stdout();
+            let mut w = stdout.lock();
+            writeln!(w, "Token {}:", currency_code).unwrap();
```
batconjurer, line 58: why the switch from `println!` to `writeln!` on stdout

shared/src/types/address.rs
```diff
@@ -298,6 +299,20 @@ pub fn kartoffel() -> Address {
     Address::decode("a1qq5qqqqqxs6yvsekxuuyy3pjxsmrgd2rxuungdzpgsmyydjrxsenjdp5xaqn233sgccnjs3eak5wwh").expect("The token address decoding shouldn't fail")
 }
 
+/// Temporary helper for testing, a hash map of tokens addresses with their
+/// informal currency codes.
+pub fn tokens() -> HashMap<Address, String> {
```
batconjurer, line 304: this could also return a `HashMap<Address, &'static str>`  which helps with borrow semantics elsewhere in the code base. You can also acheive this code in a single allocation with 
```rust
 vec![(xan(), "XAN"),
        (btc(), "BTC"),
        (eth(), "ETH"),
        (dot(), "DOT"),
        (schnitzel(), "Schnitzel"),
        (apfel(), "Apfel"),
        (kartoffel(), "Kartoffel")]
     .into_iter()
     .collect()
 ```

**Review by batconjurer — COMMENTED** [review]
A few small suggestions / questions

File: apps/src/lib/cli.rs (hunk 1 above)
Comment (tzemanovic) [reply apps/src/lib/cli.rs:427]: done in 146ad7c

File: apps/src/lib/cli.rs (hunk 2 above)
Comment (tzemanovic) [reply apps/src/lib/cli.rs]: good point, I've sorted them in 146ad7c

File: apps/src/lib/client/rpc.rs (hunk 3 above)
Comment (tzemanovic) [reply apps/src/lib/client/rpc.rs:59]: it's because it's printing in a loop to avoid system stdout lock/unlock on every statement

File: apps/src/lib/client/rpc.rs [hunk 5]
```diff
@@ -0,0 +1,152 @@
...
+pub async fn query_balance(args: args::QueryBalance) {
+    let client = HttpClient::new(args.query.ledger_address).unwrap();
+    let tokens = address::tokens();
+    match (args.token.as_ref(), args.owner.as_ref()) {
+        (Some(token), Some(owner)) => {
+            let key = token::balance_key(token, owner);
+            let balance: token::Amount = query_storage_value(client, key).await;
+            let currency_code = tokens
+                .get(token)
+                .cloned()
```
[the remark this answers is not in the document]
tzemanovic: I've done it with 🐮 type in cd8b6c1

shared/src/types/address.rs
tzemanovic: that's better, thanks! Done in cd8b6c1

batconjurer (approved):

Issue #172, opened 2021-04-30
Update client command to submit a tx


tzemanovic:
Currently, we're using `broadcast_tx_commit` method from tendermint ABCI to submit tx to the chain and wait for block inclusion and its result. This method shouldn't be used on live version. Instead, the client should just broadcast the tx to the mempool and then subscribe to a websocket to wait for the tx result from being included in a block. See https://docs.tendermint.com/master/rpc/#/Tx/broadcast_tx_commit.

PR #386, opened 2021-08-08
Add basic tendermint rpc client

batconjurer:
For talking with tendermint child process, a basic rpc client that talks over a websocket has been added. It supports both the simple transactions as well as subscriptions to tendermint events. In particular, this solves Issue #172.

Ideally we should have been able to use the subscription clients provided by tendermint-rs but they are not mature enough at the moment (in particular the subscriptions tend to hang). I have tried to re-use as much as possible from tendermint-rs however. We still rely on our [fork](https://github.com/heliaxdev/tendermint-rs/tree/tomas/update-genesis-new-prost) where I have made the module containing the rpc methods public so that we can use them.


apps/Cargo.toml
```diff
@@ -77,16 +78,18 @@ sparse-merkle-tree = {git = "https://github.com/heliaxdev/sparse-merkle-tree", b
 # temporarily using fork work-around for https://github.com/informalsystems/tendermint-rs/issues/916
 tendermint = {git = "https://github.com/heliaxdev/tendermint-rs", branch = "tomas/update-genesis-new-prost"}
 tendermint-proto = {git = "https://github.com/heliaxdev/tendermint-rs", branch = "tomas/update-genesis-new-prost"}
-tendermint-rpc = {git = "https://github.com/heliaxdev/tendermint-rs", branch = "tomas/update-genesis-new-prost", features = ["http-client"]}
+tendermint-rpc = {git = "https://github.com/heliaxdev/tendermint-rs", branch = "tomas/update-genesis-new-prost", features = ["http-client", "websocket-client"]}
 thiserror = "1.0.24"
 tokio = {version = "1.8.2", features = ["full"]}
+tokio-test = "0.4.2"
```
tzemanovic, line 84: This should be in `[dev-dependencies]`

tzemanovic (approved):
LGTM!

PR #389, opened 2021-08-09
cleanup: don't commit protobuf build artifacts

juped:
The contents of apps/src/lib/proto/generated are build
artifacts. Remove them from the repository, and add a .gitignore
serving a dual purpose: keeping the directory in the repository,
like a .keep file, and ignoring the generated artifacts (Rust sources).


cwgoes (approved):
ACK

cwgoes:
Not sure why the build is failing though.

juped:
I've got to rebase it up past the latest CI changes for it to pass CI

tzemanovic:
If you could also pls apply the same change for `shared/src/proto/generated`. 

The CI is failing on `make fmt-check`, because this cmd doesn't build. We run `make clippy` on a previous step though, which does build the project, so maybe we could set the `fmt-check` job to depend on it (`.drone.yml:152`)

PR #394, opened 2021-08-16
Fix udeps

Fraccaman:
Instead of hardcoding the rust nightly version, its better to fetch it from the Fetching nightly version from `rust-nightly-version` file.


tzemanovic (approved):
LGTM, thanks!

PR #399, opened 2021-08-17
Improve CI on master branch

Fraccaman:
This PR introduce the following changes:
- Split wasm build + test into an independent pipeline when a PR is merged into master. This should further reduce the time spent building/testing as some actions are done in parallel.
- Fix a bug with `udeps.py` script.


tzemanovic (approved):
LGTM!

PR #402, opened 2021-08-18
Fix and Improve WASM CI

### LINKED ISSUE #401: consider using wasm-opt from Binaryen
Consider using wasm-opt from Binaryen (https://github.com/WebAssembly/binaryen) to further reduce wasm size. We could possibly use `-Os` or `Oz` option that focuses on file size. This tool should be integrated in the CI step that builds the wasms

Fraccaman:
This PR updates the image used by the CI to deterministically build wasm. It also improve the built artifacts by following the advices given [here](https://rustwasm.github.io/book/reference/code-size.html#optimizing-builds-for-code-size).
This PR address and closes #401.


Fraccaman:
Using `wasm-opt` and instructing the LLVM to optimize for size we have a ~20% reduction in size.

juped:
I can confirm this update to 1.54.0 fixes the `build-wasm-scripts-docker` target on . The CI is still wrong somehow, though. How exactly was the image CI is using built?

juped:
rerolled to da7b0208b22671752f06dbf966a732bf6e816722

juped (approved):
much needed fix, works deterministically on the platforms i've tried (linux x64, mac x64), CI still doing something different for some reason

Issue #327, opened 2021-07-06
Implement IBC validity predicate


yito88:
depends on #326 

- [ ] validate clients
- [ ] validate connections
- [ ] validate channels
- [ ] validate packets

PR #407, opened 2021-08-24
Return an error when IBC VP fails

yito88:
depends on #327


shared/src/ledger/ibc/mod.rs
```diff
@@ -124,22 +128,19 @@ where
...
-                IbcPrefix::SeqAck => false,
-                IbcPrefix::Commitment => false,
-                IbcPrefix::Receipt => false,
-                IbcPrefix::Ack => false,
+                IbcPrefix::SeqSend => continue,
+                IbcPrefix::SeqRecv => continue,
+                IbcPrefix::SeqAck => continue,
+                IbcPrefix::Commitment => continue,
+                IbcPrefix::Receipt => continue,
+                IbcPrefix::Ack => continue,
```
yito88, line 136: It temporarily continues the iteration. It will be replaced with validations.

tzemanovic (approved):

PR #410, opened 2021-08-24
Basic Wallet CLI

### LINKED ISSUE #167: wallet improvements
A temporary "wallet" module in is being added in #160. This works for testing & playnet, but before a public testnet, we'll need to add a real wallet.

One option is using https://github.com/AleoHQ/wagyu, which has support for HD wallet, mnemonics and network signing

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201757896515792/1201762924595628) by [Unito](https://www.unito.io)

### LINKED ISSUE #423: client gets newly established addresses from tx's result
Depends on #422.
In the client, we should find the newly established addresses from the tx's result and report them to the user. After we have #410, the addresses should be automatically added to the wallet.

pablohildo:
This PR adds a basic wallet as described in https://github.com/anoma/anoma/issues/167#issuecomment-900982703

Closes #423


apps/src/lib/wallet/defaults.rs
```diff
@@ -1,6 +1,45 @@
-//! Temporary helper until we have a proper wallet.
+//! Default addresses and keys.
```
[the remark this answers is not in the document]
tzemanovic: to be done in https://github.com/anoma/anoma/issues/425#issuecomment-920192899

shared/src/types/key/ed25519.rs
```diff
@@ -27,6 +29,15 @@ pub struct PublicKey(ed25519_dalek::PublicKey);
 #[derive(Debug, Serialize, Deserialize)]
 pub struct SecretKey(ed25519_dalek::SecretKey);
 
+/// Ed25519 keypair
+#[derive(Debug, Serialize, Deserialize, BorshSerialize, BorshDeserialize)]
+pub struct Keypair {
```
tzemanovic, line 34: We added this newtype for Keypair, because initially we were serializing the wallet with borsh into binary. Now that the wallet is using TOML, it's no longer needed, but it still helpful in some cases as we also have newtype wrappers for pk, sk and sig, so there're less conversions where we use these.

shared/src/types/key/ed25519.rs
```diff
@@ -460,9 +572,82 @@ impl From<SecretKey> for ed25519_dalek::SecretKey {
...
+impl From<Keypair> for ed25519_dalek::Keypair {
+    fn from(keypair: Keypair) -> Self {
+        Self {
+            secret: keypair.secret.into(),
+            public: keypair.public.into(),
+        }
+    }
+}
+
+impl Signer<Signature> for Keypair {
```
tzemanovic, line 605: all the impls from here below are just wrappers for the library code on newtypes

juped (approved):
just a couple notes:

- readme should really not have `make` and `cargo run` commands (i will open this as an issue)
- there should be shell script tests directly testing the cli (i will also open this as an issue)

PR #412, opened 2021-08-25
IBC sequence validations

### LINKED ISSUE #393: IBC validity predicate for sequence validation
depends on #327

yito88:
closes #393

### DIFF
--- a/shared/src/ledger/ibc/channel.rs
+++ b/shared/src/ledger/ibc/channel.rs
@@ -367,14 +367,14 @@ where
             Some(value) => {
                 let index: u64 =
                     storage::types::decode(value).map_err(|e| {
-                        Error::SequenceError(format!(
+                        Error::InvalidSequence(format!(
                             "Decoding a prior sequece index failed: {}",
                             e
                         ))
                     })?;
                 Ok(Sequence::from(index))
             }
-            None => Err(Error::SequenceError(format!(
+            None => Err(Error::InvalidSequence(format!(
                 "The prior sequence doesn't exist: Path {}",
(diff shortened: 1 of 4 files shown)


shared/src/ledger/ibc/channel.rs
```diff
@@ -360,6 +360,26 @@ where
...
+                        Error::InvalidSequence(format!(
+                            "Decoding a prior sequece index failed: {}",
+                            e
+                        ))
+                    })?;
+                Ok(Sequence::from(index))
+            }
+            // The sequence is updated for the first time. The previous sequence
+            // is the initial number.
+            None => Ok(Sequence::from(1)),
```
tzemanovic, line 379: Does this have to start from `1`, or could we use
```suggestion
            None => Ok(Sequence::default()),
```
?

tzemanovic (approved):
besides the question above, LGTM!

shared/src/ledger/ibc/channel.rs
yito88: Good question! I think it should start from `1`.
I tried using `Sequence::default()` but the default value is 0 and 0 isn't allowed by the packet encoding in `ibc-rs`. And, all sequences using this function have the name starting with `next`. That's why I set `1` to the initial value,  (I'm not sure that the `default()` should be modified. The sequence itself can be with `0` logically)
https://github.com/informalsystems/ibc-rs/blob/c518e15a073a5187d64cd00d1a62b2c9360a9408/modules/src/ics04_channel/packet.rs#L169

Issue #405, opened 2021-08-20
Add option to print intent to stdout


Fraccaman:
It would be nice to be able to log the intent output to stdout and not only to file

tzemanovic:
We currently have 2 commands for intent, one to `craft-intent` into a file and `intent` to submit it from a file to a gossiper's RPC. To simplify this, we can remove the `craft-intent` commands and add an argument (e.g. `--to-stdout` (`-O`)) that will print an intent to stdout. From there, one can redirect it to a file with `>`/`>>`.

Additionally, let's add this option to tx commands to, for which we can output the raw txs instead of submitting them to a node.

tzemanovic:
#417 implemented the `--stdout` option. Next, we should add a way to submit an intent from a file, e.g. something like:

```shell
1.
# write a signed intent into a file
anoma intent --raw-intent intent.B.data --key a1qq5qqqqqxv6yydz9xc6ry33589q5x33eggcnjs2xx9znydj9xuens3phxppnwvzpg4rrqdpswve4n9 --stdout > signed_intent.data

# submit the intent from the file to the RPC
anoma intent --node "http://127.0.0.1:39112" --topic "asset_v1" --intent-path signed_intent.data

2.
# write an un-signed intent into a file (when no key is specified
anoma intent --raw-intent intent.B.data --stdout > unsigned_intent.data

# sign and submit the intent from the file to the RPC
anoma intent --node "http://127.0.0.1:39112" --topic "asset_v1" --key a1qq5qqqqqxv6yydz9xc6ry33589q5x33eggcnjs2xx9znydj9xuens3phxppnwvzpg4rrqdpswve4n9 --intent-path unsigned_intent.data
```
```

We should also add the `--stdout` flag to tx commands. Similarly, we can output the tx data to stdout (signed or unsinged), instead of submitting it to the ledger's mempool and allow to submit a tx from a file.

PR #417, opened 2021-08-30
Intent CLI

Fraccaman:
Change the `intent` cli according to #405 (closing #405).
- Remove `craft-intent`.
- Add `--stdout` flag to `intent` command.

### DIFF
--- a/apps/src/bin/anoma-client/cli.rs
+++ b/apps/src/bin/anoma-client/cli.rs
@@ -65,9 +65,10 @@ async fn gossip_intent(
     let data_bytes = signed_ft.try_to_vec().unwrap();
 
     if to_stdout {
-        let mut out = std::io::stdout();
-        out.write_all(&data_bytes).unwrap();
-        out.flush().unwrap();
+        // let mut out = std::io::stdout();
+        // out.write_all(&data_bytes).unwrap();
+        // out.flush().unwrap();
+        println!("{:?}", data_bytes);
     } else {
         let mut client = RpcServiceClient::connect(node_addr).await.unwrap();
 
--- a/apps/src/bin/anoma-client/cli.rs
+++ b/apps/src/bin/anoma-client/cli.rs
@@ -65,10 +65,9 @@ async fn gossip_intent(
     let data_bytes = signed_ft.try_to_vec().unwrap();
 
     if to_stdout {
-        // let mut out = std::io::stdout();
-        // out.write_all(&data_bytes).unwrap();
-        // out.flush().unwrap();
-        println!("{:?}", data_bytes);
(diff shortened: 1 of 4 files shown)


apps/src/bin/anoma-client/cli.rs
```diff
@@ -97,6 +64,32 @@ fn craft_intent(
     );
     let data_bytes = signed_ft.try_to_vec().unwrap();
 
-    let mut file = File::create(file_path).unwrap();
-    file.write_all(&data_bytes).unwrap();
+    if to_stdout {
+        let mut out = std::io::stdout();
+        out.write_all(&data_bytes).unwrap();
+        out.flush().unwrap();
```
tzemanovic, line 70: I think this should be equivalent
```suggestion
        println!("{:?}", data_bytes);
```

tzemanovic (approved):
LGTM, thanks!

apps/src/lib/cli.rs
```diff
@@ -559,9 +532,7 @@ pub mod args {
     const MATCHMAKER_PATH: ArgOpt<PathBuf> = arg_opt("matchmaker-path");
     const MULTIADDR_OPT: ArgOpt<Multiaddr> = arg_opt("address");
     const NODE: Arg<String> = arg("node");
-    const FILE_PATH_OUTPUT: ArgDefault<String> =
-        arg_default("file-path-output", DefaultFn(|| "intent.data".into()));
-    const FILE_PATH_INPUT: Arg<String> = arg("file-path-input");
+    const TO_STDOUT: ArgFlag = flag("--stdout");
```
tzemanovic, line 535:
```suggestion
    const TO_STDOUT: ArgFlag = flag("stdout");
```

apps/src/lib/cli.rs
```diff
@@ -750,30 +721,49 @@ pub mod args {
...
             app.arg(NODE.def().about("The gossip node address."))
+                .arg(SIGNING_KEY.def().about("The key to sign the intent."))
                 .arg(DATA_PATH.def().about(
                     "The data of the intent, that contains all value \
                      necessary for the matchmaker.",
                 ))
+                .arg(
+                    TO_STDOUT
+                        .def()
+                        .about("Echo the serialized intent to stdout."),
```
tzemanovic, line 765:
```suggestion
                        .about("Echo the serialized intent to stdout. Note that with this option, the intent won't be submitted to the intent gossiper RPC."),
```

apps/src/bin/anoma-client/cli.rs
tzemanovic: this is not equivalent

PR #418, opened 2021-08-31
1/4 PoS: initial version of the PoS crate

tzemanovic:
This PR is split from #385. Closes #322 

The PoS system  and its integration is described in #324. Rendered: https://github.com/anoma/anoma/blob/tomas/pos-design/docs/src/explore/design/pos.md and https://github.com/anoma/anoma/blob/tomas/pos-design/docs/src/explore/design/ledger/pos-integration.md, respectively.

Together with #419, #420, #421, these PRs contain the PoS MVP with:

- validator self-bond
  - tx WASM + client cmd
  - VP validation
- delegation
  - tx WASM + client cmd
  - VP validation
- validator unbond from self-bond
  - tx WASM + client cmd
  - VP validation
- delegator unbond from delegation
  - tx WASM + client cmd
  - VP validation
- validator withdraw from unbond
  - tx WASM + client cmd
  - VP validation
- delegator withdraw from unbond
  - tx WASM + client cmd
  - VP validation
- slashing
  - base ledger apply slash on evidence in `BeginBlock`
  - withdraw from unbond apply slashes
  - VP validation of slashes on withdraw
- RPC
  - get bonds/unbonds for a given source address
  - get a validator's voting power
  - get total voting power
  - get validator set (active/inactive)
  - add optional arg for epoch for queries above (use current epoch if not specified)

Issue #323, opened 2021-07-02
Integrate the PoS system


tzemanovic:
depends on #322

PR #419, opened 2021-08-31
2/4 PoS: shell + native VP integration

### LINKED ISSUE #231: update genesis for testnets
Currently, the genesis module is using "dev" feature to setup a single validator in genesis. We should update this for `not(feature = "dev")` with a configurable list of validators (their address, public key and voting power).

tzemanovic:
Depends on #418. Related to #323. Closes #231


apps/src/lib/node/ledger/protocol/mod.rs
```diff
@@ -220,10 +224,31 @@ fn execute_vps(
...
-                            let result = pos
-                                .validate_tx(tx_data, keys, &verifiers_addr)
-                                .map_err(Error::PosNativeVpError);
+                            let pos = PosVP { ctx };
+                            let verifiers_addr_ref = &verifiers_addr;
+                            let pos_ref = &pos;
+                            // TODO this is temporarily ran in a new thread to
+                            // avoid crashing the ledger (required `UnwindSafe`
+                            // and `RefUnwindSafe` in
+                            // shared/src/ledger/pos/vp.rs)
```
tzemanovic, line 233: this will be turned into an issue to be removed. Depends on #434, #435 and #440

apps/src/lib/node/ledger/shell.rs
```diff
@@ -364,6 +560,57 @@ impl Shell {
             response.events.push(tx_result.into());
         }
 
+        if self.new_epoch {
+            // Apply validator set update
+            let (current_epoch, _gas) = self.storage.get_current_epoch();
+            // TODO ABCI validator updates on block H affects the validator set
+            // on block H+2, do we need to update a block earlier?
```
tzemanovic, line 567: this will be turned into an issue, the behavior should be documented

shared/src/ledger/pos/vp.rs
```diff
@@ -0,0 +1,334 @@
...
+    H: StorageHasher,
+{
+    /// Context to interact with the host structures.
+    pub ctx: Ctx<'a, DB, H>,
+}
+
+// TODO this is temporarily to run PoS native VP in a new thread to avoid
+// crashing the ledger (in apps/src/lib/node/ledger/protocol/mod.rs). The
+// RefCells contained within PosVP are not thread-safe, but each thread has its
+// own instances.
```
tzemanovic, line 59: this is related to https://github.com/anoma/anoma/pull/419/files#r703352061

tzemanovic: also `RefUnwindSafe` below

shared/src/ledger/storage/mod.rs
```diff
@@ -45,8 +46,8 @@ where
     pub header: Option<Header>,
     /// The height of the committed block
     pub last_height: BlockHeight,
-    /// The epoch of the current block
-    pub current_epoch: Epoch,
+    /// The epoch of the committed block
+    pub last_epoch: Epoch,
```
tzemanovic, line 50: Tte name wasn't correct - this is in fact the last committed block's epoch. The current epoch is inside `block.epoch` that is being updated on `BeginBlock`.

shared/src/vm/host_env.rs
```diff
@@ -1282,7 +1282,7 @@ where
     H: StorageHasher,
 {
     let storage = unsafe { env.ctx.storage.get() };
-    let (epoch, gas) = storage.get_block_epoch();
+    let (epoch, gas) = storage.get_current_epoch();
```
tzemanovic, line 1285: This wasn't correct. A tx should use the epoch of the current block (the one the tx is being applied in)

shared/src/types/token.rs
```diff
@@ -44,15 +45,16 @@ impl Amount {
         self.micro as Change
     }
 
-    /// Spend a given amount
+    /// Spend a given amount.
     /// Panics when given `amount` > `self.micro` amount.
     pub fn spend(&mut self, amount: &Amount) {
-        self.micro -= amount.micro
+        self.micro = self.micro.checked_sub(amount.micro).unwrap();
```
tzemanovic, line 51: This and `receive` were in fact not panicking, but instead the values were under-flowing/overflowing. Updated to cause a panic instead

tests/src/e2e.rs
```diff
@@ -488,21 +491,38 @@ mod tests {
             .exp_string("Started node")
             .map_err(|e| eyre!(format!("{}", e)))?;
 
-        // 2. Submit a an invalid transaction (amount > source's balance)
+        // 2. Submit a an invalid transaction (trying to mint tokens should fail
```
tzemanovic, line 494: I had to change the invalid tx example here for https://github.com/anoma/anoma/pull/419/files#r703358474 change, because now the overflow that's caused by having `amount > source's balance` doesn't fail the validation, but instead crashes the tx before we run VPs, which is preferable, because the VP would fail anyway on overflow/under-flow. I've added a new example tx in `wasm`_for_tests`:
```rust
/// A tx that attempts to mint tokens in the transfer's target without debiting
/// the tokens from the source. This tx is expected to be rejected by the
/// token's VP.
```

README.md
```diff
@@ -88,20 +88,20 @@ cargo run --bin anoma -- gossip --rpc "127.0.0.1:39111"
 cargo run --bin anoman -- gossip --rpc "127.0.0.1:39111" --matchmaker-path wasm/mm_token_exch.wasm --tx-code-path wasm/tx_from_intent.wasm --ledger-address "127.0.0.1:26657"
 
 # make intents
-# 1) create file containing the json rapresentation of the intent
-echo '[{"addr":"a1qq5qqqqqxv6yydz9xc6ry33589q5x33eggcnjs2xx9znydj9xuens3phxppnwvzpg4rrqdpswve4n9","key":"a1qq5qqqqqxv6yydz9xc6ry33589q5x33eggcnjs2xx9znydj9xuens3phxppnwvzpg4rrqdpswve4n9","max_sell":70,"min_buy":100,"min_rate":2,"token_buy":"a1qq5qqqqqxuc5gvz9gycryv3sgye5v3j9gvurjv34g9prsd6x8qu5xs2ygdzrzsf38q6rss33xf42f3","token_sell":"a1qq5qqqqq8q6yy3p4xyurys3n8qerz3zxxeryyv6rg4pnxdf3x3pyv32rx3zrgwzpxu6ny32r3laduc"}]' > intent.A.data
+# 1) create file containing the json representation of the intent
```
tzemanovic, line 91: these changes are not related to PoS, but the commands were out of date

tests/src/e2e.rs
```diff
@@ -837,12 +859,13 @@ mod tests {
         while index < n_of_peers {
             let node_path = path.join(format!("anoma-{}", index));
 
-            let mut config = Config::default();
-
-            let _ledger_config = Ledger {
-                tendermint: node_path.join("tendermint").to_path_buf(),
-                db: node_path.join("db").to_path_buf(),
-                ..Default::default()
+            let mut config = Config {
```
tzemanovic, line 862: @Fraccaman I've updated this to use the ledger config with the temp dir, otherwise when running locally, the ledger used in `match_intent` test was picking up state from my local node (the default `.anoma` dir)

shared/src/ledger/pos/vp.rs
```diff
@@ -0,0 +1,334 @@
...
+    type TokenChange = token::Change;
+
+    const POS_ADDRESS: Self::Address = super::ADDRESS;
+
+    fn staking_token_address() -> Self::Address {
+        super::staking_token_address()
+    }
+
+    fn read_pos_params(&self) -> PosParams {
+        let value = self.ctx.read_pre(&params_key()).unwrap().unwrap();
```
yito88, line 246: just a question.
Why does `PosReadOnly` for `PosVP` always read the prior states?

tzemanovic: This is because the trait interface in the PoS crate doesn't fit very nicely with the integration in here (The `PosReadOnly` is also used in transactions integration, which only need to read the current state, not pre/post). We have this follow-up issue to make this nicer https://github.com/anoma/anoma/issues/443

yito88 (approved):
Looks good!

batconjurer (approved):

apps/src/lib/node/ledger/protocol/mod.rs
tzemanovic: opened https://github.com/anoma/anoma/issues/459

shared/src/ledger/pos/vp.rs
tzemanovic: https://github.com/anoma/anoma/issues/459

apps/src/lib/node/ledger/shell.rs
tzemanovic: opened #460

PR #431, opened 2021-09-06
Add `PacketSendData` for sending a packet

yito88:
depends on #327

I want to add `PacketSendData` for a requester to send a packet.
Though a requester didn't know the sequence number (nextSequenceSend), a transaction sending a packet required the packet with the sequence number.


tzemanovic (approved):
LGTM!

PR #432, opened 2021-09-06
Fix the packet validation

yito88:
depends on #327 

- Validate a deleted commitment when PacketAcknowledgement
- Validate a packet in the commitment/receipt validation, instead of each sequence validation, for an unordered channel
  - When the channel is unordered, the sequences may not be updated
- Validate a packet for the port ID, channel ID, and sequence

Updated the design doc: https://github.com/anoma/anoma/pull/283/commits/efadb017dee96c28c944ffa941e265197f65578d


tzemanovic (approved):
LGTM, thanks!

PR #433, opened 2021-09-06
Fix IBC packet timeout validation

yito88:
depends on #327

When the timeout, the corresponding channel should be closed.
The check wasn't enough and the channel validation didn't cover the timeout.

### DIFF
--- a/shared/src/ledger/ibc/channel.rs
+++ b/shared/src/ledger/ibc/channel.rs
@@ -30,6 +30,7 @@ use crate::types::address::{Address, InternalAddress};
 use crate::types::ibc::{
     ChannelCloseConfirmData, ChannelCloseInitData, ChannelOpenAckData,
     ChannelOpenConfirmData, ChannelOpenTryData, Error as IbcDataError,
+    TimeoutData,
 };
 use crate::types::storage::{DbKeySeg, Key, KeySeg};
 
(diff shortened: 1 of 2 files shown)


shared/src/ledger/ibc/channel.rs
```diff
@@ -232,13 +233,18 @@ where
...
                 }
-                match ChannelCloseInitData::try_from_slice(tx_data) {
-                    Ok(_) => Ok(()),
-                    Err(_) => self.verify_channel_close_proof(
-                        port_channel_id,
-                        channel,
-                        tx_data,
-                    ),
+                match TimeoutData::try_from_slice(tx_data) {
+                    Ok(data) => self.validate_commitment_absense(data),
```
tzemanovic, line 237:
```suggestion
                    Ok(data) => self.validate_commitment_absence(data),
```

shared/src/ledger/ibc/channel.rs
```diff
@@ -249,6 +255,30 @@ where
         }
     }
 
+    fn validate_commitment_absense(&self, data: TimeoutData) -> Result<()> {
```
tzemanovic, line 258:
```suggestion
    fn validate_commitment_absence(&self, data: TimeoutData) -> Result<()> {
```

tzemanovic (approved):
LGTM!

PR #447, opened 2021-09-09
fix tx without data causes signature verification crash

### LINKED ISSUE #446: tx without data causes signature verification crash
A tx without data that triggers a VP that tries to verify signature causes a crash in the ledger (because it's calling a host env function to verify it). The missing data needs to be handled properly in the `verify_tx_sig` function, it crashes at:
```shell
Message:  data should exist
Location: shared/src/types/key/ed25519.rs:160
```

tzemanovic:
closes #446

to-do:
- [x] add tests


yito88 (approved):
LGTM!

Issue #392, opened 2021-08-12
Support encrypted transactions


batconjurer:
At present, we only have defined unencrypted transactions. Encrypted transactions will be a new type that will wrap these and contain extra data such as fees, gas limits, and signatures. We need to do the following

 - [x]  Add in the new type / proto definitions of encrypted transactions
 - [x] Add the validation of encrypted transactions, which differs from the validation on the wrapped transaction
 - [x] Change our storage logic to always include validated encrypted transactions regardless of the validity of the wrapped transaction
 - [x] Make a custom event that indicates the success value of applying a transaction after it is decrypted.
 - [x] Make a final decision regarding the policy of how to ensure gas limits for blocks are not exceeded and document this appropriately in the tech specs
 - [x] Cover with tests

cwgoes:
Note that all transactions should be encrypted (once this feature is supported).

PR #454, opened 2021-09-10
Add encrypted tx and wrapper tx types

batconjurer:
This PR resolves the first two tasks in Issue #392. 

 * A WrapperTx type has been added that includes a encrypted tx payload and various metadata for verification.
 * This WrapperTx is included as data in a normal Tx type
 * During the CheckTx call from ABCI, a tx is forwarded to ProcessProposal. If the tx contains a wrapper tx as data, it is validated using the metadata. If validation fails, it is rejected from the mempool
 * Validations include deserializability, sufficient balance to pay fees, ciphertext validity checks from Ferveo, and signatures
 Non-encrypted Txs are still fully supported.


Acentelles (approved):

apps/src/lib/node/ledger/shell.rs
```diff
@@ -495,16 +497,113 @@ impl Shell {
...
+                tx: req.tx,
+            };
+        }
+
+        let (result, processed_tx) = match process_tx(tx.unwrap()) {
+            // This occurs if the wrapper tx signature is invalid
+            Err(err) => (TxResult::from(err), None),
+            Ok(result) => match result {
+                // If it is a raw transaction, we do no further validation
+                TxType::Raw(_) => (
```
tzemanovic, line 528: this case is temporary until we encrypt all the transactions, right?

shared/Cargo.toml
```diff
@@ -9,7 +9,15 @@ version = "0.1.0"
...
-default = []
+default = [
+  "tpke",
+  "ark-bls12-381",
+  "ark-ec",
+  "ark-serialize",
+  "rand_core",
+  "rand_new"]
+
+wasm = []
```
tzemanovic, line 20: this is not used, is it?

shared/src/types/key/ed25519.rs
```diff
@@ -448,6 +448,21 @@ impl From<PublicKey> for PublicKeyHash {
     }
 }
 
+impl From<&PublicKey> for PublicKeyHash {
```
tzemanovic, line 451: we're adding this in #410

apps/src/lib/node/ledger/shell.rs
batconjurer: Yes, in the sense that it is temporary. It allows backwards compatibility at the moment. In the future, we should also have `TxType::Decrypted(_)` and `TxType::UnDecryptable` and `TxType::PVSS` and maybe some others.

shared/Cargo.toml
batconjurer: Oops, that is a relic of an abandoned idea

shared/src/types/key/ed25519.rs
batconjurer: Then I will remove this when I merge in that PR.

tzemanovic (approved):

PR #467, opened 2021-09-15
Add build via Nix

SimSaladin:
Added a `default.nix` that can be used to build the crate(s) in the workspace as Nix derivations.

```bash
# Generate Cargo.nix using crate2nix
nix-shell -p crate2nix --command 'crate2nix generate'

# Build the apps crate
nix-build -A apps

# Run the app
result/bin/anoma --help
```

Two small changes were made to the `apps` and `shared` builders:
- Use symlinks `./proto -> ../proto` instead of references to the parent directory: Nix builds each sub-crate in a sandbox that doesn't have access to the workspace.
- Make the setup of `RUSTFMT` environment variable (and the associated tool chain) optional: the formatting is not useful for Nix builds (only paths in the temporary sandbox environment are modified).


tzemanovic (approved):
I cannot check the nix build yet, as I ran into linker issue installing crate2nix (https://github.com/kolloch/crate2nix/issues/198), but I think the changes look good, thank you! If you could pls add the instructions to build on nix somewhere (maybe in readme installing/development section).

PR #470, opened 2021-09-16
IBC storage key

### LINKED ISSUE #466: Refactoring for IBC storage key
Like PoS storage module, we want to re-use functions for IBC storage keys

yito88:
Closes #466


shared/src/types/storage.rs
```diff
@@ -258,117 +258,31 @@ impl Key {
...
-    }
-
     /// Returns a key from the given DB key path that has the height and
     /// the space type
     pub fn parse_db_key(db_key: &str) -> Result<Self> {
         let mut segments: Vec<&str> =
             db_key.split(KEY_SEGMENT_SEPARATOR).collect();
-        let key = match segments.get(2) {
-            Some(seg)
-                if *seg == Address::Internal(InternalAddress::Ibc).raw() =>
```
yito88, line 332: IBC related keys can be parsed as other keys

yito88, line 332: IBC keys can be parsed as other keys

shared/src/types/storage.rs
```diff
@@ -429,9 +343,6 @@ impl KeySeg for DbKeySeg {
         if string.contains(KEY_SEGMENT_SEPARATOR) {
             return Err(Error::InvalidKeySeg(string));
         }
-        if string == Address::Internal(InternalAddress::Ibc).raw() {
```
yito88, line 432: This is not required because these written data will be validated by IBC VP

shared/src/types/storage.rs
tzemanovic: it's nice not having to check this here!

tzemanovic (approved):
LGTM, thank you!

PR #484, opened 2021-09-20
reduce log levels for e2e tests

tzemanovic:
Hopefully, this should make the e2e tests a bit lighter as there was a lot of debug logging from 3rd party deps. Thank you for the idea @batconjurer !

### DIFF
--- a/apps/src/lib/node/ledger/shims/abcipp_shim.rs
+++ b/apps/src/lib/node/ledger/shims/abcipp_shim.rs
@@ -49,7 +49,6 @@ impl Service<Req> for AbcippShim {
     }
 
     fn call(&mut self, req: Req) -> Self::Future {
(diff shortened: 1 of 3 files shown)


apps/src/lib/node/ledger/shims/abcipp_shim.rs
```diff
@@ -49,7 +49,6 @@ impl Service<Req> for AbcippShim {
     }
 
     fn call(&mut self, req: Req) -> Self::Future {
-        tracing::debug!(?req);
```
tzemanovic, line 52: I removed this for now as some of the requests are big (including tx bytes). If it's still useful, we can maybe just log the request type and other stuff in the cases

batconjurer (approved):

PR #488, opened 2021-09-20
Genesis config

juped:
WIP

### DIFF
--- a/apps/src/lib/config/genesis.rs
+++ b/apps/src/lib/config/genesis.rs
@@ -1,17 +1,426 @@
 //! The parameters used for the chain's genesis
 
 use std::collections::HashMap;
+#[cfg(not(feature = "dev"))]
+use std::path::Path;
 
 use anoma::ledger::parameters::Parameters;
 use anoma::ledger::pos::{GenesisValidator, PosParams};
 use anoma::types::address::Address;
+#[cfg(not(feature = "dev"))]
+use anoma::types::chain::ChainId;
 #[cfg(feature = "dev")]
 use anoma::types::key::ed25519::Keypair;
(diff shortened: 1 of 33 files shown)


juped:
re-marked as draft so no one clicks the button, but this should be good

apps/src/lib/config/genesis.rs
```diff
@@ -1,17 +1,426 @@
...
+        // Public key for validator account. (default: generate)
+        pub account_public_key: Option<HexString>,
+        // Public key for staking reward account. (default: generate)
+        pub staking_reward_public_key: Option<HexString>,
+        // Validator address (default: generate).
+        pub address: Option<String>,
+        // Staking reward account address (default: generate).
+        pub staking_reward_address: Option<String>,
+        // Total number of tokens held at genesis.
+        // XXX: u64 doesn't work with toml-rs!
```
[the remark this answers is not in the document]
tzemanovic: opened https://github.com/anoma/anoma/issues/525. I think maybe for tokens we should define a custom type, since we're using `token::Amount::whole` to construct it, which has a max value `9223372036854` (`(i64::MAX - 1) / 10^6`)

apps/src/lib/node/ledger/tendermint_node.rs
```diff
@@ -40,10 +42,15 @@ pub type Result<T> = std::result::Result<T, Error>;
 
 /// Run the tendermint node.
 pub fn run(
+    // TODO we should just use ledger config replace all the args
```
tzemanovic, line 45: will open follow-up issue to refactor this

tzemanovic: opened https://github.com/anoma/anoma/issues/526

Issue #425, opened 2021-08-31
load genesis config from a file


tzemanovic:

tzemanovic:
let's also remove the conditional compilation for the "dev" feature, which only has a single validator

tzemanovic:
The genesis config should also define the aliases for genesis accounts to be used in the `wallet::defaults` (added in #410 - https://github.com/anoma/anoma/pull/410#discussion_r709332749). 

With "dev" conditional compilation, we also want to keep the wallet default keys, but we should not use these in non-"dev" build (there will be no default keys).

tzemanovic:
Done in #534

PR #494, opened 2021-09-21
faucet account

### LINKED ISSUE #486: add a "faucet" account for testnets
add a "faucet" account with a VP that allows anyone to withdraw limited amount of tokens from it for testnets

tzemanovic:
closes #486

This PR adds a VP for testnet "faucet" account. The account allows anyone to withdraw a limited sum (up to 1k tokens) in a single tx.

This account will be added to testnet's genesis config (pending on #425) with a large sum of tokens and the wallet's default addresses.

The first commit 52109c2 is unrelated change - refactored vp_user to inline the key type match, but applied first because the newly added `vp_testnet_faucet` is based on it.


yito88 (approved):
Looks good!
I suggested the almost same thing as you fixed

PR #496, opened 2021-09-21
on chain validator creation

### LINKED ISSUE #485: PoS - new validator registration
Add a tx and cli cmd to submit it for a new validator registration. This will also need some updates in the validation to allow this.

tzemanovic:
closes #485 

This PR adds a tx to initialize a validator on-chain and updates PoS VP to allow this. Plus some refactoring + setup improvements to allow for testing this.

I've tested it manually this way (e2e tests for this will be added later once we can setup shorter parameters):

```shell
# Modify PoS default parameters to `pipeline_len: 1` and `unbonding_len: 1` to make things faster

# start from fresh
rm -rf .anoma

# start node 0
cargo run --bin anoman ledger

# send a tx to init validator, using .anoma1 as base dir for node 1
cargo run --bin anomac -- --base-dir .anoma1 init-validator --alias new-validator --source Bertha --unsafe-dont-encrypt

# copy config files from node 0 to node 1
cp .anoma/tendermint/config/genesis.json .anoma1/tendermint/config/genesis.json
cp .anoma/tendermint/config/config.toml .anoma1/tendermint/config/config.toml

# update ports in `.anoma1/config.toml` to values different from node 0

# update `persistent_peers` in `.anoma1/tendermint/config/config.toml` to `"{node_0_id}@0.0.0.0:26656"`

# start-up the node 1
cargo run --bin anoman -- --base-dir .anoma1 ledger

# The nodes should connect and stay in sync

# Bond some tokens to the new validator in node 1 (replace the address with the new-validator's address)
cargo run --bin anomac -- bond --source Bertha --validator a1qq5qqqqqg4pyg3j9x3qnjd3cxgmyz3fk8qcrys3hxdp5xwfnx3zyxsj9xgunxsfjg5u5xvzy62f2e5 --amount 199000

# Check voting powers for the follow epoch, e.g.
cargo run --bin anomac voting-power --epoch 1

# Send some transactions to get to the next epoch.
# Validator set should update on new epoch.

# Check that the validator set in tendermint is updated
curl "http://localhost:26657/validators"

# Send more transactions and make sure that the new validator can produce blocks
```

I recommend reviewing by commit as the whole diff is not very clear:
- 385990e36aaaa908f7b0bbd0aceb3fb6b1d69c98: Adds a `tx_init_validator` WASM, a nicer function in the VM env `init_validator` that handles validator and staking reward address creation.
- d03772ffcf7ec3907360492ff8e07be25c71fff8: Adds a CLI cmd for `init_validator` and refactors the tx.rs functions to be methods on `Context` for convenience.
- ee6fbf0e2b919a5518f483b1dc55a91acfec4385: a slightly nicer alias overwrite msg
- bc9e2b8bc90199fd62744b04a6e004734fcd768a: Updates the `become_validator` function in pos crate to also initialize validator's total deltas and voting power fields.
- 67866de696d0ae9f785a6ae7353239b36bebbbf0: Updates the pos validation function to allow on-chain validator creation. The `NewValidator` is used to check that all the required fields are set. We also check that the new validator has been added to the validator set at pipeline offset
- 8fa794e6d9a40cdcb7e52cd8dfa9284e0b3d96c8: Updated PoS docs
- 1a9bad189b4452a85058849f0a2f02940f36a962: Make the cli cmd and args descriptions consistent and fix wrong label for slashes query
- 5f5bf3bfa6477c07766b354b3fec91e05e26fc37: Avoid overwriting the tendermint validator key if it's already set in "dev" build. Also ensures that the directory for validator key and state exist before trying to write the files.
- 42542740c7e2d123eb709f0491f5923ff33646f6: Refactor the rest of the client CLI to be methods on `Context` and move gossip and utils into sub-modules (the diff is not very nice, but there are no changes in the functions implementations other than the `ctx` usage being replaced with `self`
- 918f656ec04198ac9d7fbe5a54e0ed67bd610aee: Add P2P and RPC address to ledger config, write it into tendermint config
- 04ac83718eb5eb12ef1972123e73641054d8580d: Make sure that consensus params are not 0


apps/src/lib/client/tx.rs
```diff
@@ -4,307 +4,511 @@ use std::convert::TryFrom;
...
+                        if validator_address_alias.is_empty() {
+                            println!(
+                                "Empty alias given, using {} as the alias.",
+                                validator_address.encode()
+                            );
+                            validator_address.encode()
+                        } else {
+                            validator_address_alias
+                        };
+                    let added = self.wallet.add_address(
```
batconjurer, line 258: You don't need the extra `added` variable

apps/src/lib/client/tx.rs
```diff
@@ -4,307 +4,511 @@ use std::convert::TryFrom;
...
+                    if added {
+                        println!(
+                            "Added alias {} for address {}.",
+                            validator_address_alias,
+                            validator_address.encode()
+                        );
+                    }
+                    let rewards_address_alias =
+                        format!("{}-rewards", validator_address_alias);
+                    let added = self.wallet.add_address(
```
batconjurer, line 271: again, don't need `added`

batconjurer (commented):
In principle, the code some fine. I'm just unhappy with the scattered `impl Context`s. I just wish there was a way to easily determine what methods Context actually has without a lot of code spelunking .But I haven't thought of a better way yet either

tzemanovic:
> In principle, the code some fine. I'm just unhappy with the scattered `impl Context`s. I just wish there was a way to easily determine what methods Context actually has without a lot of code spelunking .But I haven't thought of a better way yet either

Thank you for reviewing! Agreed, I removed the `impl Context`s back to the way it was

PR #500, opened 2021-09-23
sym-link rust-nightly-version

tzemanovic:
Similar to what was done in #467 for the proto files, this adds a symlink to the `rust-nightly-version` in `apps` and `shared` that use it in their build.rs script. This fixes rust-analyzer which runs from the workspace root and fails to find these from the previous relative path `../rust-nightly-version`.


tzemanovic:
Nvm, this doesn't help. The issue actually seems to be with the path to `.proto` - it looks like the rust-analyzer doesn't follow symlinks

Issue #491, opened 2021-09-20
Fixed address length


tzemanovic:
Fix implicit and internal address length. The bech32m encoded address should all be 80 chars (with `a1` prefix). Currently, implicit addresses are 82 and internal 12

tzemanovic:
done in #502

PR #502, opened 2021-09-24
change addr prefix to "atest1"

### LINKED ISSUE #478: switch to "atest" prefix for addresses
in `shared/src/types/address.rs`: the bech32m address encoding prefix `ADDRESS_HRP` should be "atest" for "dev" and testnets. We should only use "a" for mainnet, once we're there. We'll need to update all the default addresses for this (replace `s/a1/atest1`)

tzemanovic:
closes #478 and #491

I removed borsh encoding usage from address bech32m encoding and replaced with a fixed length utf-8 strings as bytes, so that all the address types are equal length (84 with prefix "atest1", 80 with "a1").

I also switched `make install` to release build without "dev" feature.


shared/src/types/address.rs
```diff
@@ -14,17 +14,41 @@ use sha2::{Digest, Sha256};
...
+// TODO use "a" for live network
+const ADDRESS_HRP: &str = "atest";
 const ADDRESS_BECH32_VARIANT: bech32::Variant = Variant::Bech32m;
 pub(crate) const HASH_LEN: usize = 40;
 
+/// An address string before bech32m encoding must be this size.
+pub const FIXED_LEN_STRING: usize = 45;
+
+/// Raw strings used to produce internal addresses. All the strings must begin
+/// with `PREFIX_INTERNAL` and be `FIXED_LEN_STRING` characters long.
```
juped, line 35: should align these in the source file so it's clear they're the same length

shared/src/types/address.rs
```diff
@@ -14,17 +14,41 @@ use sha2::{Digest, Sha256};
...
+mod internal {
+    pub const POS: &str = "ano::Proof of Stake                          ";
+    pub const POS_SLASH_POOL: &str =
+        "ano::Proof of Stake Slash Pool               ";
```
juped, line 39: (do them all like this one)

shared/src/types/address.rs
```diff
@@ -131,6 +137,73 @@ impl Address {
         }
     }
 
+    /// Convert an address to a fixed length utf-8 string bytes
```
juped, line 140: 7-bit ascii?

shared/src/types/address.rs
```diff
@@ -131,6 +137,73 @@ impl Address {
...
+            }
+            Address::Internal(internal) => match internal {
+                InternalAddress::PoS => internal::POS,
+                InternalAddress::PosSlashPool => internal::POS_SLASH_POOL,
+                InternalAddress::Ibc => internal::IBC,
+                InternalAddress::Parameters => internal::PARAMETERS,
+            }
+            .to_string(),
+        }
+        .into_bytes()
```
juped, line 157: an assertion that it's FIXED_LEN_STRING bytes might be nice

shared/src/types/address.rs
```diff
@@ -14,17 +14,41 @@ use sha2::{Digest, Sha256};
...
 
 /// human-readable part of Bech32m encoded address
-const ADDRESS_HRP: &str = "a";
+// TODO use "a" for live network
+const ADDRESS_HRP: &str = "atest";
 const ADDRESS_BECH32_VARIANT: bech32::Variant = Variant::Bech32m;
 pub(crate) const HASH_LEN: usize = 40;
 
+/// An address string before bech32m encoding must be this size.
+pub const FIXED_LEN_STRING: usize = 45;
```
juped, line 32: FIXED_LEN_STRING_BYTES?

shared/src/types/address.rs
tzemanovic: Good idea, I've added debug assert here. There' a test for this `test_encoded_address_length` that checks all the addresses.

shared/src/types/address.rs
tzemanovic: done

shared/src/types/address.rs
tzemanovic: changed, thanks!

Issue #453, opened 2021-09-10
package default wasms used in the node


tzemanovic:
Related to #451.

The ledger node requires some pre-built VP wasms (vp_token and vp_user) that are used for genesis block initialization. These are currently loaded at runtime from paths relative to the root dir of this repo. 

Because these wasms are stored on-chain, it is important for testnets that the releases use the correct wasms for consensus on the genesis block. The genesis config will be loaded from a file (pending on #425) and so it might be better not to include these in the built binary, but have them accessible from some shared path that is tied to each testnet (maybe we can use the chain-id?), so that different chains can add their own genesis VPs.

Issue #451, opened 2021-09-10
package default wasms used in the client


tzemanovic:
Currently, the client is relying on relative paths to the compiled .wasm files that it's using and therefore has to run from the root dir of this repo. We should package these with the client somehow, so that it can run from any dir.

One way would be to put these into some shared dir on `make install`.

Another would be to `include_bytes!` in the client so that they are part of the built binary, which has the advantage that different version don't share the same dir, but it will increase the client size a bit, but probably not more than 1MB for all the ones that we have.

The wasms that the client uses are listed in `tx.rs` constants.

It would be convenient to also allow to use custom wasm builds for any of the commands that use the pre-builts, provided via optional argument.

PR #503, opened 2021-09-24
wasm dir global arg

tzemanovic:
This PR adds `--wasm-dir` global arg (related to #451 and #453). They also uses `ANOMA_BASE_DIR` and `ANOMA_WASM_DIR` env vars as default when the `base-dir` and/or `wasm-dir` are not specified.

### DIFF
--- a/apps/src/bin/anoma-node/cli.rs
+++ b/apps/src/bin/anoma-node/cli.rs
@@ -10,7 +10,7 @@ pub fn main() -> Result<()> {
     match cmd {
         cli::cmds::AnomaNode::Ledger(sub) => match sub {
             cli::cmds::Ledger::Run(_) => {
-                ledger::run(ctx.config.ledger);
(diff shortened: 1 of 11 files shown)


apps/src/bin/anoma-node/cli.rs
```diff
@@ -10,7 +10,7 @@ pub fn main() -> Result<()> {
     match cmd {
         cli::cmds::AnomaNode::Ledger(sub) => match sub {
             cli::cmds::Ledger::Run(_) => {
-                ledger::run(ctx.config.ledger);
+                ledger::run(ctx.config.ledger, ctx.global_args.wasm_dir);
```
juped, line 13: should this go into the config?

tzemanovic: yeah, good idea, I'll put it in there

tzemanovic: done in be445b8

juped (approved):

Issue #482, opened 2021-09-20
CI run in a PR branch - don't try to merge


tzemanovic:
Currently, the CI that runs in a PR attempts to merge the PR branch in master. This means that if auto-merge fails, the CI will fail too. In some cases auto-merge can also produce a wrong commit in `push-wasm` step, as is a case in this run https://ci.heliax.dev/anoma/anoma/1246/3/6, in which the PR branch was behind master (missing latest merge from master, which also modified WASMs), which created conflicts in the re-built WASMs that then didn't get updated (only the WASMs without conflicts got updated).

It would be preferable not to do this and just run with the branch as is. For this, we will need to update/change how we do the scripts integrity check, which currently only allows using the latest hash values.

Fraccaman:
Done.

PR #505, opened 2021-09-24
wasm_for_tests: Cargo.lock updates

juped:
Missing the hex package (3b5825a8, "add ed sk wrapper and pk/sk hex
encoding") and a different tendermint-rs branch (7b6eb821, "Added
ability for storage to lookup [...]").

### DIFF
--- a/wasm_for_tests/wasm_source/Cargo.lock
+++ b/wasm_for_tests/wasm_source/Cargo.lock
@@ -32,6 +32,7 @@ dependencies = [
  "borsh",
  "chrono",
  "ed25519-dalek",
+ "hex",
  "itertools",
  "loupe",
  "parity-wasm",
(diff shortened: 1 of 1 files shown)


tzemanovic:
Unfortunately, I don't think the CI's auto-merge can currently resolve the checksum conflicts correctly as neither of the hashes will be right, but this problem should go way with #482.

Anyway, I've updated the `wasm_for_tests` lock file in #502, so I think we can close this. I think for this crate we don't need to rebuild in the CI, we can just rebuild locally whenever there are breaking changes as we don't need deterministic checksums (breaking changes will trigger failures in the tests that use these).

juped:
yes, this is superseded, thanks.

to be clear, there's no checksum conflict - it either literally performed the three way merge wrong (very implausible) or checked against the wrong thing somehow.

tzemanovic:
> to be clear, there's no checksum conflict - it either literally performed the three way merge wrong (very implausible) or checked against the wrong thing somehow.

oh I see, it hasn't even changed, that is strange indeed

PR #508, opened 2021-09-27
allow to explicitly set tx signer or signing-key for any tx cmd

tzemanovic:
This PR:
- moves the explicit `--signing-key`/`--signer` arguments from the `cmds::TxCustom` to the common transaction arguments `args::Tx`, to allow to override the default signer for any tx command
- adds a helper function `tx::sign_tx`
- makes `anoma_apps::cli::context` module public
- use ref for context look-ups (owned values are not needed)


yito88 (approved):
LGTM!

PR #514, opened 2021-09-29
shut down with error on ledger crash

### LINKED ISSUE #512: fix ledger crash process exit status
A panic inside ledger's shell is not properly propagated to the process status - it shuts down, but exits with success

tzemanovic:
closes #512


apps/src/lib/node/ledger/tendermint_node.rs
```diff
@@ -141,9 +149,9 @@ fn monitor_process(
     kill_switch: Sender<bool>,
```
yito88, line 149: For consistency, `kill_switch` can be replaced with `abort_sender`

tzemanovic: done

yito88 (approved):
Looks good!

PR #523, opened 2021-10-06
review of user-docs

awasunyin:
Hey @tzemanovic, 

I've reviewed the user-docs this morning. My branch was created from yours `tomas/docs-user-guide` and I made edits. For this release I won't have time for a in-depth developer UX review of the docs, I've focused on things that I found critical.

Here are some notes:

**The main changes I've made are:**
- Renamed the title to Anoma – DOCS
- Edited the introduction
- Added space between emoji + text 
- Took a look at all the pages, made minor edits (e.g. edited excalidraw links)

**Suggestions for the Engineering team:**
- I would suggest to remove the `playnet` section - keeping it would be confusing
- **Not critical for this release**, but I'd recommend the team to look at other user docs for inspiration on the structure. Good examples are (they're not perfect, they all have pros and cons):
  1. Solana: https://docs.solana.com/running-validator
  2. NEAR: https://docs.near.org/docs/develop/basics/getting-started
  3. Polkadot: https://wiki.polkadot.network/docs/getting-started

**Questions:**
- Shouldn't `2.1. Install` go before `2. User Guide`? Atm, the User Guide starts with
```
> This guide assumes that the Anoma binaries are installed and available on path. These are:
[...]
```

### DIFF
--- a/docs/book.toml
+++ b/docs/book.toml
@@ -6,7 +6,7 @@ language = "en"
 multilingual = false
 site-url = "https://anoma.github.io/anoma/"
 src = "src"
-title = "Anoma - Technical Specifications"
+title = "Anoma - DOCS"
 
 [output.html]
 additional-css = ["assets/custom.css"]
--- a/docs/src/README.md
+++ b/docs/src/README.md
@@ -1,4 +1,18 @@
 # Introduction
+Welcome to Anoma's docs!
+
+## About Anoma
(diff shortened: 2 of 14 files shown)


tzemanovic (approved):
Thank you for the review and updates! I'll update the user guide, I've added the install section after, so that's why the wording in the first section doesn't make sense anymore.

Unfortunately, the spaces between emoji in the headers change their links :( it adds an extra dash in the beginning, e.g. `ledger.html#initialize-an-account` becomes `ledger.html#-initialize-an-account`. So there are some internal links that will need to be updated. I've added a linkcheck plugin but it doesn't support links to headings yet :(

I'll remove the playnet page too, thank you for the examples!

PR #551, opened 2021-10-11
config refactor + add tendermint consensus.timeout_commit config

### LINKED ISSUE #526: refactor tendermint_node run fn
Refactor `tendermint_node::run` fn to use a config for all the values passed to it. The variables for Tendermint could be split into another struct `config::Tendermint` to be used inside the `config::Ledger`, e.g.:
```
pub struct Tendermint {
    pub tendermint: PathBuf,
    pub rpc_address: SocketAddr,
    pub p2p_address: SocketAddr,
    pub p2p_persistent_peers: Vec<tendermint::net::Address>,
    pub p2p_pex: bool,
}
```

_Originally posted by @tzemanovic in https://github.com/anoma/anoma/pull/488#r723990497_

tzemanovic:
closes #526 

- 083e073: only re-ordering definitions in the config mod
- 28ece77: organize config fields into sub-structs
- 4f3d0cf: added `consensus.timeout_commit` config, we'll want to use this for testnet
- 19f3675: removed `rpc.timeout_broadcast_tx_commit` config, we're no longer using this to submit txs, instead we subscribe to application event via websocket

### DIFF
--- a/apps/src/lib/config/mod.rs
+++ b/apps/src/lib/config/mod.rs
@@ -292,7 +292,7 @@ impl Tendermint {
             ),
             p2p_persistent_peers,
             p2p_pex: true,
-            consensus_timeout_commit: tendermint::Timeout::from_str("10s")
+            consensus_timeout_commit: tendermint::Timeout::from_str("1s")
                 .unwrap(),
         }
     }


juped:
this e2e test failure looks like it might be a real one (error serializing to toml)?

PR #556, opened 2021-10-11
build: add targets to build release with docker

juped:
Add Makefile targets to build a release or a package via the
anoma-build Docker image defined in the root.


juped:
pls update ci

tzemanovic (approved):

PR #568, opened 2021-10-15
cleanup: don't commit protobuf build artifacts (reopened #389)

juped:
The contents of apps/src/lib/proto/generated and
shared/src/proto/generated are build artifacts. Remove them from the
repository, and add .gitignore files serving a dual purpose: keeping
the directories in the repository, like a .keep file, and ignoring
the generated artifacts (Rust sources).


Fraccaman:
pls restart ci

PR #615, opened 2021-11-03
add conditional compilation on rlimit for unix only

tzemanovic:
on non-unix platforms, rlimit shouldn't be used otherwise it fails to compile with e.g.:

```
   Compiling anoma_apps v0.2.0 (C:\dev\heliax\anoma\apps)
error[E0432]: unresolved imports `rlimit::Resource`, `rlimit::Rlim`
  --> apps\src/lib\node\ledger\storage\rocksdb.rs:26:14
```

Thanks for report @karbyshev !

### DIFF
--- a/apps/src/lib/node/ledger/storage/rocksdb.rs
+++ b/apps/src/lib/node/ledger/storage/rocksdb.rs
@@ -14,17 +14,14 @@
 //!   - `subspace`: any byte data associated with accounts
 //!   - `address_gen`: established address generator
 
-use std::cmp::{min, Ordering};
+use std::cmp::Ordering;
 use std::collections::HashMap;
-use std::convert::TryInto;
-use std::io;
 use std::path::Path;
 
 use anoma::ledger::storage::types::PrefixIterator;
 use anoma::ledger::storage::{types, BlockState, DBIter, Error, Result, DB};
 use anoma::types::storage::{BlockHeight, Key, KeySeg, KEY_SEGMENT_SEPARATOR};
 use anoma::types::time::DateTimeUtc;
-use rlimit::{Resource, Rlim};
 use rocksdb::{
     BlockBasedOptions, Direction, FlushOptions, IteratorMode, Options,
     ReadOptions, SliceTransform, WriteBatch, WriteOptions,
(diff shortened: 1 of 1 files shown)


yito88 (approved):
Looks good! Thanks!

apps/src/lib/node/ledger/storage/rocksdb.rs
```diff
@@ -14,16 +14,15 @@
 //!   - `subspace`: any byte data associated with accounts
 //!   - `address_gen`: established address generator
 
-use std::cmp::{min, Ordering};
+use std::cmp::Ordering;
 use std::collections::HashMap;
-use std::convert::TryInto;
```
yito88, line 19: Is it better to add `#[cfg(unix)]` here?

apps/src/lib/node/ledger/storage/rocksdb.rs
```diff
@@ -37,24 +36,31 @@ pub struct RocksDB(rocksdb::DB);
...
+
     let mut cf_opts = Options::default();
     // ! recommended initial setup https://github.com/facebook/rocksdb/wiki/Setup-Options-and-Basic-Tuning#other-general-options
     cf_opts.set_level_compaction_dynamic_level_bytes(true);
     // compactions + flushes
     cf_opts.set_max_background_jobs(6);
     cf_opts.set_bytes_per_sync(1048576);
+
+    #[cfg(unix)]
+    use std::convert::TryInto;
```
yito88, line 56: Is it better to put `use` at the top of the file?

karbyshev: @yito88 I personally have no preferences (since I am new in rust). Is the preferred coding style to always put imports on top of the file?

tzemanovic: Looking back at the changes, I think it may actually be nicer to move the unix only logic into an inlined sub-module, so we don't have the `#[cfg(unix)]` in so many places. I'll push another commit for it

tzemanovic: updated in e284c9b

juped:
rebased and rerunning, pls check

karbyshev:
Looks good.

karbyshev (approved):

PR #617, opened 2021-11-03
init-network set open P2P if not localhost

tzemanovic:
this is a better default config for testnets as we want to enable P2P connections between users


juped:
force pushed this for our release candidate

PR #616, opened 2021-11-03
init-network create archive for anoma-network-config release

tzemanovic:
This PR adds tar.gz creation in `init-network` cmd that can be used to make a release for the anoma-network-config repo using a script from https://github.com/heliaxdev/anoma-network-config/pull/4. The command to `join-network` has been updated to then use the GH releases in https://github.com/anoma/anoma/pull/619

PR #601, opened 2021-10-29
[apps/client/utils] join-network: get configs from the whole base dir

tzemanovic:

PR #619, opened 2021-11-03
join network from anoma-network-config GH releases

tzemanovic:
replaces #601. Depends on #616 and https://github.com/heliaxdev/anoma-network-config/pull/4


juped:
force pushed this for our release candidate

PR #622, opened 2021-11-04
Make all txs encrypted, add in ABCI++, support commit and reveal scheme for txs in a block

### LINKED ISSUE #392: Support encrypted transactions

batconjurer:
This is not a feature complete PR for either ABCI++ or our TPKE protocol. This work closes #392, #630.

These feature is gated to support backwards compatibility with the current master branch.


batconjurer:
pls update wasm

apps/Cargo.toml
```diff
@@ -44,9 +44,25 @@ dev = ["anoma/dev"]
 std = ["ed25519-dalek/std", "rand/std", "rand_core/std"]
 # for integration tests and test utilies
 testing = ["dev"]
+ABCI = [
```
tzemanovic, line 47: I think we should add this one to default features for now

batconjurer: I tried that but ran into a lot of trouble when trying to enable the ABCI++ feature. Both would get turned on

batconjurer:
In order to run shell unit tests in parallel, we should use tmp directories so that arenot have multiple threads competing for the rocksdb file lock.

tzemanovic:
pls update ci

apps/src/lib/client/tendermint_websocket_client.rs
```diff
@@ -207,7 +232,7 @@ impl TendermintWebsocketClient {
                 subscribed: None,
                 received_responses: Arc::new(Mutex::new(HashMap::new())),
                 connection_timeout: connection_timeout
-                    .unwrap_or_else(|| Duration::new(300, 0)),
+                    .unwrap_or_else(|| Duration::new(30, 0)),
```
tzemanovic, line 235:
```suggestion
                    .unwrap_or_else(|| Duration::new(300, 0)),
```

apps/src/lib/client/utils.rs
```diff
@@ -406,8 +416,18 @@ pub fn init_network(
             // directories.
             config.ledger.shell.base_dir = config::DEFAULT_BASE_DIR.into();
             // Add a ledger P2P persistent peers
-            config.ledger.tendermint.p2p_persistent_peers =
-                persistent_peers.clone();
+            // TODO: Reviewers! Is there a reason I shouldn't do this?
```
tzemanovic, line 419:
```suggestion
```
:(

apps/src/lib/client/utils.rs
```diff
@@ -406,8 +416,18 @@ pub fn init_network(
...
+            config.ledger.tendermint.p2p_persistent_peers = persistent_peers
+                    .iter()
+                    .enumerate()
+                    .filter_map(|(index, peer)|
+                        // we do not the validator in its own persistent peer list
```
tzemanovic, line 424:
```suggestion
                        // we do not add the validator in its own persistent peer list
```

apps/src/lib/client/utils.rs
tzemanovic: yeah, it's strange, not sure what's changed

apps/src/lib/node/ledger/protocol/mod.rs
```diff
@@ -83,20 +82,26 @@ impl Default for VpsResult {
...
-        .add_base_transaction_fee(tx_bytes.len())
+        .add_base_transaction_fee(tx_length)
         .map_err(Error::GasError)?;
-
-    let tx = Tx::try_from(tx_bytes).map_err(Error::TxDecodingError)?;
-
-    match process_tx(tx).unwrap() {
-        TxType::Raw(tx) => {
+    match tx {
+        TxType::Raw(_) => unreachable!(),
```
tzemanovic, line 103: let's add an error for this

apps/src/lib/node/ledger/shell/process_proposal.rs
```diff
@@ -0,0 +1,588 @@
...
+    pub fn process_and_decode_proposal(
+        &mut self,
+        req: shim::request::ProcessProposal,
+    ) -> shim::response::ProcessProposal {
+        // check the wrapper tx
+        match process_tx(
+            Tx::try_from(req.tx.as_ref())
+                .expect("Deserializing tx should not fail"),
+        ) {
+            Ok(TxType::Wrapper(_)) => {}
```
tzemanovic, line 137: the data in tx types should always be borsh serialized TxType enum

apps/src/lib/node/ledger/shell/mod.rs
```diff
@@ -0,0 +1,545 @@
...
+    /// modifications for testing purposes
+    pub(super) struct TestShell {
+        pub shell: Shell,
+    }
+
+    impl TestShell {
+        /// Create a new shell
+        pub fn new() -> Self {
+            Self {
+                shell: Shell::new(
```
tzemanovic, line 464: let's paramaterize the Shell over the storage and use MockDB for this

apps/src/lib/node/ledger/shell/finalize_block.rs
```diff
@@ -0,0 +1,898 @@
...
+    ///   2: Invalid order of decrypted txs
+    ///   3. More decrypted txs than expected
+    ///   4. Runtime error in WASM
+    pub fn finalize_block(
+        &mut self,
+        req: shim::request::FinalizeBlock,
+    ) -> Result<shim::response::FinalizeBlock> {
+        let mut response = shim::response::FinalizeBlock::default();
+        // begin the next block and check if a new epoch began
+        let (height, new_epoch) = if cfg!(test) {
```
tzemanovic, line 47: this condition could be surprising!

apps/src/lib/node/ledger/shell/mod.rs
```diff
@@ -0,0 +1,545 @@
...
+            base_dir,
+            wasm_dir,
+            next_wrapper: 0,
+        }
+    }
+
+    /// Iterate lazily over the wrapper txs in order
+    #[cfg(not(feature = "ABCI"))]
+    fn next_wrapper(&mut self) -> Option<&WrapperTx> {
+        if self.next_wrapper == self.storage.wrapper_txs.len() {
```
tzemanovic, line 151:
```suggestion
        if self.next_wrapper >= self.storage.wrapper_txs.len() {
```

apps/Cargo.toml
tzemanovic: Unfortunately, this doesn't work nicely with cargo workspace :(

apps/src/lib/node/ledger/tendermint_node.rs
```diff
@@ -35,10 +41,34 @@ pub enum Error {
...
+                    &path
+                );
+                Ok(path)
+            }
+            Err(msg) => Err(Error::TendermintPath(msg)),
+        }
+    } else {
+        Ok(String::from("tendermint"))
+    }
+}
```
tzemanovic, line 70:
```suggestion
/// Returns an error if the env var is defined but not a valid Unicode.
fn from_env_or_default() -> Result<String> {
    match std::env::var("TENDERMINT") {
        Ok(path) => {
            tracing::info!("Using tendermint path from env variable: {}", path);
            Ok(path)
        }
        Err(std::env::VarError::NotPresent) => Ok(String::from("tendermint")),
        Err(std::env::VarError::NotUnicode(msg)) => {
            Err(Error::TendermintPath(msg))
        }
    }
}
```

Makefile
```diff
@@ -60,12 +63,20 @@ check:
 	make -C $(wasms) check && \
```
tzemanovic, line 63: on line 62 (I cannot comment on it), we should also add `--features ABCI` to the `cargo check`

apps/src/lib/node/ledger/tendermint_node.rs
```diff
@@ -91,16 +123,31 @@ pub fn run(
...
-            "--home",
-            &home_dir_string,
-        ])
-        .spawn()
-        .map_err(Error::StartUp)?;
+    let tendermint_node = if !cfg!(feature = "ABCI") {
+        Command::new(&tendermint_path)
+            .args(&[
+                "start",
+                "--mode",
```
tzemanovic, line 130: Same here

tzemanovic: nvm, this is already guarded

apps/src/lib/node/ledger/tendermint_node.rs
```diff
@@ -249,7 +297,7 @@ fn update_tendermint_config(
 
     // In "dev", only produce blocks when there are txs or when the AppHash
     // changes
-    config.consensus.create_empty_blocks = !cfg!(feature = "dev");
+    config.consensus.create_empty_blocks = true; // !cfg!(feature = "dev");
```
tzemanovic, line 300: maybe we should remove the explicit update, so we can change this in tendermint config, or even better add this to our ledger config so we can change it without re-compiling

shared/src/ledger/storage/mod.rs
```diff
@@ -27,6 +33,11 @@ use crate::types::storage::{
     BlockHash, BlockHeight, DbKeySeg, Epoch, Epochs, Key, BLOCK_HASH_LENGTH,
 };
 use crate::types::time::DateTimeUtc;
+#[cfg(feature = "ferveo-tpke")]
+use crate::types::transaction::WrapperTx;
+// TODO: What a miserable hack. But it's soooo much less typing
+#[cfg(not(feature = "ferveo-tpke"))]
+type WrapperTx = ();
```
tzemanovic, line 40: could we not use the conditional compilation on the `pub wrapper_txs` fields instead?

shared/src/vm/wasm/run.rs
```diff
@@ -136,9 +135,16 @@ where
...
             error,
         })?;
-    apply_tx
+    match apply_tx
         .call(tx_data_ptr, tx_data_len)
-        .map_err(Error::RuntimeError)?;
+        .map_err(Error::RuntimeError)
+    {
+        Err(Error::RuntimeError(err)) => {
+            tracing::info!("{:?}", err.trace());
```
tzemanovic, line 143:
```suggestion
            tracing::debug!("Tx WASM failed with {}", err);
```

tests/src/e2e/gossip_tests.rs
```diff
@@ -154,7 +154,7 @@ fn match_intents() -> Result<()> {
             "--topic",
             "asset_v1",
             "--signing-key",
-            BERTHA_KEY,
+            ALBERT_KEY,
```
tzemanovic, line 157: weird, this shouldn't work

tests/src/e2e/setup.rs
```diff
@@ -440,10 +459,21 @@ where
...
-    let cmd = CargoBuild::new()
-        .package(APPS_PACKAGE)
-        .manifest_path(manifest_path)
-        .bin(bin_name);
+    let cmd = if !cfg!(feature = "ABCI") {
+        CargoBuild::new()
+            .package(APPS_PACKAGE)
+            .manifest_path(manifest_path)
+            .no_default_features()
+            .features("std ABCI-plus-plus")
```
tzemanovic, line 467:
```suggestion
            .features("ABCI-plus-plus")
```

tests/src/e2e/setup.rs
```diff
@@ -440,10 +459,21 @@ where
...
+            .bin(bin_name)
+    } else {
+        CargoBuild::new()
+            .package(APPS_PACKAGE)
+            .manifest_path(manifest_path)
+            .no_default_features()
+            .features("std ABCI")
```
tzemanovic, line 474:
```suggestion
            .features("ABCI")
```

apps/src/lib/node/ledger/shims/abcipp_shim.rs
```diff
@@ -56,46 +62,64 @@ impl Service<Req> for AbcippShim {
...
+                        deliver_tx.tx.clone().into(),
+                        #[cfg(feature = "ABCI")]
+                        deliver_tx.tx.into(),
+                    ))
                     .map_err(Error::from)
                     .and_then(|res| match res {
-                        Response::PrepareProposal(resp) => {
-                            Ok(Resp::BeginBlock(resp.into()))
+                        Response::ProcessProposal(resp) => {
+                            tracing::info!("{:?}", resp);
```
tzemanovic, line 86: Let's remove this, the tx bytes is spammy and we print the tx result elsewhere
```suggestion
```

tests/src/e2e/ledger_tests.rs
```diff
@@ -247,48 +293,59 @@ fn invalid_transactions() -> Result<()> {
...
     ledger.exp_string("Anoma ledger node started")?;
-    ledger.exp_string("Started node")?;
+    if !cfg!(feature = "ABCI") {
+        ledger.exp_string("started node")?;
+    } else {
+        ledger.exp_string("Started node")?;
+    }
     // Wait to commit a block
-    ledger.exp_regex(r"Committed block hash.*, height: [0-9]+")?;
+    // ledger.exp_regex(r"Committed block hash.*, height: [0-9]+")?;
```
tzemanovic, line 305: Does this still work?
```suggestion
    ledger.exp_regex(r"Committed block hash.*, height: [0-9]+")?;
```

batconjurer:
pls update wasm

shared/src/ledger/storage/mod.rs
batconjurer: As per our previous conversation, I think I am not going to persist the wrappers in the storage at all. Instead, they will be flushed to a separate file on ledger shutdown/crash. This will remove that line

tzemanovic:
pls update ci

tzemanovic:
pls update ci

tzemanovic:
pls update wasm

apps/src/lib/node/ledger/shell/finalize_block.rs
```diff
@@ -0,0 +1,940 @@
...
+        for tx in &req.txs {
+            if ErrorCodes::from_u32(tx.result.code).unwrap()
+                == ErrorCodes::InvalidSig
+            {
+                let wrapper =
+                    WrapperTx::try_from(&Tx::try_from(tx.tx.as_ref()).unwrap())
+                        .unwrap();
+                let mut tx_result =
+                    Event::new_tx_event(&TxType::Wrapper(wrapper), height.0);
+                tx_result["code"] = "2".into();
```
tzemanovic, line 64:
```suggestion
                tx_result["code"] = tx.result.code.to_string();
```

tzemanovic (approved):
LGTM!

PR #609, opened 2021-11-02
replace the `anoma` process with the sub-command

tzemanovic:
- 1st commit, related to #592 - `anoma` sub-commands replace the current process
- 2nd commit, closes #583

PR #623, opened 2021-11-04
Handle termination signal on Windows

karbyshev:
Implemented in pair with @batconjurer

### DIFF
--- a/apps/Cargo.toml
+++ b/apps/Cargo.toml
@@ -94,6 +94,7 @@ tendermint-proto = {git = "https://github.com/heliaxdev/tendermint-rs", branch =
 tendermint-rpc = {git = "https://github.com/heliaxdev/tendermint-rs", branch = "tomas/lowercase-node-id", features = ["http-client", "websocket-client"]}
 thiserror = "1.0.24"
 tokio = {version = "1.8.2", features = ["full"]}
+tokio-test = "0.4.2"
 toml = "0.5.8"
 tonic = "0.5.0"
 tower = "0.4"
@@ -102,6 +103,7 @@ tower-abci = {git = "https://github.com/heliaxdev/tower-abci", branch = "tomas/t
 tracing = "0.1.26"
 tracing-log = "0.1.2"
 tracing-subscriber = "0.2.18"
+winapi = "0.3.9"
 websocket = "0.26.2"
 
 [dev-dependencies]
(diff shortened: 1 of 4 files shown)


tzemanovic:
Thanks for the fixes! In #609 we're removing the signal hook and replacing it windows compatible code in `apps/src/bin/anoma/cli.rs`, so no changes should be necessary. If you could, pls rebase on top of it. 

Furthermore, I think we should get rid of the signal hook in `tendermint_node.rs` too and just use the `tokio::signal::ctrl_c`, which is portable.

apps/Cargo.toml
```diff
@@ -94,6 +94,7 @@ tendermint-proto = {git = "https://github.com/heliaxdev/tendermint-rs", branch =
 tendermint-rpc = {git = "https://github.com/heliaxdev/tendermint-rs", branch = "tomas/lowercase-node-id", features = ["http-client", "websocket-client"]}
 thiserror = "1.0.24"
 tokio = {version = "1.8.2", features = ["full"]}
+tokio-test = "0.4.2"
```
tzemanovic, line 97: We shouldn't use this crate in non-dev dependencies. We'll need to replace its usage too (more below)

apps/src/lib/node/ledger/tendermint_node.rs
```diff
@@ -140,6 +155,25 @@ fn kill_on_term_signal(abort_sender: Sender<bool>) {
     });
 }
 
+#[cfg(target_os = "windows")]
+fn kill_on_term_signal(abort_sender: Sender<bool>) {
+    use tokio::signal::ctrl_c;
+    let _ = std::thread::spawn(|| {
+        tokio_test::block_on(async move {
```
tzemanovic, line 162: `tokio_test::block_on` creates a new runtime, we already have one in the shell on `run_shell`. I think we'll want to refactor the ledger to have the main entrypoint async so that this function can be async too

tzemanovic:
@karbyshev thank you for starting these changes! We've refactored the ledger's async usage and switched to use the tokio ctrl_c handler everywhere. With changes from #684, the compilation now works on Windows (sadly not yet e2e tests, but we have #686 open to fix that too). If you could, pls try to run a local node once #684 is merged to see if that works on windows now.

tzemanovic:
replaced by #658 and #684

PR #629, opened 2021-11-05
[build] avoid build context in docker image builds

SimSaladin:
This makes the image build steps much faster by avoiding sending (completely unnecessary) build contexts to the docker daemon. Especially useful if there are build artifacts in the current directory (e.g. in `target`).


juped:
pls update ci

juped:
this should be pretty helpful

PR #647, opened 2021-11-09
[tests/e2e] the env var should be set on build instead of run

tzemanovic:
This fixes an issue of `ANOMA_DEV=true` leaking into the e2e tests, which should always run *without* the dev feature


tzemanovic:
pls restart ci

tzemanovic:
already fixed on master via #622

PR #650, opened 2021-11-11
Tx data contains tx type

batconjurer:
We support various kinds of tx types: Raw, Decrypted, and Wrapper. Previously, we would have to try and deserialize a Tx's data into each of these different types to figure out which type it was. With this PR, we wrap all transactions in TxType enum and serialize that instead. This way, we can always deserialize and immediately know the type.

### DIFF
--- a/apps/src/lib/node/ledger/shell/finalize_block.rs
+++ b/apps/src/lib/node/ledger/shell/finalize_block.rs
@@ -125,6 +125,9 @@ where
                     }
                     Event::new_tx_event(&processed_tx, height.0)
                 }
+                TxType::Protocol(protocol_tx) => {
+                    todo!()
+                }
                 TxType::Raw(_) => unreachable!(),
             };
 
@@ -189,6 +192,7 @@ where
 
         if new_epoch {
             self.update_epoch(&mut response);
+            self.update_dkg();
         }
 
         response.gas_used = self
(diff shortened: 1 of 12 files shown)


apps/src/lib/node/ledger/shell/finalize_block.rs
```diff
@@ -56,9 +56,13 @@ where
             if ErrorCodes::from_u32(tx.result.code).unwrap()
                 == ErrorCodes::InvalidSig
             {
-                let wrapper =
-                    WrapperTx::try_from(&Tx::try_from(tx.tx.as_ref()).unwrap())
-                        .unwrap();
+                let wrapper = if let Ok(TxType::Wrapper(wrapper)) =
+                    TxType::try_from(Tx::try_from(tx.tx.as_ref()).unwrap())
```
tzemanovic, line 60: I'm not sure if this unwrap is safe, in `shell/process_proposal`, it looks like there are many cases where `TxResult` is converted to `Response::ProcessProposal` with default (empty) tx bytes

tzemanovic (approved):
besides the one comment, LGTM!

apps/src/lib/node/ledger/shell/finalize_block.rs
batconjurer: The orignal tx is saved in the shim. FinalizeBlock sees the exact same tx as ProcessProposal

batconjurer: See abcipp_shim.rs lines 87-94

PR #652, opened 2021-11-12
Genesis source: allow to use aliases for token balances

tzemanovic:
Balances of tokens can now refer to an alias of any type of account or for an implicit account derived from a public key of an established account or a validator account (using syntax `"{alias}.public_key" = 100`.


juped:
This is good syntax (shame about the quotes, but not a big deal). Will look at the code shortly.

PR #653, opened 2021-11-16
Tx data contains tx type

batconjurer:
We support various kinds of tx types: Raw, Decrypted, and Wrapper. Previously, we would have to try and deserialize a Tx's data into each of these different types to figure out which type it was. With this PR, we wrap all transactions in TxType enum and serialize that instead. This way, we can always deserialize and immediately know the type.

### DIFF
--- a/apps/src/lib/node/ledger/shell/process_proposal.rs
+++ b/apps/src/lib/node/ledger/shell/process_proposal.rs
@@ -167,17 +167,16 @@ where
         match process_tx(req_tx.clone()) {
             Ok(TxType::Wrapper(_)) => {}
             Ok(_) => {
-                let mut resp: shim::response::ProcessProposal =
-                    shim::response::TxResult {
+                return shim::response::ProcessProposal {
+                    result: shim::response::TxResult {
                         code: ErrorCodes::InvalidTx.into(),
                         info: "Transaction rejected: Non-encrypted \
                                transactions are not supported"
                             .into(),
-                    }
-                    .into();
-                // this ensures that emitted events are of the correct type
-                resp.tx = req.tx;
-                return resp;
+                    },
+                    // this ensures that emitted events are of the correct type
+                    resp.tx = req.tx,
+                }
             }
             Err(_) => {
                 // This will be caught later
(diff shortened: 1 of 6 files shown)


apps/src/lib/node/ledger/shell/process_proposal.rs
```diff
@@ -140,17 +149,37 @@ where
...
+                        code: ErrorCodes::InvalidTx.into(),
+                        info: "The submitted transaction was not \
+                               deserializable"
+                            .into(),
+                    }
+                    .into();
+
+                // this ensures that emitted events are of the correct type
+                resp.tx = req.tx;
+                return resp;
```
tzemanovic, line 166:
```suggestion
                shim::response::ProcessProposal {
                    result: shim::response::TxResult {
                        code: ErrorCodes::InvalidTx.into(),
                        info: "The submitted transaction was not \
                               deserializable"
                            .into(),
                    },
                    // this ensures that emitted events are of the correct type
                    tx: req.tx
                }
```

apps/src/lib/node/ledger/shell/process_proposal.rs
```diff
@@ -140,17 +149,37 @@ where
...
+                    shim::response::TxResult {
+                        code: ErrorCodes::InvalidTx.into(),
+                        info: "Transaction rejected: Non-encrypted \
+                               transactions are not supported"
+                            .into(),
+                    }
+                    .into();
+                // this ensures that emitted events are of the correct type
+                resp.tx = req.tx;
+                return resp;
```
tzemanovic, line 182:
```suggestion
                shim::response::ProcessProposal {
                    result: shim::response::TxResult {
                        code: ErrorCodes::InvalidTx.into(),
                        info: "Transaction rejected: Non-encrypted \
                               transactions are not supported"
                            .into(),
                    },
                    // this ensures that emitted events are of the correct type
                    resp.tx = req.tx,
                }
```

tzemanovic (approved):
Just 2 nit suggestions. Otherwise LGTM!

apps/src/lib/node/ledger/shell/process_proposal.rs
tzemanovic: oh, I'm missing the explicit return

PR #618, opened 2021-11-03
Update ibc-rs

yito88:
Closes #607 

- Update the toolchain from 1.54.0 to 1.56.1
- Update `prost` to 0.9
- Update
    - [tendermint-rs (ABCI++)](https://github.com/heliaxdev/tendermint-rs/tree/yuji/abci++-v0.23.0) and [tendermint-rs](https://github.com/heliaxdev/tendermint-rs/tree/yuji/rebase_v0.23.0) based on tendermint-rs 0.23.0
    - [tower-abci (ABCI++)](https://github.com/heliaxdev/tower-abci/tree/yuji/abci++-v0.23.0) and [tower-abci](https://github.com/heliaxdev/tower-abci/tree/yuji/rebase_v0.23.0)
    - [ibc-rs (ABCI++)](https://github.com/heliaxdev/ibc-rs/tree/yuji/abci++-v0.23.0) and  [ibc-rs](https://github.com/heliaxdev/ibc-rs/tree/yuji/rebase_v0.23.0)
    - [sparse-merkle-tree](https://github.com/heliaxdev/sparse-merkle-tree/tree/yuji/prost-0.9)
- Modified IBC VP and types for the latest IBC module handlers (`*Reader` returns `Result` instead of `Option`)
    - I'd like to refactor some functions in another PR
- Small fixes for clippy

PR #656, opened 2021-11-17
Emit an IBC event

### LINKED ISSUE #480: Emit IBC events for IBC relayer
Set IBC-related event type and attributes to `TxResult` and set `Event` for IBC relayer to get IBC events from Anoma ledger
- IBC event type
  - https://github.com/informalsystems/ibc-rs/blob/master/modules/src/ics02_client/events.rs
  - https://github.com/informalsystems/ibc-rs/blob/master/modules/src/ics03_connection/events.rs
  - https://github.com/informalsystems/ibc-rs/blob/master/modules/src/ics04_channel/events.rs 
- IBC-related attributes
  - e.g. for creating a client:
    - client ID: String
    - client type: String
    - height: String

yito88:
Closes #480 (Blocked by #618)

- Store IBC event to the write log
- Merge IBC event's attributes to the tx result
- Validate the IBC event in IBC VP
- Changed the IBC directories


yito88:
pls update wasm

apps/src/lib/node/ledger/events.rs
```diff
@@ -32,6 +35,7 @@ impl Display for EventType {
         match self {
             EventType::Accepted => write!(f, "accepted"),
             EventType::Applied => write!(f, "applied"),
+            EventType::Ibc(t) => write!(f, "{}", t),
```
tzemanovic, line 38: I think we should prefix the IBC events with e.g.:
```suggestion
            EventType::Ibc(t) => write!(f, "ibc:{}", t),
```
because Tendermint merges all the events in a block into one

(same on line 50)

shared/src/ledger/storage/write_log.rs
```diff
@@ -7,6 +7,7 @@ use thiserror::Error;
```
tzemanovic, line 7: regarding the changes in this module:
- can there only be up to 1 IBC event per tx or should be it possible to have multiple events? Maybe it could be a vec
- I think it would be better to separate the IBC event from `StorageModification` - we could add a new field in the `WriteLog` instead (the key is constant, so we don't need to use it at all and I don't think we should `#[allow(clippy::large_enum_variant)]` on `StorageModification` as this is on a hot path)

shared/src/ledger/storage/write_log.rs
```diff
@@ -204,6 +241,17 @@ impl WriteLog {
             .collect()
     }
 
+    /// Get the IBC event of the current transaction
+    pub fn get_ibc_event(&self) -> Option<IbcEvent> {
```
tzemanovic, line 245: I think this can return `Option<&IbcEvent>` to avoid cloning (when you need an owned value for the `TxResult`, you can call `.cloned()` on it)

shared/src/ledger/storage/write_log.rs
yito88: > can there only be up to 1 IBC event per tx or should be it possible to have multiple events?
only 1 IBC event per tx

> I think it would be better to separate the IBC event from `StorageModification`
That makes sense. Thanks!

apps/src/lib/node/ledger/events.rs
yito88: I'm not sure how Tendermint merges them. Do the event type strings are merged?
The event type string is parsed to convert a Tendermint event to `IbcEvent` in ibc-rs.
for example, https://github.com/informalsystems/ibc-rs/blob/ae364f9e93d055e63f49ac457f0acf6df110c62c/modules/src/core/ics02_client/events.rs#L31
from_str(): https://github.com/informalsystems/ibc-rs/blob/ae364f9e93d055e63f49ac457f0acf6df110c62c/modules/src/events.rs#L178

tzemanovic: as discussed the IBC relayer expects this format, so let's keep it as is

yito88:
pls restart ci

Fraccaman:
Request failed with status code 400

vm_env/src/imports.rs
```diff
@@ -173,6 +174,14 @@ pub mod tx {
             .expect("Decoding address created by the ledger shouldn't fail")
     }
 
+    // Emit an IBC event
```
tzemanovic, line 177:
```suggestion
    /// Emit an IBC event. There can be only one event per transaction. On
    /// multiple calls, only the last emitted event will be used.
```

tzemanovic: I force pushed a fixup

tzemanovic (approved):
LGTM, thanks!

tzemanovic:
pls update wasm

PR #664, opened 2021-11-22
Renamed all abci++ deps to abcipp for nix and giggles

batconjurer:
Renamed abci++ branches and deps to abcipp so that nix can properly parse urls. Also, some 3rd party updates had deprecated crap that needed updating to appease clippy.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4,11 +4,11 @@ version = 3
 
 [[package]]
 name = "addr2line"
-version = "0.16.0"
+version = "0.17.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "3e61f2b7f93d2c7d2b08263acaa4a363b3e276806c68af6134c44f523bf1aacd"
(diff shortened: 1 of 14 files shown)


SimSaladin (approved):

batconjurer:
pls update ci

batconjurer:
pls update wasm

batconjurer:
pls update wasm

batconjurer:
pls update wasms

SimSaladin:
This is fixes the `Cargo.nix` generation via `crate2nix`. There's still a minor issue that `crate2nix` doesn't handle the `tendermint-proto`/`tendermint-proto-abci` (and other packages) renaming correctly behind feature flags (it uses the same source for both when it should use different branches).

This can be worked around by choosing one of `ABCI`, `ABCI-plus-plus` at generation time:

```bash
crate2nix generate --no-default-features --features "std ABCI"
```

And then building with the same flags:

```nix
cargo_nix.workspaceMembers.anoma_apps.build.override { features = ["std" "ABCI"]; }
```

PR #668, opened 2021-11-23
[tests/e2e/ledger] add a test for many txs applied in the same block

tzemanovic:


tests/src/e2e/ledger_tests.rs
```diff
@@ -401,3 +402,74 @@ fn invalid_transactions() -> Result<()> {
     client.assert_success();
     Ok(())
 }
+
+/// Test that multiple txs submitted in the same block all get the tx result.
+///
+/// In this test we:
+/// 1. Run the ledger node with 10ms consensus timeout
```
batconjurer, line 409:
```suggestion
/// 1. Run the ledger node with 10s consensus timeout
```

tzemanovic: I did it again T.T

batconjurer (approved):
Other than a tiny nit, LGTM

tests/src/e2e/ledger_tests.rs
batconjurer: I guess you just love living life in the fast lane :rofl:

PR #685, opened 2021-12-01
[nix] improved nix integration

SimSaladin:
Make it so that `nix-shell` drops you in a shell with the same toolchains available as what is used in the `crate2nix`-based derivation (when doing `nix-build`).


tzemanovic:
pls restart ci

tzemanovic (approved):
LGTM, thanks!

PR #689, opened 2021-12-03
Modify MerkleTree

### LINKED ISSUE #671: Modify MerkleTree for Tendermint proof verification
I'd like to modify `MerkleTree` in the storage so that Anoma's proof can be verified as Tendermint proof.

I investigated the proof verification in Cosmos (ibc-go). It has a 2-layer tree. The commitment prefix (specifying the subtree) is required (ICS23). First, it verifies the proof of the subtree storing IBC-related data. Then, it verifies the proof of the base tree with the subtree's root.

Currently, Anoma's Merkle tree is 1 layer and we can get only 1 proof of the tree. So, by modifying the tree to 2-layer and returning both proofs, Cosmos can verify Anoma's proof.

yito88:
closes #671


shared/src/ledger/storage/merkle_tree.rs
```diff
@@ -0,0 +1,550 @@
...
+        let subtree = self
+            .subtrees
+            .get(&store_type)
+            .expect("The subtree should exist");
+        let value = subtree.get(&H::hash(sub_key.to_string()))?;
+        Ok(!value.is_zero())
+    }
+
+    /// Update the tree with the given key and value
+    pub fn update(&mut self, key: &Key, value: Vec<u8>) -> Result<()> {
```
tzemanovic, line 170: to avoid cloning
```suggestion
    pub fn update(&mut self, key: &Key, value: impl AsRef<[u8]>) -> Result<()> {
```

shared/src/ledger/storage/merkle_tree.rs
```diff
@@ -0,0 +1,550 @@
...
+            StoreType::PoS => write!(f, "pos"),
+            StoreType::Ibc => write!(f, "ibc"),
+        }
+    }
+}
+
+/// Merkle tree storage
+pub struct MerkleTree<H: StorageHasher + Default> {
+    base: SparseMerkleTree<H, H256, DefaultStore<H256>>,
+    subtrees: HashMap<StoreType, SparseMerkleTree<H, H256, DefaultStore<H256>>>,
```
tzemanovic, line 107: I'm not sure if we really need this HashMap as we'll always have static number of the subtrees. I'd be in favor of adding a field for each subtree type (like for the `base`) instead

shared/src/ledger/storage/merkle_tree.rs
```diff
@@ -0,0 +1,550 @@
...
+        Ok(())
+    }
+
+    /// Get the root
+    pub fn root(&self) -> MerkleRoot {
+        (*self.base.root()).into()
+    }
+
+    /// Get the stores of the base and sub trees
+    pub fn stores(&self) -> Result<MerkleTreeStores> {
```
tzemanovic, line 207: I think this function shouldn't then need to clone anything

tzemanovic (approved):
if you could pls rebase then I think we can merge

yito88:
pls update wasm

Issue #679, opened 2021-11-29
init-account: using a known wallet alias


tzemanovic:
When one runs `init-account` with `--alias` that is already known in the wallet, the confirmation dialog doesn't allow to choose `n` and just repeats the dialog:

```bash
anoma client init-account \
  --alias my-new-acc \
  --public-key my-key \
  --source my-key
> You're trying to create an alias "my-new-acc" that already exists for an address in your store.
> Would you like to replace it? [y/N]: n
> You're trying to create an alias "my-new-acc" that already exists for an address in your store.
> Would you like to replace it? [y/N]: n
```

When fixing this, we could also allow to change the alias to a different value on conflict.

PR #690, opened 2021-12-07
Alias infloop fix

murisi:
This is an attempt at resolving #679. With these changes:
* The dialog no longer repeats when decline option is selected
* The alias can be changed to a different value on conflict
* Overwrite is now `p`, reselect is now `s`, and declining is now `k`


tzemanovic (approved):
LGTM, thank you!

PR #683, opened 2021-11-30
remove accounts subspaces from storage memory

tzemanovic:

PR #691, opened 2021-12-07
ledger: add VP wasm compilation cache

### LINKED ISSUE #157: VP wasm compilation cache
We re-compile VPs everytime before we run them, even if the code doesn't change. Would it be possible to store pre-compiled VPs? The compilation gas cost could be charged as a one-off when the VP is being deployed. 

If this is possible, we would need to make sure re-compile validity predicates in case of any changes that could change the compilation output (i.e. protocol upgrades).

tzemanovic:
closes #157, depends on #683 (just for sysinfo to determine available memory) - commits included here for now

- [x] add config for the in-memory VP cache max bytes (I'll re-use sysinfo crate from #683 for rocksdb cache size)


tzemanovic:
pls restart ci

Fraccaman:
Request failed with status code 400

tzemanovic:
replaced by #697

PR #693, opened 2021-12-08
wasm debug logging

### LINKED ISSUE #678: remove WASM `log_string`s from release build
add conditional compilation to the debugging and log statements in wasm

tzemanovic:
closes #678 

I had to separate the tx and VP preludes into crates, because we cannot declare a macro with the same name `log_debug` in one crate (fails with error[E0428]: the name `debug_log` is defined multiple times). I've updated the diagram in docs: https://github.com/anoma/anoma/blob/tomas/wasm-debug-logging/docs/src/explore/design/overview/crates.svg

I ran into an issue with rebuilding `wasm_for_tests`, which increased the output files by a couple kb, but it was enough to start making the matchmaker (mm_token_exch) to start crashing in the `e2e::gossip_tests::match_intents`. I've added `wasm-opt` that we're using in non-test wasm to reduce the file sizes a bit again to get around this for now, but it looks that some dependencies even when unused end-up bloating the built wasm size, so we should investigate if we can remove some unnecessary dependencies from them (opened a follow-up issue #695)


tzemanovic:
pls restart ci

tzemanovic:
pls update wasm

Fraccaman:
pls update wasm

tzemanovic:
pls restart ci

tzemanovic:
pls update wasm

Fraccaman:
pls update wasm

juped:
pls update ci

PR #698, opened 2021-12-10
update wasmer 2.0.1

tzemanovic:
The new version fixes leak that affected the ledger on linux. Wasmer singlepass compiler doesn’t work on windows with version 2.0.0, it’s fixed in next version.


tzemanovic:
pls update wasm

PR #715, opened 2021-12-16
ledger/storage: enable DB atomic flushes, flush blocks without waiting

### LINKED ISSUE #372: consider giving a config option to flush storage on every block
We should consider giving a config option to flush storage on every block. Right now, it writes to memtable and only flushes on a clean shutdown.

Maybe the default should be to always flush as it's the safer option.

tzemanovic:
closes #372


tzemanovic:
pls update wasm

juped:
this falsely looks simple, but it actually confuses the diff algorithm a ton - can you rebase onto `v0.3.0`?

PR #728, opened 2021-12-21
tx queue in storage

### LINKED ISSUE #711: TxQueue should be persisted in local storage rather than separate file
The TxQueue is an essential part of a validator's local state. This queue is a global cache that can be overwritten every block. Currently it is persisted in  special file when the Shell is dropped from scope. On bootup, this file is needed to resume. However, in some cases, this file may not exist, e.g. on out-of-memory crashes. This locks validators out of restarting and forces them to replay from scratch.

Because it is crucial that the tx queue be in sync with the last persisted block, we should maintain a global tx queue in our storage and when persisting state, atomically commit both the block and the new queue.

The atomic commits also relates to #372

tzemanovic:
closes #711 

I've also modified the `make clippy(-abci-plus-plus)` recipes to explicitly set `ANOMA_DEV=false`, so that if you enable it in your shell, you can still run clippy on the same build that the CI uses.

This is a semi-breaking change from v0.3.0. Syncing an existing chain from scratch will work, but loading a state written by v0.3.0 will not work.

### DIFF
--- a/.changelog/unreleased/improvements/711-711.md
+++ b/.changelog/unreleased/improvements/711-711.md
@@ -1,0 +1,2 @@
+- Fix the `anoma client utils join-network` to respect `--base-dir` argument, if
+  specified ([#711](https://github.com/anoma/anoma/issues/711))
--- a/.drone.yml
+++ b/.drone.yml
@@ -17,7 +17,7 @@ steps:
     pull: never
   - commands:
       - >-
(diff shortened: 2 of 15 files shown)


tzemanovic:
pls update ci

batconjurer (approved):
Looks good to me. I recall talking with you about enabling atomic commits in rocksdb. Is this something we should do here or is it not a problem?

Issue #665, opened 2021-11-22
set CLI version to branch name if not release


tzemanovic:
Set the `anoma --version` to a branch name and the head commit hash when not building from a release tagged branch. The tagged release branch can continue to use `crate_version!()` clap macro

tzemanovic:
closed by #733

PR #733, opened 2021-12-21
apps: include `git describe` version string

juped:
Include a version string from `git describe` to use in --version
output, falling back on the crate version if no git repository is
found.

This doesn't actually work, because Cargo is bad and there's no way to force build.rs to rerun. Forcing it to rerun on .git changes might seem to work, but it doesn't - we may be in a worktree and .git may be a file pointing to the main repository. Better approaches welcome.


juped:
Addresses #665, albeit correctly (one cannot include branch names in version strings)

Issue #579, opened 2021-10-25
Create separate command for running the matchmaker


cwgoes:
Perhaps this is blocked on internal architectural changes, I'm not sure, but the matchmaker and gossip node should be at least in principle separable (split by an API) and they should be run with different commands, e.g. `anoma gossip` and `anoma matchmaker`, which does not seem to currently be the case

(feel free to provide additional context here if there are parts I misunderstand)

PR #746, opened 2021-12-27
split out matchmaker part 2 - moving matchmaker from wasm to dylib

tzemanovic:
In preparation for #579, we're moved the matchmaker WASM to a dylib that can currently be loaded by a intent gossiper node in place of a WASM file.

In this PR, we're adding a `Matchmaker` derive macro that can be used to automatically derive the binding code between the matchmaker runner and the implementation with a state that is managed by the runner. Thanks for this, we no longer need to (de-)serialize the matchmaker's state on each run.

- 71d8098535306b865e00f3cfeb35acd79349a11a: clean-up redundant async runtime on gossiper RPC
- 2db32bd3c778d2cea5d6d22b9dc88dfc01c28012: renamed "vm_macro" crate to just "macros" as it's no longer just for VM as we use it for matchmaker dylib now too
- 2142dc09309598b054b536e0e6dc0ef58bdf5e5a: re-purposed and updated the matchmaker macro to deriving macro, adding constructor, destructor and add_intent binding code
- f531579f58f4a20ab52c71c1d232853410022160: use the macro
- ad61bda85a5ef43de59a760200c41f6f11a3b813: update docs (see at https://github.com/anoma/anoma/blob/tomas/mm-state-macro/docs/src/user-guide/intent-gossip-and-matchmaker.md#-custom-matchmaker)


tzemanovic:
pls update wasm

Fraccaman (approved):

PR #756, opened 2021-12-29
update wasmer to 2.1.1

tzemanovic:


tzemanovic:
we only want to backport the last commit to maint

PR #770, opened 2022-01-04
[nix] use flakes

SimSaladin:
Flakes is a new feature of Nix which simplifies dependency management and allows more user-friendly Nix commands (via the unified `nix` executable). Here we add a `flake.nix` which allows us to use the new flake-enabled Nix commands. For example, this allows us to say `nix run github:anoma/anoma -- client` to launch the anoma client from scratch with nothing but Nix installed on the machine. Also, because flake evaluation is cached, the flake-based shell (`nix develop`) launches much faster (0.2 seconds) than `nix-shell` (2 seconds).

Aside: don't be alarmed by the huge diff: most of it is generated files under the `crate2nix` directory.

- The `Cargo.nix` and `crate-checksums.json` files (created with `crate2nix`) are now checked in, because flake evaluation is stricter in that untracked paths are not available to expressions. An alternative would be to do IFD with `crate2nix`, but I couldn't get this to work because of brittle Cargo dependency vendoring. This is hardly Nix-specific issue though, since `cargo vendor` also fails at the moment (due to the "conflicting" crates that are picked with feature flags).
- Replaced `shell.nix` and `default.nix` with `flake-compat` shims. This means that `nix-build` and `nix-shell` work like they used to. Compatibility with older Nix versions is preserved.
- The `generateCargoNix` flake app can be used to update the generated files: `nix run .#generateCargoNix`. (This is a mechanical task that can be offloaded to CI.)
- Added a `cargo` wrapper to the development shell which intercepts rustup-style cargo invocations (`cargo +VERSION ...`) and redirects to correct toolchain from the rust overlay. (This allows all `make` targets to work inside the development shell, without requiring `rustup`.)
- Added `binaryen` to the shell (needed for `make build-wasm-scripts`).
- Pinned tendermint to 0.34 in the `anoma` convenience derivation (which is the default package).
- Defined variations with the `ABCI-plus-plus` feature flag enabled (in opposition to `default`). These are identified by the suffix `:ABCI-plus-plus`.
- Added `scripts/ci/nix-check.sh` which checks the flake functionality (not yet integrated to CI).
- Updated README and User Guide w.r.t. using Nix.


tzemanovic:
This looks awesome!

Issue #634, opened 2021-11-08
add a command to find a result of a previously applied tx


tzemanovic:
This should be available via tendermint's events, which the client is already using to wait for tx's result (hopefully, this data is available in tendermint indefinitely, but we should check that)

tzemanovic:
for this, we should print the hash of submitted txs (wrapper and inner txs for ABCI++)

PR #771, opened 2022-01-05
Murisi/tx result extraction

murisi:
An attempt to resolve issue #634 . An implementation of a command line option `tx-result` to extract the results of a transaction from its hash. More precisely a transaction result comprises its `info`, `height`, `hash`, `code`, `gas_used`, and `initialized_accounts`. Note also that the hash can either be an application or an acceptance hash.

The following changes were made:
* Add CLI option `tx-result` to allow user to request result
* Made function that uses ABCI/++ query mechanisms to extract results
* A correction to `find_tx` to prevent it from crashing on ABCI++

Note that currently for this patch to work with ABCI++, heliaxdev/tendermint-rs#2 needs to be applied to `tendermint-rs`.

### DIFF
--- a/apps/src/lib/cli.rs
+++ b/apps/src/lib/cli.rs
@@ -142,7 +142,6 @@ pub mod cmds {
                 .subcommand(TxUpdateVp::def().display_order(1))
                 .subcommand(TxInitAccount::def().display_order(1))
                 .subcommand(TxInitValidator::def().display_order(1))
-                .subcommand(TxResult::def().display_order(1))
                 // PoS transactions
                 .subcommand(Bond::def().display_order(2))
                 .subcommand(Unbond::def().display_order(2))
@@ -153,6 +152,7 @@ pub mod cmds {
                 .subcommand(QueryBonds::def().display_order(3))
                 .subcommand(QueryVotingPower::def().display_order(3))
                 .subcommand(QuerySlashes::def().display_order(3))
+                .subcommand(QueryResult::def().display_order(3))
                 // Intents
                 .subcommand(Intent::def().display_order(4))
                 .subcommand(SubscribeTopic::def().display_order(4))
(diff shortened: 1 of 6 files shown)


apps/src/lib/cli.rs
```diff
@@ -673,6 +677,25 @@ pub mod cmds {
...
+
+        fn parse(matches: &ArgMatches) -> Option<Self> {
+            matches
+                .subcommand_matches(Self::CMD)
+                .map(|matches| TxResult(args::TxResult::parse(matches)))
+        }
+
+        fn def() -> App {
+            App::new(Self::CMD)
+                .about("Get the result associated with a transaction.")
```
tzemanovic, line 694:
```suggestion
                .about("Get the result of a transaction.")
```

apps/src/lib/cli.rs
```diff
@@ -142,6 +142,7 @@ pub mod cmds {
                 .subcommand(TxUpdateVp::def().display_order(1))
                 .subcommand(TxInitAccount::def().display_order(1))
                 .subcommand(TxInitValidator::def().display_order(1))
+                .subcommand(TxResult::def().display_order(1))
```
tzemanovic, line 145: I think this might be a bit better fix for the queries group 3 below

tzemanovic (approved):
This is looking great, thanks! I think we'll need to print the transactions results a bit nicer to make it clearer the inner tx's hash is the one with the result that can be queried, but that's for another issue. Once we merge https://github.com/heliaxdev/tendermint-rs/pull/2, we should update the lock file to the new ref here

apps/src/lib/cli.rs
murisi: Thanks! Good point. I'll rename `TxResult` to `QueryResult`, and place it in `display_order` 3.

PR #773, opened 2022-01-05
apps/ledger: handle sigterm on unix

### LINKED ISSUE #768: Ledger does a dirty shutdown on SIGTERM since 0.3.0
Since 0.3.0, ledger state always gets corrupted when the `anoman` process is killed with `SIGTERM` (which is the default signal e.g. systemd sends to stop a service). To start the node again after `SIGTERM` it's necessary to reset the state - a panic occurs at startup otherwise.

Using `SIGINT` instead (e.g. what CTRL-C does) does do a clean shutdown, so I've been using that as a workaround with systemd (`KillSignal=SIGINT` in the service definition).

`SIGTERM` used to do a clean shutdown in 0.2.0. It would be nice to get back that behavior, since it is what is expected per POSIX.

tzemanovic:
closes #768

### DIFF
--- a/.changelog/unreleased/bug-fixes/768-768.md
+++ b/.changelog/unreleased/bug-fixes/768-768.md
@@ -1,0 +1,2 @@
+- Ledger: Handle Unix and Windows interrupt and termination signals to shut down
+  cleanly. ([#768](https://github.com/anoma/anoma/issues/768))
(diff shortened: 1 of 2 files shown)


juped:
We should probably also trap SIGHUP and SIGPIPE for clean shutdown, but I think not SIGQUIT because ctrl-\ for a forced dirty shutdown is useful

tzemanovic:
I'll also add ctrl+break for windows

PR #774, opened 2022-01-05
More flexible ledger/intent gossip configuration

SimSaladin:
Operations improvements, because I really don't enjoy patching TOML with sed all the time.

- Enable setting config values via environment variables
- Add variables for configuring Tendermint instrumentation
- Allow missing values in the config file (filled in with defaults defined in the code)


tzemanovic (approved):
Very handy improvement, thanks for updating the docs too!

PR #775, opened 2022-01-06
enabling intent gossiper p2p connections

tzemanovic:
This PR enables remote peer discovery for the intent gossiper node.

We've added libp2p Identify (to automatically identify nodes periodically, return information about them, and answers identify queries from other nodes) and Ping protocol (to keep established connections live).

The updated `e2e::gossip_tests::run_gossip` now tests that two peers that don't know about each other can establish direct connection via a common seed node that gossips their info.

For convenience we can set one of the validator nodes in testnets to be a seed node (anoma-network-config change for devnet https://github.com/heliaxdev/anoma-network-config/commit/0945bcbc22669a0ac8352f4dfaf991b2cf97dc7d - note that the field `intent_gossip_bootstrap` has been later renamed to `intent_gossip_seed` in here 43c32da).


juped:
merged

Issue #403, opened 2021-08-19
add client option for cmds that submit a tx to not wait for it being applied


tzemanovic:
We currently wait for the result coming from a custom event via websocket (added in #377). Add an argument to the transaction arguments (`struct Tx` in `cli::args`) that is a flag that allows to not wait for the transaction to be applied in the block.

tzemanovic:
it might be better to have #634 first

PR #778, opened 2022-01-07
Murisi/broadcast only

murisi:
Implemented a feature to allow client to exit as soon as a transaction is in the mempool. See #403 . This change only makes sense if there is a way to later extract transaction results. See the pull request at #634 . The specific changes made are as follows:
* Added a flag `--broadcast-only` to allow user to indicate that client should exit once the transaction is in mempool
* Factored out the waiting/blocking code in `broadcast_tx` so that the client can effect early exits
* Added code to print transaction hashes even if the transactions have not been accepted/applied to the blockchain

### DIFF
--- a/apps/src/lib/cli.rs
+++ b/apps/src/lib/cli.rs
@@ -1112,6 +1112,7 @@ pub mod args {
             Err(_) => config::DEFAULT_BASE_DIR.into(),
         }),
     );
+    const BROADCAST_ONLY: ArgFlag = flag("broadcast-only");
     const CHAIN_ID: Arg<ChainId> = arg("chain-id");
     const CHAIN_ID_OPT: ArgOpt<ChainId> = CHAIN_ID.opt();
     const CHAIN_ID_PREFIX: Arg<ChainIdPrefix> = arg("chain-prefix");
@@ -1998,6 +1999,8 @@ pub mod args {
         pub dry_run: bool,
         /// Submit the transaction even if it doesn't pass client checks
         pub force: bool,
+        /// Do not wait for the transaction to be added to the blockchain
+        pub broadcast_only: bool,
         /// The address of the ledger node as host:port
         pub ledger_address: TendermintAddress,
         /// If any new account is initialized by the tx, use the given alias to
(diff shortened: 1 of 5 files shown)


apps/src/lib/client/tx.rs
```diff
@@ -758,134 +775,163 @@ async fn save_initialized_accounts(
...
-/// Broadcast a transaction to be included in the blockchain.
-///
-/// Checks that
-/// 1. The tx has been successfully included into the mempool of a validator
-/// 2. The tx with encrypted payload has been included on the blockchain
-/// 3. The decrypted payload of the tx has been included on the blockchain.
+/// Extract the wrapper transaction's hash and, if in ABCI++ mode, the inner
+/// transaction's hash. Useful for determining when parts of the given
+/// tranasaction make it on-chain.
+
```
tzemanovic, line 781:
```suggestion
```

apps/src/lib/client/tx.rs
```diff
@@ -686,9 +693,19 @@ async fn submit_tx(
             args.gas_limit.clone(),
             tx,
         );
-        match broadcast_tx(args.ledger_address.clone(), tx, keypair).await {
-            Ok(result) => (ctx, result.initialized_accounts),
-            Err(err) => {
+        // Either broadcast or submit transaction and collect result into
+        // sum type
+        let result = if args.broadcast_only {
+            Err(broadcast_tx(args.ledger_address.clone(), tx, keypair).await)
```
tzemanovic, line 699: I think the `Err` usage is a bit confusing here, you can probably use `itertools::Either` instead

tzemanovic (approved):
Just a couple small comments, besides that LGTM, thanks!

apps/src/lib/client/tx.rs
murisi: Thanks! Done - now using `itertools::Either` instead.

apps/src/lib/cli.rs
```diff
@@ -2025,6 +2028,10 @@ pub mod args {
             .arg(FORCE.def().about(
                 "Submit the transaction even if it doesn't pass client checks.",
             ))
+            .arg(BROADCAST_ONLY.def().about(
+                "Do not wait for the transaction to be added to the \
+                 blockchain.",
```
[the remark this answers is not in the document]
murisi: Done - now using the suggested documentation string instead. Thanks!

PR #788, opened 2022-01-11
docs: add quick start guide

tzemanovic:
rendered page: https://github.com/anoma/anoma/blob/tomas/quick-start/docs/src/quick-start.md

### DIFF
--- a/docs/src/quick-start.md
+++ b/docs/src/quick-start.md
@@ -67,7 +67,7 @@ $ tmux
 
 # inside the tmux/or not
 
-$ ./anoman ledger
+$ ./anoma ledger
 
 # can detach the tmux (Ctrl-B then D)
 ```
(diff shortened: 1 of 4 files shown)


gabriella-fw (commented):
I would say "applied in **a** future block" (line 150) and "check **a **future epoch" (line 246) but if everything else is factually correct, then I have no other comments :)

gabriella-fw (approved):

docs/src/quick-start.md
```diff
@@ -66,7 +66,7 @@ $ tmux
 
 # inside the tmux/or not
 
-$ ./anoman ledger
+$ ./anoma ledger
```
juped, line 69: i think this is just wrong?

juped (commented):
should base on tag v0.3.0, not somewhere random

tzemanovic:
I've rebased on v0.3.0

PR #789, opened 2022-01-11
Update quick-start.md

gabriella-fw:
change wget to curl


tzemanovic:
thanks @gabriella-fw! I've added this to #788 manually

PR #791, opened 2022-01-11
force non-dev release build and build check

tzemanovic:
(yup, I did `make package` with `ANOMA_DEV=true` in my env :)


tzemanovic:
pls update ci

PR #799, opened 2022-01-13
increment ABCI++ ports for e2e tests

### LINKED ISSUE #717: tests/e2e: use different ports for ABCI++ from ABCI
Right now, they use the same ports, so timing dependent, it sometimes fails with e.g.:
```
failed to listen on 127.0.0.1:26657: listen tcp 127.0.0.1:26657: bind: address already in use
```
(from https://ci.heliax.dev/anoma/anoma/2378/3/7)

In e2e/setup.rs, we already manipulate the source genesis file. For ABCI++, we can increase the port in the validator's `net_address` which will offset all the other ports used by Anoma too

tzemanovic:
closes #717

### DIFF
--- a/.changelog/unreleased/improvements/717-717.md
+++ b/.changelog/unreleased/improvements/717-717.md
@@ -1,0 +1,3 @@
+- Testing: Increments network configuration ports used for E2E
+  tests and ABCI++ enabled E2E tests to avoid sharing resources.
+  ([#717](https://github.com/anoma/anoma/issues/717))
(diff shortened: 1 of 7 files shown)


juped:
does still seem to have the same nondeterministic failures

tzemanovic:
> does still seem to have the same nondeterministic failures

yeah, we just found the problem and the last 2 commits should make it pass now but let's wait and see

juped:
will this fail if e.g. two separate CI runs are in test-e2e simultaneously?

Fraccaman:
I dont think so, everything runs inside his own docker container. Should be fine even if builds run in parallel

PR #808, opened 2022-01-17
Update WASM checksums

GGiecold:
Title says it all.


juped:
They aren't out of date - did you do `make build-wasm-scripts` locally? This will give you different hashes from the deterministic ones (`make build-wasm-scripts-docker`).

PR #811, opened 2022-01-18
Clean up changelog files

juped:
non-functional changes


tzemanovic:
good idea! let's also update the contributing.md

PR #819, opened 2022-01-24
make: install the default mm_token_exch lib

### LINKED ISSUE #816: `make install` doesn't install the mm_token_exch matchmaker dylib

tzemanovic:
closes #816


Makefile
```diff
@@ -89,7 +89,10 @@ clippy-fix:
 	$(cargo) +$(nightly) clippy --fix -Z unstable-options --all-targets --allow-dirty --allow-staged
 
 install: tendermint
-	ANOMA_DEV=false $(cargo) install --path ./apps
+	$(cargo) build --release \
+		--manifest-path matchmaker/mm_token_exch/Cargo.toml && \
+	cp target/release/libmm_token_exch.{dll,dylib,so} ${HOME}/.cargo/bin
```
juped, line 94: This is a shared object, it should belong in some kind of lib directory

Makefile
```diff
@@ -89,7 +89,10 @@ clippy-fix:
...
+	# ANOMA_DEV=false $(cargo) install --path ./apps
```
juped, line 95: Do you want install to install only the shared object for the matchmaker now?

Makefile
juped: install(1) is more idiomatic in a makefile install target than cp(1) for various reasons

Makefile
tzemanovic: ah, no sorry

Makefile
tzemanovic: yeah, I was thinking about putting it into `~/.cargo/lib` instead, but cargo doesn't use any such dir atm

tzemanovic: I'll switch to `install`, thanks!

juped: I imagine there would be a ~/.cargo/lib if the rust-lang/cargo#8294 you mentioned were supported; we could consider just unilaterally making it (`install -d`). It won't be in anyone's LD_LIBRARY_PATH, though we locate it with a command line argument anyway so this doesn't matter.

tzemanovic: I switched it to the `~/.cargo/lib` dir and added additional attempt in the matchmaker runner to try to load from that dir, the current directory or directly from the given path when it's absolute. I've also made the file extension optional to make it more flexible (previously, it had to be specified without extension)

PR #862, opened 2022-02-02
move network config WASM dirs inside chain dir

### LINKED ISSUE #838: place wasm-dir inside chain id dir
I propose that we move the wasm-dir under base-dir/chain-dir, because we'll want to be able to run multiple chains, which can have different wasm checksums. This might be a little bit of a breaking change for our tooling (e.g. deployment scripts will actually be simpler) and anoma-network-config, so it's probably best to do this early

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201757896515792/1201762706751668) by [Unito](https://www.unito.io)

### LINKED ISSUE #843: join-network with non-default base-dir
When the `--base-dir` or `ANOMA_BASE_DIR` is specified and is not the default, the command will fail if the given path is not a child of the current working directory.

e.g.

```shell
cd /User/x/foo
ANOMA_BASE_DIR=/User/x/bar anomac utils join-network --chain-id=anoma-devnet-2.6.b1a39df490b6b
```

fails with:

    The application panicked (crashed).
    Message:  called `Result::unwrap()` on an `Err` value: Custom { kind: Other, error: "paths in archives must not have `..` when setting path for " }
    Location: anoma/apps/src/lib/client/utils.rs:116This is because it's trying to recreate the network config archive with paths relative to the current working directory (with the example above that would be `../bar`), but as the error says "paths in archives must not have `..` when setting path".

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201757896515792/1201762706313717) by [Unito](https://www.unito.io)

tzemanovic:
closes #838, fixes #843

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201757896515792/1201762701328741) by [Unito](https://www.unito.io)

### DIFF
--- a/.changelog/unreleased/improvements/838-wasm-dir-inside-chain-dir.md
+++ b/.changelog/unreleased/improvements/838-wasm-dir-inside-chain-dir.md
@@ -1,0 +1,2 @@
+- Network config: The WASM dir were moved inside chain directories.
+  ([#838](https://github.com/anoma/anoma/issues/838))
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -247,6 +247,7 @@ dependencies = [
  "concat-idents",
  "escargot",
  "eyre",
+ "fs_extra",
  "ibc 0.8.0 (git+https://github.com/heliaxdev/ibc-rs?branch=yuji/abcipp-v0.23.0)",
  "ibc 0.8.0 (git+https://github.com/heliaxdev/ibc-rs?branch=yuji/rebase_v0.23.0)",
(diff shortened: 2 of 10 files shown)


tzemanovic:
@juped I think this is ok to backport to v0.4, as long as it's not used to generate new network configs, because those wouldn't be backwards compatible. It is however compatible with the v0.4 network configs on `join-network`, for which it will move the wasm dir to the new location

tzemanovic:
bad 🤖

juped:
If we make a `v0.4.1`, I expect we'll want to use this layout.

juped (approved):

PR #863, opened 2022-02-03
Update tendermint-rs and ibc-rs

yito88:
* Rust toolchain 1.56.1 to 1.58.1
* tendermint-rs v0.23.0 to v0.23.5
* ibc-rs v0.8.0 to v0.11.0

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201757896515792/1201762852675246) by [Unito](https://www.unito.io)


tzemanovic:
naughty 🤖

tzemanovic:
Nice, I'm looking forward to using https://blog.rust-lang.org/2022/01/13/Rust-1.58.0.html#captured-identifiers-in-format-strings and https://blog.rust-lang.org/2021/12/02/Rust-1.57.0.html#cargo-support-for-custom-profiles!

shared/src/ledger/ibc/vp/channel.rs
```diff
@@ -851,21 +915,67 @@ where
...
+
     fn channel_counter(&self) -> Ics04Result<u64> {
         let key = channel_counter_key();
         self.read_counter(&key)
             .map_err(|_| Ics04Error::implementation_specific())
     }
+
+    fn max_expected_time_per_block(&self) -> Duration {
+        // TODO set the proper duration
+        Duration::new(5, 0)
```
yito88, line 977: I'm not sure what this duration should be.
This duration will be used to calculate `delay_period_height`. It checks if a client state (light client) is not too old.

juped: I think this is in the configuration?

yito88: Thank you! I'll modify it to get the duration from the configuration.
`ibc-go` gets it from the configuration.

shared/src/ledger/ibc/handler.rs
```diff
@@ -858,6 +867,25 @@ pub trait IbcActions {
         Ok(())
     }
 
+    /// Set the timestamp and the height for the client update
+    fn set_client_update_time(&self, client_id: &ClientId) -> Result<()> {
+        let timestamp = Timestamp::now();
```
tzemanovic, line 872: I think will probably be different timestamp on each node. Won't it break consensus in the storage?

shared/src/ledger/ibc/storage.rs
```diff
@@ -155,54 +167,54 @@ pub fn capability_index_key() -> Key {
 #[cfg(any(feature = "ibc-vp", feature = "ibc-vp-abci"))]
```
tzemanovic, line 167: (Unrelated to this PR) I didn't realize, but I think we can probably remove all the `#[cfg(any(feature = "ibc-vp", feature = "ibc-vp-abci"))]` annotations, as they will be always enabled anyway

shared/src/ledger/ibc/vp/channel.rs
```diff
@@ -851,21 +916,74 @@ where
...
+        }
+    }
+
     fn channel_counter(&self) -> Ics04Result<u64> {
         let key = channel_counter_key();
         self.read_counter(&key)
             .map_err(|_| Ics04Error::implementation_specific())
     }
+
+    fn max_expected_time_per_block(&self) -> Duration {
```
tzemanovic, line 976: I think when querying this value we should return the `Err` instead of the default value (in case the tx runs out of gas):
```suggestion
    fn max_expected_time_per_block(&self) -> Result<Duration> {
```

shared/src/ledger/ibc/handler.rs
yito88: The timestamp will be checked if a client state is valid on this node.

shared/src/ledger/ibc/vp/channel.rs
yito88: Thank you for your suggestion! Yes, it should return Result. I'll propose and modify that in ibc-rs.

shared/src/ledger/ibc/storage.rs
yito88: I tried that before and wasm build failed. Some functions depend on `ibc-rs`.
After `ibc-rs` supports `no-std`, we can remove all.
Or, we can avoid annotating each function by splitting the `storage` file.

Fraccaman:
I think that in `anoma/wasm/rust-toolchain.toml` the toolchain should be updated to 1.58.1 right? @yito88

wasm/rust-toolchain.toml
```diff
@@ -1,3 +1,3 @@
 [toolchain]
-channel = "1.56.1"
+channel = "1.58"
```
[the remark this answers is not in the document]
Fraccaman: yep needs to be 1.58.1

shared/src/ledger/ibc/handler.rs
yito88: Thank you for the suggestion! In ibc-rs, the time is acquired from Tendermint header.
I've fixed it.

vm_env/src/imports.rs
```diff
@@ -216,6 +217,19 @@ pub mod tx {
         BlockHeight(unsafe { anoma_tx_get_block_height() })
     }
 
+    /// Get time of the current block header as rfc3339
+    pub fn get_block_time() -> String {
```
tzemanovic, line 221: I think maybe we could wrap the string with `Rfc3339String` here instead of in `vm_env/src/ibc.rs`

shared/src/types/time.rs
```diff
@@ -14,6 +14,9 @@ use tendermint_stable::time::Time;
 #[cfg(feature = "ABCI")]
 use tendermint_stable::Error as TendermintError;
 
+/// The length of the block's time as RFC 3339 string
+pub const BLOCK_TIME_LENGTH: usize = 27;
```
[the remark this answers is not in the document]
tzemanovic: I think it's probably why the unit tests are failing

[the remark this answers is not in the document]
yito88: Thanks! I've fixed it

yito88:
pls update wasm

tzemanovic (approved):
LGTM! just needs a changelog I think

GGiecold (approved):

juped:
When building wasms, I get:

```
warning: Patch `tendermint-light-client v0.23.5 (https://github.com/heliaxdev/tendermint-rs?branch=yuji/rebase_v0.23.5#a1439b37)` was not used in the crate graph.
Check that the patched package version and available features are compatible
with the dependency requirements. If the patch has a different version from
what is locked in the Cargo.lock file, run `cargo update` to use the new
version. This may also occur with an optional dependency that is not enabled.
```

tzemanovic:
> When building wasms, I get:
> 
> ```
> warning: Patch `tendermint-light-client v0.23.5 (https://github.com/heliaxdev/tendermint-rs?branch=yuji/rebase_v0.23.5#a1439b37)` was not used in the crate graph.
> Check that the patched package version and available features are compatible
> with the dependency requirements. If the patch has a different version from
> what is locked in the Cargo.lock file, run `cargo update` to use the new
> version. This may also occur with an optional dependency that is not enabled.
> ```

It looks like the `tendermint-light-client` patch in `wasm/wasm_source/Cargo.toml` is unused. Instead the lock file is changed to use `tendermint-light-client-verifier`, so I think we need to patch that instead

tzemanovic:
> > When building wasms, I get:
> > ```
> > warning: Patch `tendermint-light-client v0.23.5 (https://github.com/heliaxdev/tendermint-rs?branch=yuji/rebase_v0.23.5#a1439b37)` was not used in the crate graph.
> > Check that the patched package version and available features are compatible
> > with the dependency requirements. If the patch has a different version from
> > what is locked in the Cargo.lock file, run `cargo update` to use the new
> > version. This may also occur with an optional dependency that is not enabled.
> > ```
> 
> It looks like the `tendermint-light-client` patch in `wasm/wasm_source/Cargo.toml` is unused. Instead the lock file is changed to use `tendermint-light-client-verifier`, so I think we need to patch that instead

I pushed a fix for this

tzemanovic:
one more [98719eb](https://github.com/anoma/anoma/pull/863/commits/98719eb64bf5a60a70e0193b749d8a8c7223b352)

tzemanovic:
I'll rebuild the wasm and merge

tzemanovic:
pls update wasm

juped:
shouldn't have been rebased forward, was already in the correct place

Issue #686, opened 2021-12-02
tests/e2e: switch from rexpect to expectrl


tzemanovic:
Let's switch e2e tests from rexpect to expectrl (https://github.com/zhiburt/expectrl). 

We had to patch rexpect for https://github.com/philippkeller/rexpect/issues/38 to build with Rust 1.55+ in #618. Expectrl reportedly doesn't have this issue and as a bonus supports windows.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201757896515792/1201763131740023) by [Unito](https://www.unito.io)

tzemanovic:
we made a fork of expectrl when we started working on this in #879 we'll want to use for this, because the lazy expect was way too heavy for our node logs https://github.com/heliaxdev/expectrl/tree/tomas/expect-eager

PR #879, opened 2022-02-04
DKG

### LINKED ISSUE #357: Integrate DKG from Ferveo into Anoma
We need to implement the DKG protocol into Anoma using Ferveo. This needs to be properly spec'ed and implemented and added to Anoma.

* [x] Add the dkg state machine to Anoma shell
* [x] Be able to persist the state machine on shutdown / crash and reboot from persistence
* [x] Create a new dkg instance from genesis
* [x] Create the necessary protocol tx types to handle dkg messages
* [x] #670
* [x] Update ProcessProposal to verify DKG messages with the state machine
* [x] Update FinalizeProposal to apply verified DKG messages
* [x] Update FinalizeProposal to start a new dkg instance and send an announcement at the start of a new epoch
* [x] Update FinalizeProposal to send a pvss transcript message if everyone has announced their session keys
* [x] Update PrepareProposal to add a single pvss transcript if any are in the mempool
* [x] Update PrepareProposal to add an aggregate transcript message if the dkg is in success state
* [x] Make changes to Ferveo as necessary
* [x] Add generated key to storage 
* [x] Make this epochs encryption key queryable
* [x] #672
* [x] Add tests

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201757896515792/1201762888081051) by [Unito](https://www.unito.io)

batconjurer:
Closes #357 

This is an MVP implementation of the distributed key generation in Anoma. It currently sits behind the ABCI++ feature gate. This feature adds the following 
 * A DKG state machine to the ledger
 * Generation of an encryption key once per epoch
 * The client queries the current encryption key before encrypting a tx (although it does not use the query result but a hardcoded key)
 * Protocol transactions and the ability for the ledger to issue txs as part of running the DKG protocol
 
 There are also numerous other fixes in here, e.g. making the keys read from the wallet thread-safe and replacing rexpect in the e2e tests with expectrl (closing #686 )
 
 Note that this PR disables some clippy lints due to failing lints in `blake2:0.10.0` and `x25519-dalek`.
 
 TODO: 
 * The `run_gossiper` test is failing for reasons I don't understand.


tzemanovic:
:(

Issue #455, opened 2021-09-10
refactor `TxResult` event


tzemanovic:
We only need to have the `"applied.hash"` field separately for the tendermint query when we're waiting for tx result in the client. For everything else, we should serialize the whole `TxResult` into a single field, so we can use the same type in the client to decode it. 

Currently we're doing this manually in `impl From<serde_json::Value> for TxResponse {` in `apps/src/lib/client/tx.rs` and in `fn finalize_block` in `apps/src/lib/node/ledger/shell.rs`

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201757896515792/1201762912053900) by [Unito](https://www.unito.io)

PR #910, opened 2022-02-22
Undecryptable Tx UX improvements

batconjurer:
Updated the tx result from an undecryptable tx to give an error code and message stating message was not decryptable


batconjurer:
pls update wasm

apps/src/lib/node/ledger/shell/finalize_block.rs
```diff
@@ -205,7 +213,7 @@ where
                         tx_result["code"] = ErrorCodes::InvalidTx.into();
                     }
                     tx_result["gas_used"] = result.gas_used.to_string();
-                    tx_result["info"] = result.to_string();
+                    tx_result["info"].push_str(&result.to_string());
```
tzemanovic, line 216: I'm not sure if we should be appending strings like this as it's gonna make the tx results harder to parse. Ideally it should all go via `TxResult`

apps/src/lib/node/ledger/shell/finalize_block.rs
```diff
@@ -146,7 +146,13 @@ where
                     if !cfg!(feature = "ABCI") {
                         self.storage.tx_queue.pop();
                     }
-                    Event::new_tx_event(&tx_type, height.0)
+                    let mut event = Event::new_tx_event(&tx_type, height.0);
+                    if let DecryptedTx::Undecryptable(_) = inner {
+                        event["info"] =
+                            "Transaction could not be decrypted. ".into();
+                        event["code"] = ErrorCodes::Undecryptable.into();
```
tzemanovic, line 153: I'm a little confused about this. When the `inner` is `Undecryptable`, shouldn't we also `continue` the loop? It looks like it still goes on to apply it below

apps/src/lib/node/ledger/shell/finalize_block.rs
tzemanovic: I guess that would include all the tx error conditions (we have this issue to refactor it https://github.com/anoma/anoma/issues/455)

apps/src/lib/node/ledger/shell/finalize_block.rs
tzemanovic: oh, it's because it's not yet being used, right?

apps/src/lib/node/ledger/shell/finalize_block.rs
batconjurer: So the error code is enough to know that it's not decryptable. So I can remove the extra info line. I'm not entirely happy though because all other error codes mean that the proposed block should be voted against by validators. This is an accepting code path (so not an error from the protocol pov), but it is an error from the user pov

apps/src/lib/node/ledger/shell/finalize_block.rs
tzemanovic: I see, thank you!

apps/src/lib/node/ledger/shell/finalize_block.rs
tzemanovic: Yeah, it would be good to preserve the information. Could we maybe add it as an optional field to the `TxResult` so that it can still be parsed from the "info" field?

batconjurer: I've added a log field

batconjurer:
pls update wasm

tzemanovic (approved):

PR #913, opened 2022-02-23
Support protocol transactions

batconjurer:
Supports a new type of transaction intended to be sent by validators, so called _protocol transactions_.

* New transaction type
* Generation of keys for validators to sign protocol txs
* A service to broadcast protocol txs from the ledger
* Improved client tx broadcasting


batconjurer:
pls update wasm

batconjurer:
pls update wasm

apps/src/lib/config/genesis.rs
```diff
@@ -523,23 +548,15 @@ pub struct Genesis {
 impl Genesis {
     /// Sort all fields for deterministic encoding
     pub fn init(&mut self) {
-        self.validators.sort();
+        self.validators
+            .sort_by(|val, other| val.account_key.cmp(&other.account_key));
```
tzemanovic, line 552: I'm not sure if this is safe to do as there's no requirement for the account keys to be unique and this order must be deterministic as the genesis hash relies on it. Instead, I think we should add `PartialOrd, Ord, PartialEq, Eq` on `DkgPublicKey` using the encoded data, since we have a wrapper already

batconjurer: I'm not sure I understand your recommendation

batconjurer: Actually, I can probably revert it back to the old way. The necessary ordering for the DKG is done by ferveo as well

apps/src/lib/client/tx.rs
```diff
@@ -59,15 +58,29 @@ const TX_UNBOND_WASM: &str = "tx_unbond.wasm";
...
     let tx_code = ctx.read_wasm(args.code_path);
     let data = args.data_path.map(|data_path| {
         std::fs::read(data_path).expect("Expected a file at given data path")
     });
     let tx = Tx::new(tx_code, data);
-    let (ctx, tx, keypair) = sign_tx(ctx, tx, &args.tx, None).await;
-    let (ctx, initialized_accounts) =
-        process_tx(ctx, &args.tx, tx, &keypair).await;
+    let (ctx, tx) = sign_tx(ctx, tx, &args.tx, None).await;
+    let (ctx, initialized_accounts) = process_tx(ctx, &args.tx, tx).await;
```
tzemanovic, line 83: not having the pass the keypair around is a nice improvement! I wonder if we could make it even nicer on the caller side, because we're always calling both of these together in here with e.g.:
```rust
let (ctx, initialized_accounts) = sign_and_process_tx(ctx, tx, &args.tx, None).await;
```

tzemanovic: (not necessarily in this PR)

batconjurer: That's true. We could definitely merge those into one call

apps/src/lib/wallet/mod.rs
```diff
@@ -115,6 +115,56 @@ impl Wallet {
         (alias.into(), key)
     }
 
+    /// Store the validator address in the wallet. Generate keypair
```
tzemanovic, line 118: should this be just
```suggestion
    /// Generate keypair
```
?

apps/src/lib/node/ledger/shell/mod.rs
```diff
@@ -542,21 +681,28 @@ mod test_utils {
 
     impl TestShell {
         /// Create a new shell
```
tzemanovic, line 683:
```suggestion
        /// Returns a new shell paired with a broadcast receiver, which will
        /// receives any protocol txs sent by the shell.
```

apps/src/bin/anoma-node/cli.rs
```diff
@@ -6,6 +6,7 @@ use eyre::{Context, Result};
 
 pub fn main() -> Result<()> {
     let (cmd, mut ctx) = cli::anoma_node_cli();
+    ctx.config.ledger.tendermint.tendermint_mode = ctx.global_args.mode.clone();
```
tzemanovic, line 9: I think maybe this would be better in somewhere in the `cli::Context::new`. The `global_args.mode` defaults to Validator when it's not specified though, so it looks like this will always override it to be Validator?

batconjurer: I thought that some of the e2e tests specificied full node. Otherwise they would fail.

tzemanovic: Ah right, what I was thinking about is if you set it in your config, but you don't set the global arg, it will get overridden on start-up to the default

tzemanovic: it might be better to make the global arg optional and only override it if it's set

batconjurer:
pls update wasm

tzemanovic (approved):
there's something failing build from the last change, but besides that this LGTM, thanks!

batconjurer:
pls update wasm

PR #905, opened 2022-02-18
common signing interface + ed25519 dependency change

murisi:
Changes:
* Replaced ed25519-dalek with ed25519-consensus to resolve #753 .
* ~Added secp256k1 support to resolve #162~ moved to #914
* Added common signing interface to resolve #225 .
* Removed old rand dependency.

depends on #889

PR #914, opened 2022-02-23
add secp256k1

tzemanovic:
depends on anoma/anoma#905

closes anoma/namada#4 

- [ ] need to add cli arg to choose scheme for commands that generate keys


juped:
updated, check if it's correct

tzemanovic:
replaced by https://github.com/anoma/anoma/pull/1163

PR #882, opened 2022-02-07
NFT Implementation

Fraccaman:

PR #916, opened 2022-02-24
re-add tx ibc wasm

tzemanovic:
fix-up for accidental removal from #882 (based on the `fraccaman/nft-impl`, which was before the toolchain update so CI won't pass)


juped:
not quite complete, it's also removed other places

tzemanovic:
> not quite complete, it's also removed other places

oh yeah, thanks!

tzemanovic:
ah, looks like I also don't have the up-to-date base

PR #915, opened 2022-02-23
Add a VP for implicit account addresses

GGiecold:
Addresses https://github.com/anoma/anoma/issues/193

PR #917, opened 2022-02-24
protocol VP storage and default implicit VP

juped:
reopened #915, based on fd4b7ab36929f47369ae82c82966891cb0ccc625 and with edited history (may require further splitting)


shared/src/ledger/mod.rs
```diff
@@ -5,5 +5,6 @@ pub mod ibc;
 pub mod native_vp;
 pub mod parameters;
 pub mod pos;
+pub mod protocol_vps;
```
juped, line 8: can we standardize on "protocol vp" or "protocol vps" or something else, it's "mod protocol_vps" and "struct ProtocolVp" right now

shared/src/types/address.rs
```diff
@@ -47,6 +47,8 @@ mod internal {
         "ano::IBC Burn Address                        ";
     pub const IBC_MINT: &str =
         "ano::IBC Mint Address                        ";
+    pub const PROTOCOL: &str =
+        "ano::Implicit VP                             ";
```
juped, line 51: needs more generic name like "protocol VP storage", since this is a permanent constant

wasm/wasm_source/src/vp_implicit.rs
```diff
@@ -0,0 +1,601 @@
...
+//! sending tokens (receiving tokens is permissive).
+//!
+//! It allows to bond, unbond and withdraw tokens to and from PoS system with a
+//! valid signature.
+//!
+//! It allows to fulfil intents that were signed by this account's key if they
+//! haven't already been fulfilled (fulfilled intents are added to the owner's
+//! invalid intent set).
+//!
+//! Any other storage key changes are allowed only with a valid signature.
```
juped, line 13: I think we don't want to give implicit accounts arbitrary storage as of my last recollection, but this is a design detail

juped (commented):
additionally, the genesis config (`apps/src/lib/config/genesis.rs`) needs to be changed to load an implicit VP into storage

GGiecold:
pls update wasm

juped (commented):
please avoid merging upstreams into topic branches

PR #920, opened 2022-02-24
wallet: canonicalize all aliases to lowercase

juped:
#884


GGiecold (commented):
Thanks for carrying out that code change 🙏🏻

juped:
i messed it up, but have fixed it now

PR #921, opened 2022-02-24
wallet: remove redundant clone()

juped:
The initial version of 'wallet: canonicalize all aliases to lowercase'
(ac3f0ba8e) was made against a less stringent clippy, and I failed
to check for new warnings when applying it. Rectify the redundant
clone warning.

### DIFF
--- a/apps/src/bin/anoma-wallet/cli.rs
+++ b/apps/src/bin/anoma-wallet/cli.rs
@@ -153,7 +153,7 @@ fn key_list(
 fn key_export(ctx: Context, args::KeyExport { alias }: args::KeyExport) {
     let mut wallet = ctx.wallet;
     wallet
-        .find_key(alias.clone().to_lowercase())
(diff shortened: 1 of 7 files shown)


tzemanovic:
there are 3 e2e tests failing, I'll fix 'em up

juped:
just need to downcase the command line arguments which can take aliases, i think (--signing-key is the only one in the tests i think, but there are others)

tzemanovic:
> just need to downcase the command line arguments which can take aliases, i think (--signing-key is the only one in the tests i think, but there are others)

yeah, the lower-casing is in somewhat ad-hoc locations so it's kinda error prone. I'm trying to refactor it a bit to encapsulate the lower-casing behavior

tzemanovic:
@juped Everything should be passing now

tzemanovic:
I also renamed the default aliases in 7a50f56 to snake case to keep them readable in lowercase

PR #943, opened 2022-03-07
Fixed the setting of Tendermint mode in cli

batconjurer:
Fixed a bug which unwrapped a None when setting tendermint mode from global args (introduced in #913)


tzemanovic (approved):
LGTM!

PR #946, opened 2022-03-09
Refactors `rpc` module

### LINKED ISSUE #840: RPC queries - refactor library code
Move the RPC library code from `apps/src/lib/client/rpc.rs` to the `shared` Anoma crate. Factor out the code's side-effects to allow to query storage data with any client that `impl tendermint_rpc::client::Client` and return the typed values (e.g. `Result<pos::Bonds, QueryError>`. The prints back to stdout/stderr should be factored out and kept in the `apps` crate.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201757896515792/1201762840944343) by [Unito](https://www.unito.io)

grarco:
Moves `rpc` modules from `apps` to `shared`. Refactors return types and functions. Fixes anoma/namada#66


grarco:
pls update wasm

grarco:
pls update wasm

PR #918, opened 2022-02-24
Deterministic order verifiers/changed_keys

Fraccaman:
Switching from `Hashset` to `BTreeSet` for `verifiers` and `changed_keys`. Fix #891.

PR #947, opened 2022-03-09
Pin test host env for stable memory location

tzemanovic:
~depends and based on #918~ rebased on v0.5.0


tzemanovic:
pls update wasm

tzemanovic:
pls update wasm

PR #954, opened 2022-03-10
governance: init proposal

Fraccaman:
- Added CLI commands:
    - Submit on chain proposal
    - Create offline proposal
    - Query existing proposals
- Added initial governance VP 
- Added governance parameters to genesis


grarco (approved):

Makefile
```diff
@@ -235,4 +235,8 @@ test-miri:
 	$(cargo) +$(nightly) clean
 	MIRIFLAGS="-Zmiri-disable-isolation" $(cargo) +$(nightly) miri test
 
-.PHONY : build check build-release clippy install run-ledger run-gossip reset-ledger test test-debug fmt watch clean build-doc doc build-wasm-scripts-docker build-wasm-scripts clean-wasm-scripts dev-deps test-miri
+copy-wasm:
+	rm -rf .anoma/anoma-devchain.000000000000000/wasm/*.wasm .anoma/anoma-devchain.000000000000000/wasm/*.json
```
tzemanovic, line 239: I think we can remove this, it's been fixed in https://github.com/anoma/anoma/pull/936, which uses the root wasm directory directly in dev-mode.

Fraccaman: ah nice! ill remove it :)

apps/src/lib/cli.rs
```diff
@@ -1278,6 +1370,10 @@ pub mod args {
     const OWNER: ArgOpt<WalletAddress> = arg_opt("owner");
     const PROTOCOL_KEY: ArgOpt<WalletPublicKey> = arg_opt("protocol-key");
     const PUBLIC_KEY: Arg<WalletPublicKey> = arg("public-key");
+    const PROPOSAL_OFFLINE: ArgFlag = flag("offline");
```
tzemanovic, line 1373: let's keep these sorted pls

apps/src/lib/cli.rs
```diff
@@ -1278,6 +1370,10 @@ pub mod args {
...
+    const PROPOSAL_ID: Arg<u64> = arg("proposal-id");
+    const OPTIONAL_PROPOSAL_ID: ArgOpt<u64> = arg_opt("proposal-id");
```
tzemanovic, line 1375: (just to keep them together after sorting)
```suggestion
    const PROPOSAL_ID_OPT: ArgOpt<u64> = arg_opt("proposal-id");
```

apps/src/lib/cli/context.rs
```diff
@@ -56,7 +56,7 @@ impl Context {
         let mut config = Config::load(
             &global_args.base_dir,
             &global_config.default_chain_id,
-            global_args.mode.clone().unwrap(),
+            TendermintMode::Validator,
```
tzemanovic, line 59: this shouldn't change here

tzemanovic: (we got related PR https://github.com/anoma/anoma/pull/983)

apps/src/lib/client/rpc.rs
```diff
@@ -165,6 +166,78 @@ pub async fn query_balance(ctx: Context, args: args::QueryBalance) {
     }
 }
 
+/// Query Proposals
+pub async fn query_proposal(_ctx: Context, args: args::QueryProposal) {
+    async fn print_proposal(client: &HttpClient, id: u64, details: bool) {
```
tzemanovic, line 171: It would be better to avoid further queries when one of the required fields is missing (especially because we're using the loop from 0 when proposal ID is not specified). One way would be to:
```suggestion
    async fn print_proposal(client: &HttpClient, id: u64, details: bool) -> Option<()> {
```
then you can use `?` on the queries, e.g.: `let author_key = gov_storage::get_author_key(id)?;` and return `Some(())` at the end to make it typecheck.
The caller should then handle `None` result to print the error message `eprintln!("No valid proposal was found with id {}", id)`. I'm not sure if we'll be deleting any data from tallied and applied proposals, but if so we might want to skip printing the error message in the loop

apps/src/lib/client/tx.rs
```diff
@@ -861,6 +929,27 @@ async fn process_tx(
...
+            }
+            true
+        }
+        None => {
+            eprintln!("Can't find address {}.", address);
+            safe_exit(1);
+        }
+    }
+}
+
```
tzemanovic, line 952:
```suggestion
async fn get_token_balance(
    client: &HttpClient,
    token: &Address,
    address: &Address,
) -> Option<token::Amount> {
    let balance_key = token::balance_key(token, address);
    rpc::query_storage_value::<token::Amount>(client, &balance_key)
}
```

and then at the caller side, e.g.:
```
        // default == 0
        let balance = get_token_balance(&client, &m1t(), &proposal.author).await.unwrap_or_default();
        if balance < min_amount {
            eprintln!("Address {} doesn't have enough funds.", address);
            safe_exit(1);
        }
        ...
```

tzemanovic: `get_token_balance` should probably go in the queries module, but we can re-use it in couple more places

apps/src/lib/config/genesis.rs
```diff
@@ -109,10 +111,31 @@ pub mod genesis_config {
...
         // Wasm definitions
         pub wasm: HashMap<String, WasmConfig>,
     }
 
+    #[derive(Clone, Debug, Deserialize, Serialize)]
+    pub struct GovernanceParamasConfig {
+        // Min funds to stake to submit a proposal
+        // XXX: u64 doesn't work with toml-rs!
+        pub min_proposal_fund: u64,
+        // Maximum size of proposal in kilobytes
```
tzemanovic, line 125:
```suggestion
        // Maximum size of proposal in kibibytes (KiB)
```

shared/src/ledger/governance/mod.rs
```diff
@@ -0,0 +1,493 @@
...
+                        (
+                            Some(min_funds_parameter),
+                            None,
+                            Some(post_balance),
+                        ) => post_balance >= min_funds_parameter,
+                        _ => false,
+                    }
+                }
+                (KeyType::PARAMETER, _) => false,
+                (KeyType::UNKNOWN, _) => false,
```
tzemanovic, line 337: I think in order to avoid having to change this again for https://github.com/anoma/anoma/issues/955 (and in general in other VPs), we could have two cases here:
- an unknown key type that is within this governance address space (the first segment == this address), that is rejected
- and another case for unknown keys from other accounts, which this VP shouldn't care about and so it should be permissive returning `true`

shared/src/ledger/governance/mod.rs
```diff
@@ -0,0 +1,493 @@
...
+        counter_key,
+        gov_storage::get_content_key(counter),
+        gov_storage::get_author_key(counter),
+        gov_storage::get_funds_key(counter),
+        gov_storage::get_voting_start_epoch_key(counter),
+        gov_storage::get_voting_end_epoch_key(counter),
+    ]);
+
+    // Check that expected set is a subset the actual one
+    keys.is_superset(&mandatory_keys)
```
tzemanovic, line 388: one thing I'm not sure if this VP would handle correctly is submitting a transaction that creates multiple proposals, because only the first one (with the current counter value) would be validated by this function (with the whitelist trusted txs it's not such an issue, but with custom txs it could be)

shared/src/ledger/governance/parameters.rs
```diff
@@ -0,0 +1,95 @@
...
+    Ord,
+    Hash,
+    BorshSerialize,
+    BorshDeserialize,
+)]
+/// Governance parameter structure
+pub struct GovParams {
+    /// Minimum amount of locked funds
+    pub min_proposal_fund: u64,
+    /// Maximum kilobyte length for proposal code
```
tzemanovic, line 23:
```suggestion
    /// Maximum kibibyte length for proposal code
```

shared/src/types/storage.rs
```diff
@@ -223,6 +223,15 @@ impl Key {
...
     }
 
+    /// Return the segment at the index parameter
+    pub fn get_at(&self, index: usize) -> Option<DbKeySeg> {
+        if self.segments.len() > index {
+            Some(self.segments[index].clone())
+        } else {
+            None
+        }
+    }
```
tzemanovic, line 233:
```suggestion
    pub fn get_at(&self, index: usize) -> Option<&DbKeySeg> {
        self.segments.get(index)
    }
```
(can be chained with `.cloned()` if needed)

shared/src/types/transaction/mod.rs
```diff
@@ -5,6 +5,8 @@
 pub mod decrypted;
 /// tools for encrypted data
 pub mod encrypted;
+/// txs to manage governance
```
tzemanovic, line 8: (nit) I see we've done this before, but we should really be putting the module level docs at the top of the governance.rs module, not here (module level docs start with `//!` - https://doc.rust-lang.org/rustdoc/how-to-write-documentation.html)

tzemanovic (changes requested):
this is looking great! I just added couple suggestions

apps/src/lib/client/rpc.rs
grarco: At the moment, query functions like `gov_storage::get_author_key(id)` do panic in case of an invalid `id`. They do so because we use them also in the Governance VP. Can this work or do you think we should change their behavior to, say, return an `Option`?

tzemanovic: ah, sorry, my mistake, what I meant is to say is to use the `?` for all the queries, not the storage key, e.g. `let author = query_storage_value::<Address>(client, &author_key).await?;`, so that if one of them fails, it will return early

grarco:
pls update wasm

grarco:
pls update wasm

Fraccaman:
pls update wasm

tzemanovic (approved):
LGTM, thanks for the updates!

tzemanovic:
I've added network config update for the next release https://github.com/heliaxdev/anoma-network-config/pull/9

PR #976, opened 2022-03-15
Write and get a block header

### LINKED ISSUE #974: Get the consensus state at the given height on the host chain

yito88:
closes #974


tzemanovic (approved):
I think the change looks good. I'll just do a little touch up in the `apps/src/lib/node/ledger/storage/rocksdb.rs` mod rustdoc

PR #981, opened 2022-03-16
Set counterpart channel ID

### LINKED ISSUE #980: Set the counterparty channel ID when the channel ack
Need to set the counterpart channel ID when the ack because the channel initialization doesn't set the ID

yito88:
closes #980


tzemanovic:
pls update wasm

tzemanovic (approved):

PR #989, opened 2022-03-18
encode commitment with Protobuf

### LINKED ISSUE #988: Encode the commitment with Protobuf
The proof verification function assumes that the commitment is encoded with Protobuf.

yito88:
closes #988


tzemanovic:
pls update wasm

PR #991, opened 2022-03-22
Query with the specified height

### LINKED ISSUE #987: Get the proof of a value with the specified height
The proof of a value with the specified height is required for the IBC proof verification.
Currently, we can get only the proof at the latest height.

yito88:
closes #987


tzemanovic:
pls update wasm

tzemanovic:
pls update wasm

shared/src/ledger/storage/mockdb.rs
```diff
@@ -289,11 +289,59 @@ impl DB for MockDB {
...
         Ok(self.0.borrow().get(&key.to_string()).cloned())
     }
 
+    fn read_subspace_val_with_height(
+        &self,
+        key: &Key,
+        _height: BlockHeight,
+    ) -> Result<Option<Vec<u8>>> {
+        // read the value at the latest height for the mock DB
+        self.read_subspace_val(key)
```
tzemanovic, line 342: I feel a bit hesitant about ignoring the height as this diverges from the rockdb behavior. I’d be tempted to just use `unimplemented!()` here

tzemanovic (approved):
Besides the one comment, LGTM!

shared/src/ledger/storage/mockdb.rs
yito88: Thank you for the comment! make sense! I've replaced it with `unimplemented!()`

PR #992, opened 2022-03-22
client/utils: fix init-network Tendermint node ID derivation

tzemanovic:
depends on #983 

Tendermint node ID should be derived from the standard 32-bytes encoding of ed25519 public key.

todo:
- [x] add a test case in e2e


murisi (approved):
My mistake. Before, the signing scheme tag byte in the `common::PublicKey` was corrupting the Tendermint node IDs. In the ID derivation, I should have branched on the signing scheme type as I did when computing `tm_node_keypair_json`. But given that Ed25519 is currently the only available signing scheme, these changes/simplifications look fine to me. Thanks!

tzemanovic:
No worries, it highlighted the fact that we were missing a test case covering this, so it was actually helpful :)

PR #998, opened 2022-03-24
simplify storage key rule

### LINKED ISSUE #955: simplify the protocol storage key rule
Instead of selecting storage keys that are passed to a VP invocation based on the addresses included in the keys, the protocol should just pass all the changed storage keys to every VP that is triggered. This is to avoid having a different rule for `tx_prelude::insert_verifier`, for which the invoked VP already receives all the storage keys.

tzemanovic:
closes #955


apps/src/lib/node/ledger/protocol/mod.rs
```diff
@@ -236,30 +206,39 @@ where
...
-                    keys,
-                    &verifiers_addr,
-                    vp_wasm_cache.clone(),
-                )
-                .map_err(Error::VpRunnerError),
-                Vp::Native(internal_addr) => {
+            let accept = match &addr {
+                Address::Established(_) => {
+                    let (vp, gas) = storage
+                        .validity_predicate(addr)
```
tzemanovic, line 218: note here that the VP loading (and compiling, if it's not cached already) was done sequentially in a single-thread before (in `check_vps`), but now we'll be loading them inside the parallel iterator

Fraccaman (approved):

Fraccaman:
LGTM, need to update `checksums.json` :)

PR #1000, opened 2022-03-25
Fix feature flag for masp proofs

adrianbrink:


tzemanovic:
this has been fixed in the internal testnet branch

Issue #1007, opened 2022-03-28
Non-existence proof verification failed


yito88:
We can verify the non-existence proof with another proof spec because the left or right leaf's key and value are hashed in Sparse Merkle Tree.

However, we cannot switch the proof spec in the IBC client module. The proof spec is used for both the existence and non-existence proof verification.

yito88:
The proof and the verification work well, however, sparse-merkle-tree(SMT) isn't compatible with the current IBC integration. We need to replace the Merkle tree implementation.

The problem is that the non-existence proof-verification requires a different proof spec from that of the existence proof verification.
IBC VP couldn't verify the receipt absence on the counterparty chain (when packet-timeout) because IBC VP (ibc-rs and cosmos(ibc-go)) uses the same proof spec for the existence and non-existence proof verification. It is set in an IBC client state.

- Proof spec of the leaf for the existence proof
```rust
LeafOp {
    hash: H::hash_op().into(),
    prehash_key: H::hash_op().into(),
    prehash_value: H::hash_op().into(),
    length: LengthOp::NoPrefix.into(),
    prefix: H256::zero().as_slice().to_vec(),
}
```

- Proof spec of the leaf for the non-existence proof
```rust
LeafOp {
    hash: hash_op.into(),
    prehash_key: HashOp::NoHash.into(),
    prehash_value: HashOp::NoHash.into(),
    length: LengthOp::NoPrefix.into(),
    prefix: H256::zero().as_slice().to_vec(),
}
```

The non-existence proof needs a different proof spec because sparse-merkle-tree (SMT) returns the proof with the hashed key and value of neighbors in SMT.

For the existence proof, we can replace the key and value in the proof generated by SMT with a non-hashed(original) key and value to verify it with ICS23 verification which checks if the key and value are equal to the target key and value. So, the proof spec's `LeafOp` (it says how to calculate the leaf hash) should set the `prehash_key` and `prehash_value` with the hash operation (Sha256 in this case) to calculate their hashes after the comparison.

```rust
pub struct ExistenceProof {
    #[prost(bytes = "vec", tag = "1")]
    pub key: ::prost::alloc::vec::Vec<u8>,
    #[prost(bytes = "vec", tag = "2")]
    pub value: ::prost::alloc::vec::Vec<u8>,
    #[prost(message, optional, tag = "3")]
    pub leaf: ::core::option::Option<LeafOp>,
    #[prost(message, repeated, tag = "4")]
[...]

PR #1009, opened 2022-03-29
Fix: do not replace hashed key when computing non-existence proof

karbyshev:
Add non-existence proof verification test.

### DIFF
--- a/shared/src/ledger/storage/merkle_tree.rs
+++ b/shared/src/ledger/storage/merkle_tree.rs
@@ -574,6 +574,7 @@ mod test {
         let pos_val = [2u8; 8].to_vec();
         tree.update(&pos_key, pos_val).unwrap();
         let pos_val2 = [3u8; 8].to_vec();
+        // to prevent the subtree from being empty after deletion
         tree.update(&pos_key2, pos_val2).unwrap();
 
         let specs = tree.proof_specs();
--- a/shared/src/ledger/storage/merkle_tree.rs
+++ b/shared/src/ledger/storage/merkle_tree.rs
@@ -610,10 +610,8 @@ mod test {
         // Check the base root
         assert_eq!(sub_root, tree.root().0);
 
-        // Delete the pos key
+        // Delete the pos key and verify non-existence proof for it
         tree.delete(&pos_key).unwrap();
-        assert!(tree.has_key(&ibc_key).unwrap());
-        assert!(!tree.has_key(&pos_key).unwrap());
 
         let nep = tree.get_non_existence_proof(&pos_key).unwrap();
         let subtree_nep = nep.ops.get(0).unwrap();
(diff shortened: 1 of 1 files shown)


shared/src/ledger/storage/merkle_tree.rs
```diff
@@ -617,5 +608,67 @@ mod test {
         }
         // Check the base root
         assert_eq!(sub_root, tree.root().0);
+
+        // Delete the pos key
+        tree.delete(&pos_key).unwrap();
+        assert!(tree.has_key(&ibc_key).unwrap());
+        assert!(!tree.has_key(&pos_key).unwrap());
```
yito88, line 615: These assertions are not needed for the test. Deletion has been tested in another unit test.

shared/src/ledger/storage/merkle_tree.rs
```diff
@@ -579,11 +567,14 @@ mod test {
...
             Address::Internal(InternalAddress::PoS).to_db_key().into();
         let pos_key = key_prefix.push(&"test".to_string()).unwrap();
+        let pos_key2 = key_prefix.push(&"test2".to_string()).unwrap();
 
         let ibc_val = [1u8; 8].to_vec();
         tree.update(&ibc_key, ibc_val.clone()).unwrap();
         let pos_val = [2u8; 8].to_vec();
         tree.update(&pos_key, pos_val).unwrap();
+        let pos_val2 = [3u8; 8].to_vec();
+        tree.update(&pos_key2, pos_val2).unwrap();
```
yito88, line 577: Could you add a comment?
```suggestion
        // to prevent the subtree from being empty after deletion
        tree.update(&pos_key2, pos_val2).unwrap();
```

shared/src/ledger/storage/merkle_tree.rs
```diff
@@ -273,18 +272,7 @@ impl<H: StorageHasher + Default> MerkleTree<H> {
...
-                proof: Some(Ics23Proof::Nonexist(NonExistenceProof {
-                    key: sub_key.to_string().as_bytes().to_vec(),
-                    ..nep
-                })),
-            },
-            // the proof should have a NonExistenceProof
-            _ => unreachable!(),
-        };
-        self.get_proof(key, sub_proof)
+        self.get_proof(key, cp)
```
yito88, line 275: Just a comment.
This fix is a workaround for the added test. Non-existence proof verification will still fail in IBC VP because the proof spec is the same as existence proof verification.
The issue has been opened: #1007

yito88 (commented):
Thank you! I left some comments

shared/src/ledger/storage/merkle_tree.rs
karbyshev: Fair point.

yito88:
I've investigated issue #1007 and we've decided to replace sparse merkle tree.
After the replacement, the test will be changed. So, we can close this PR for now.
The test you added was pretty helpful for us to find the issue.
Thank you so much for your help!

PR #995, opened 2022-03-23
Add integration with Geth to the ledger

batconjurer:
This PR adds the following:
 - Updates all abci++ deps to the tip of the tendermint abci++ branch
 - Adds vote extensions to the ledger
 - Adds a geth subprocess to the ledger
 - Takes ethereum headers from the process and adds them to vote extensions. 
 
 To do in later PRs:
 - Currently only headers are added to vote extensions. Logs from smart contracts are not. 
 - While we achieve consensus on the seen eth headers, these are not persisted in storage. This requires a native VP to be  written.

PR #1010, opened 2022-03-29
Verify Eth Headers

batconjurer:
This PR builds on #995 

Adds an ethash algorithm (mostly cribbed from [here](https://github.com/rust-ethereum/ethash)) to verify ethereum headers. Because the hash must use a pseudorandom cache that is
1.  expensive to compute
2. but valid for a 30000 blocks
we add a background process to precompute this cache, keeping verification times fast. 

This required refactoring the Anoma types for interacting with Ethereum types.

N.B. This requires the latest commit of Tendermint ABCI++ branch


apps/src/lib/node/ledger/shell/vote_extensions.rs
```diff
@@ -0,0 +1,197 @@
...
+            Self::try_from(VoteExtension::from(ext))
+        }
+    }
+
+    impl<D, H> Shell<D, H>
+    where
+        D: DB + for<'iter> DBIter<'iter> + Sync + 'static,
+        H: StorageHasher + Sync + 'static,
+    {
+        /// INVARIANT: This method must be stateless.
```
tzemanovic, line 48:
```suggestion
        /// INVARIANT: This method must be stateless.
        /// It uses mutable `self` only to receive eth headers.
```

Makefile
```diff
@@ -127,7 +127,10 @@ audit:
 test: test-unit test-e2e test-wasm
 
 test-e2e:
-	RUST_BACKTRACE=1 $(cargo) test e2e -- --test-threads=1
+	RUST_BACKTRACE=1 $(cargo) test e2e \
```
tzemanovic, line 130: is this change needed? It seems to work with the defaults

tzemanovic (approved):
The code changes LGTM, we need to resolve the CI before merging (and changelog pls)

Makefile
batconjurer: I did that at first to use the mock eth node, but I forgot the we compile as part of the e2e test. So it is not needed.

batconjurer:
pls update wasm

tzemanovic:
the design of the bridge has been revised and so this PR won't be used (remains open for salvaging common parts)

PR #975, opened 2022-03-15
governance: vote

Fraccaman:
- Add vote transaction
- Add governance VP logic to handle the vote
- Improve CLI proposal query command to check the status of a proposal

Based on #954

PR #1015, opened 2022-03-31
governance: add proposal execution check to vps

Fraccaman:
- Let vps check/authorize if a change to their storage space was made by a proposal.
- Add protocol parameters query

depends on #975


Fraccaman:
close in favor of #1043

PR #1033, opened 2022-04-08
Fix VP read_pre

### LINKED ISSUE #1032: VP read_pre didn't read the updated values by previous transactions
A validity predicate failed when two transactions updated the same account.
That's because `ledger::vp_env::read_pre()` read only from the storage. It couldn't read from `write_log` which had values updated by the previous transaction.

yito88:
closes #1032


tzemanovic:
pls update wasm

tzemanovic (approved):
ah, I missed this before. Nice find! ~~I think we can find a way to unify the host interface for native vp_env and WASM vp_env and will help to avoid issues like this~~ edit: actually, this is another matter

PR #1042, opened 2022-04-12
governance: tally

Fraccaman:


grarco:
pls update wasm

apps/src/lib/cli.rs
```diff
@@ -1949,6 +1976,57 @@ pub mod args {
...
+            Self {
+                query,
+                proposal_id,
+                offline,
+                proposal_folder,
+            }
+        }
+
+        fn def(app: App) -> App {
+            app.add_args::<Tx>()
```
tzemanovic, line 2007:
```suggestion
            app.add_args::<Query>()
```

apps/src/lib/client/rpc.rs
```diff
@@ -821,16 +1069,65 @@ pub async fn get_public_key(
 /// Check if the given address is a known validator.
 pub async fn is_validator(
     address: &Address,
+    epoch: Option<Epoch>,
```
tzemanovic, line 1072: I don't think we should add the epoch arg here, as the current implementation and usage is to check if a given address is a validator, but it doesn't care about whether it's in the active or inactive validator set (the active set size is limited to 128 slots, but this is done only for the consensus for Tendermint). I think even the validators that are in the inactive set can participate in governance.

apps/src/lib/client/rpc.rs
```diff
@@ -1253,3 +1551,206 @@ pub async fn query_result(_ctx: Context, args: args::QueryResult) {
...
+pub async fn compute_tally(
+    client: &HttpClient,
+    epoch: Epoch,
+    delegator_voters: &HashMap<Address, ProposalVote>,
+    validator_voters: &HashMap<Address, ProposalVote>,
+    active_validators: &[Address],
+) -> TallyResult {
+    println!("val: {:?}", validator_voters);
+    println!("del: {:?}", delegator_voters);
+    println!("act: {:?}", active_validators);
```
tzemanovic, line 1606: should we remove this?
```suggestion
```

apps/src/lib/node/ledger/mod.rs
```diff
@@ -58,6 +61,34 @@ const ENV_VAR_RAYON_THREADS: &str = "ANOMA_RAYON_THREADS";
...
+                Key::from_str(key.as_str()).expect("Key should be parsable");
+            if gov_storage::get_commit_epoch(&key).unwrap()
+                != self.storage.last_epoch.0
+            {
+                // NOTE: `iter_prefix` iterate over the matching prefix. In this
+                // case  a proposal with grace_epoch 110 will be
+                // matched by prefixes  1, 11 and 110. Thus we
+                // have to skip to the next iteration of
+                //  the cycle for all the prefixes that don't actually match
+                //  the desired epoch.
```
tzemanovic, line 81: oh, this is unfortunate - I wonder if we should instead write the proposal IDs with a fixed length string with zeroes in the prefix (maybe 20 chars to fit u64::MAX)

apps/src/lib/node/ledger/shell/finalize_block.rs
```diff
@@ -50,6 +59,177 @@ where
...
+                    .storage
+                    .read(&proposal_funds_key)
+                    .map(|(bytes, _gas)| {
+                        if let Some(bytes) = bytes {
+                            match token::Amount::try_from_slice(&bytes[..]) {
+                                Ok(amount) => amount,
+                                Err(_) => token::Amount::default(),
+                            }
+                        } else {
+                            token::Amount::default()
```
tzemanovic, line 76: also here `Error::BadProposal` as the funds are requires

docs/src/explore/design/ledger/governance.md
```diff
@@ -0,0 +1,81 @@
...
+Governance introduce two internal address with their corresponding native vps:
+- Governance address, which is in charge of validating on-chain proposals and votes
+- Treasury address, which is in charge of holding treasury funds
+
+Also, it introduces some protocol parameters:
+- `min_proposal_fund`
+- `max_proposal_code_size`
+- `min_proposal_period`
+- `max_proposal_content_size`
+- `min_grace_epoch`
```
tzemanovic, line 16:
```suggestion
- `min_proposal_grace_epochs`
```

docs/src/explore/design/ledger/governance.md
```diff
@@ -0,0 +1,81 @@
...
+and follow these rules:
+- `$id` must be equal to `counter + 1`.
+- `startEpoch` must:
+    - be grater than `currentEpoch`, where current epoch is the epoch in which the transaction is executed and included in a block
+    - be a multiple of `min_proposal_period`.
+- `endEpoch` must:
+    - be at least `min_proposal_period` epoch greater than `startEpoch`
+    - be a multiple of `min_proposal_period`
+- `graceEpoch` must:
+    - be at least `min_grace_epoch` epochs greater than `endEpoch`
```
tzemanovic, line 46:
```suggestion
    - be at least `min_proposal_grace_epochs` epochs greater than `endEpoch`
```

apps/src/lib/node/ledger/mod.rs
Fraccaman: its a little bit ugly but having a fixed size string would complicate how we handle the counter/proposal_id. I think this is fine for now

apps/src/lib/node/ledger/shell/finalize_block.rs
```diff
@@ -50,6 +59,177 @@ where
...
+                            );
+                            match tx_result {
+                                Ok(tx_result) => {
+                                    if tx_result.is_accepted() {
+                                        let author_key =
+                                            gov_storage::get_author_key(id);
+                                        self.storage
+                                            .read(&author_key)
+                                            .map(|(bytes, _gas)| {
+                                                if let Some(bytes) = bytes {
```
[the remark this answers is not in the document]
Fraccaman: its required but we are not enforcing it to be an address (so deserialization could fail). I think we should enforce the `proposal author` address existence in the governance_vp.

tzemanovic:
pls update wasm

PR #1043, opened 2022-04-12
governance: proposal execution

Fraccaman:
depends on #1042


Fraccaman:
pls update wasm

apps/src/lib/node/ledger/shell/finalize_block.rs
```diff
@@ -50,6 +60,201 @@ where
...
+                                &mut BlockGasMeter::default(),
+                                &mut self.write_log,
+                                &self.storage,
+                                &mut self.vp_wasm_cache,
+                                &mut self.tx_wasm_cache,
+                            );
+                            self.storage.delete(&pending_execution_key).expect(
+                                "Should be able to delete the storage.",
+                            );
+                            match tx_result {
```
tzemanovic, line 111: If the tx is rejected or there's an error, we have to clear out the write log (`write_log.drop_tx()`). 

We also have to call `self.gas_meter.finalize_transaction()` every time (this doesn't depend on the result), otherwise the gas used by the proposal tx would count towards the first block transaction.

apps/src/lib/node/ledger/shell/finalize_block.rs
```diff
@@ -50,6 +60,201 @@ where
...
+                            let tx = Tx::new(code, Some(encode(&id)));
+                            let tx_type =
+                                TxType::Decrypted(DecryptedTx::Decrypted(tx));
+                            let pending_execution_key =
+                                gov_storage::get_proposal_execution_key(id);
+                            // in order to filter if the vp has been triggered
+                            // by the protocol or by a transaction,
+                            // we set the following key in a protected storage
+                            // space as a flag. After the execution, we delete
+                            // it.
```
tzemanovic, line 95:
```suggestion
                            // In order to authorize proposal code to perform
                            // changes in validity predicates that allow
                            // governance changes, we set the following key in a
                            // protected storage space as a flag.
                            // We MUST delete it after the execution.
```

shared/src/ledger/governance/utils.rs
```diff
@@ -0,0 +1,339 @@
...
+        Some(bytes) => {
+            let epoched_validator_set =
+                ValidatorSets::try_from_slice(&bytes[..]).ok();
+            if let Some(epoched_validator_set) = epoched_validator_set {
+                let validator_set = epoched_validator_set.get(epoch);
+                if let Some(validator_set) = validator_set {
+                    let active_validators = validator_set
+                        .active
+                        .iter()
+                        .map(|validator| validator.address.clone())
```
tzemanovic, line 308:
```suggestion
                        .into_iter()
                        .map(|validator| validator.address)
```

shared/src/ledger/treasury/mod.rs
```diff
@@ -0,0 +1,180 @@
...
+                                        pre_balance,
+                                        post_balance,
+                                    ) {
+                                        (
+                                            Some(max_transfer_amount),
+                                            Some(pre_balance),
+                                            Some(post_balance),
+                                        ) => {
+                                            post_balance > pre_balance
+                                                || (pre_balance - post_balance
```
tzemanovic, line 132: I think this will panic with underflow when the treasure receives tokens. We can use e.g. `checked_sub` - maybe we should implement it on `token::Amount`:

```rust
impl Amount {
  pub fn checked_sub(self, rhs: Self) -> Option<Self> {
    let micro = self.micro.checked_sub(rhs.micro)?;
    Some(Self { micro })
  }
}
```

shared/src/ledger/treasury/mod.rs
```diff
@@ -0,0 +1,180 @@
...
+            match key_type {
+                KeyType::PARAMETER => {
+                    let proposal_id = u64::try_from_slice(tx_data).ok();
+                    match proposal_id {
+                        Some(id) => is_proposal_accepted(&self.ctx, id),
+                        _ => false,
+                    }
+                }
+                KeyType::BALANCE(addr) => {
+                    let proposal_id = u64::try_from_slice(tx_data).ok();
```
tzemanovic, line 72: I think maybe we could make receiving tokens permissive

tests/src/e2e/ledger_tests.rs
```diff
@@ -945,10 +946,21 @@ fn ledger_many_txs_in_a_block() -> Result<()> {
...
+/// 11. Query the proposal and check the result
+/// 12. Wait proposal grace and check proposal author funds
+/// 13. Check governance address funds are 0
 #[test]
 fn proposal_submission() -> Result<()> {
     let test = setup::network(|genesis| genesis, None)?;
 
+    let anomac_help = vec!["--help"];
+
+    let mut client = run!(test, Bin::Client, anomac_help, Some(40))?;
```
tzemanovic, line 960:
```suggestion
    // We run `--help` to make sure that the client is compiled before we
    // start-up the ledger, so that the client commands that follow are not
    // delayed by the compilation.
    let mut client = run!(test, Bin::Client, ["--help"], Some(40))?;
```

apps/src/lib/node/ledger/shell/finalize_block.rs
tzemanovic: sry, nvm `self.gas_meter.finalize_transaction()` - I forgot we're using a different meter for this

tzemanovic:
pls update wasm

PR #1045, opened 2022-04-12
Filter out StorageModification::Temp from the tx_write_log when WriteLog::commit_tx

### LINKED ISSUE #1040: IBC token VP failed when multiple transactions
When multiple transactions for IBC token transfers, `read_temp` needs to read the updated balance after the previous transactions.
~~We need `vp_read_temp_pre()` and `vp_read_temp_post()`.~~
We can filter our the temporary values from write_log when commit_tx because the temporary writes are only for the transaction, not for following transactions.

yito88:
closes #1040


tzemanovic (approved):
The change LGTM, thanks! I'm just gonna fix the formatting and update the docs in `docs/src/explore/design/ledger/wasm-vm.md:95` with `write_temp`, `read_temp_pre`, `read_temp_post`

tzemanovic:
actually, I have an alternative proposal. I think in `WriteLog::commit_tx`, we could instead filter out `StorageModification::Temp` from the `tx_write_log`, so that we can never have them in the `block_write_log`. Then any temporary key-values are in scope only for each transaction and we wouldn't need the pre/post combination. What do you think?

yito88:
@tzemanovic Thank you for the suggestion! That sounds better since the temporary writes are only for the transaction, not for other transactions. I'll fix this PR.

PR #1046, opened 2022-04-12
Reuse IBC port

### LINKED ISSUE #1011: IBC port ID cannot be reused
It should be able to be reused if the port isn't bound with any channel.

yito88:
closes #1011

### DIFF
--- a/shared/src/ledger/ibc/handler.rs
+++ b/shared/src/ledger/ibc/handler.rs
@@ -799,12 +799,18 @@ pub trait IbcActions {
 
     /// Bind a new port
     fn bind_port(&self, port_id: &PortId) -> Result<()> {
-        let index_key = storage::capability_index_key();
-        let cap_index = self.get_and_inc_counter(&index_key)?;
         let port_key = storage::port_key(port_id);
-        self.write_ibc_data(&port_key, cap_index.to_be_bytes());
-        let cap_key = storage::capability_key(cap_index);
-        self.write_ibc_data(&cap_key, port_id.as_bytes());
+        match self.read_ibc_data(&port_key) {
+            Some(_) => {}
+            None => {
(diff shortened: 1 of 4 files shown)


shared/src/ledger/ibc/handler.rs
```diff
@@ -799,12 +799,18 @@ pub trait IbcActions {
...
     /// Bind a new port
     fn bind_port(&self, port_id: &PortId) -> Result<()> {
-        let index_key = storage::capability_index_key();
-        let cap_index = self.get_and_inc_counter(&index_key)?;
         let port_key = storage::port_key(port_id);
-        self.write_ibc_data(&port_key, cap_index.to_be_bytes());
-        let cap_key = storage::capability_key(cap_index);
-        self.write_ibc_data(&cap_key, port_id.as_bytes());
+        match self.read_ibc_data(&port_key) {
+            Some(_) => {}
```
tzemanovic, line 804: do we need to check that existing channel is closed or is it fine to re-use without checking?

yito88: I think we don't need to check the channel state.
Only our ledger (module) uses these ports for now. Different channels in the module can use the same port name if the port is claimed.
Indeed, e2e test in ibc-rs [uses always the same port "transfer"](https://github.com/informalsystems/ibc-rs/blob/37dbe8c42ae09403432ebfcfdb2a8a34d86909e9/e2e/run.py#L242) for multiple channels.

tzemanovic: I see, thanks!

tzemanovic (approved):

tzemanovic:
pls update wasm

tzemanovic (approved):

tzemanovic:
pls update wasm

Issue #1049, opened 2022-04-13
Better handle errors encountered when running `anoma` binaries


james-chf:
Currently in the anoma CLIs we sometimes handle errors by calling `eprintln!` with a human-friendly error message, then exiting the process with status code 1, other times we just attempt an `.unwrap()` which panics and displays an output more helpful to developers. 

We could consolidate how we're handling errors:
- start returning `Result<()>` instead of safe exiting inside our functions; e.g. use `.await?` rather than `await.unwrap()` to return errors when they happen rather than panicking
- change existing `eprintln!` calls to returning a `Result<()>`
- at the top level, catch returned `Err(error)`s and always both `eprintln!` the `Display` trait output for users and log (i.e. `tracing::error!`) the `Debug` output for developers, as currently we may only get one or the other

This could be done nearer Namada release as part of polishing up if there is time

PR #1050, opened 2022-04-13
Improve error handling in `anomac utils join-network`

james-chf:
Some ideas for error handling, to discuss later on as part of #1049

### DIFF
--- a/apps/src/lib/client/utils.rs
+++ b/apps/src/lib/client/utils.rs
@@ -25,6 +25,7 @@ use tendermint_config::net::Address as TendermintAddress;
 use tendermint_config_abci::net::Address as TendermintAddress;
 #[cfg(feature = "ABCI")]
 use tendermint_stable::node::Id as TendermintNodeId;
+use thiserror::private::AsDynError;
 
 use crate::cli::context::ENV_VAR_WASM_DIR;
 use crate::cli::{self, args};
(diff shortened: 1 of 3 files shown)


apps/src/lib/client/utils.rs
```diff
@@ -44,161 +44,177 @@ pub const NET_OTHER_ACCOUNTS_DIR: &str = "other";
...
+                None
+            }
+        });
+        if let Some(wasm_dir) = wasm_dir.as_ref() {
+            if wasm_dir.is_absolute() {
+                eprintln!(
+                    "The arg `--wasm-dir` cannot be an absolute path. It is \
+                    nested inside the chain directory."
+                );
+                cli::safe_exit(1);
```
james-chf, line 90: this could be changed to returning something like `Err("The arg `--wasm-dir` cannot be an absolute path. It is nested inside the chain directory.")`

apps/src/lib/client/utils.rs
```diff
@@ -44,161 +44,177 @@ pub const NET_OTHER_ACCOUNTS_DIR: &str = "other";
...
-                    &base_dir,
-                    &chain_id,
-                    global_args.mode.clone(),
-                );
-                config.wasm_dir = wasm_dir;
-                config.write(&base_dir, &chain_id, true).unwrap();
-            })
-            .await
-            .unwrap();
+                    exit_if_error(config.write(&base_dir, &chain_id, true));
```
james-chf, line 209: NB: having to call an `exit_if_error` function here directly as we are running this block of code in another thread IIUC, though thinking it might be ok to run this block in the same thread as the rest of the function

apps/src/lib/client/utils.rs
```diff
@@ -666,16 +693,16 @@ pub fn init_network(
...
+                .iter()
+                .enumerate()
+                .filter_map(|(index, peer)|
+                    // we do not add the validator in its own persistent peer list
+                    if index != ix {
+                        Some(peer.to_owned())
+                    } else {
+                        None
+                    })
+                .collect();
```
james-chf, line 705: unrelated formatting change

apps/src/lib/client/utils.rs
```diff
@@ -44,161 +44,188 @@ pub const NET_OTHER_ACCOUNTS_DIR: &str = "other";
 const RELEASE_PREFIX: &str =
     "https://github.com/heliaxdev/anoma-network-config/releases/download";
 
+fn log_and_exit<E: std::error::Error>(error: E) {
+    eprintln!("ERROR: {}", error);
+    tracing::error!("{:?}", error); // ?: is using {:?} the best way to pass the error to tracing?
+    cli::safe_exit(1);
+}
+
+// TODO: can these two exit_if_error helpers be combined?
```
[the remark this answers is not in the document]
james-chf: I tried some things out with `map_err` though was getting compiler errors about returning a reference to an owned error, will have to come back to this

james-chf:
This PR won't be merged but may still push things here for discussion

PR #1054, opened 2022-04-15
Fix loading of genesis validator VP

james-chf:
Before, it was trying to load e.g. `vp_user.wasm` instead of `vp_user.<hash>.wasm`.


apps/src/lib/wasm_loader/mod.rs
```diff
@@ -26,7 +26,8 @@ pub enum Error {
     ServerError(String, String),
 }
 
-/// A hash map where keys are file names and values their expected sha256 hash
+/// A hash map where keys are simple file names and values their full file name
+/// including SHA256 hash
```
james-chf, line 30: I think it would make sense to have the checksums.json just have the hash like in the previous docstring though this docstring describes the current set up

tzemanovic (approved):

Issue #1035, opened 2022-04-11
anomac join-network should download MASP parameters directly


james-chf:
Currently users have to download MASP params separately and install them to the correct location on their machine before they can join or interact with Namada testnet. The `join-network` command should do this automatically for them.

One possible place to serve MASP params for download would be in the appropriate GitHub release on https://github.com/heliaxdev/anoma-network-config/releases

There is already some code for downloading/installing MASP params at https://github.com/anoma/masp/blob/59bf5ca2e3bb195e8f5a609f055054e6cb625199/masp_proofs/src/lib.rs#L51 - maybe this could be used? Test MASP params are already automatically downloaded and installed this way if no MASP params are present on the machine e.g.

```
~/Code/heliax/anoma/internal-testnet-n1
mbp> anoma client transfer \ 
        --ledger-address $LEDGER_ADDRESS \
        --source ${ADDRESS_NAME}-established \
        --payment-address $PAYMENT_ADDRESS \
        --token XAN \
        --amount 1000
2022-04-12T19:48:45.329651Z  INFO anoma_apps::cli::context: Chain ID: anoma-masp-0.3.51d2f83a8412b95

MASP parameters not present, downloading
```

We may need different MASP params per chain in which case we could store them somewhere specific e.g. the chain directory.

james-chf:
After more scoping and discussion, the MASP params should instead be bundled straight into the Anoma .tar.gz release (https://github.com/anoma/anoma/issues/1055)

PR #1060, opened 2022-04-19
Use the chain-specific directory for MASP params

james-chf:
Relates to #1035

The client can easily start using MASP params from the chain-specific directory (e.g. `.anoma/<chain-id>/masp/`), but the ledger cannot as it doesn't have easy access to the chain ID when verifying a shielded transaction (it only has a `Transaction` struct). Eventually the ledger won't need access to the MASP params files anymore as the verifying key should be embedded in the binary, at which point we won't need to worry about the ledger needing to access the MASP params.


james-chf:
Leaving a PR of this branch for reference but won't be merged in its current state

PR #1063, opened 2022-04-21
various wallet store and API changes

tzemanovic:
Depends on https://github.com/anoma/anoma/pull/992, this PR commits start from 630aa7c4e832a1699d75de33cab381b71ee33efd

### DIFF
--- a/apps/src/lib/wallet/store.rs
+++ b/apps/src/lib/wallet/store.rs
@@ -1,7 +1,7 @@
 use std::collections::HashMap;
 use std::fs;
 use std::io::prelude::*;
-use std::io::{self, ErrorKind, Write};
+use std::io::{self, Write};
 use std::path::{Path, PathBuf};
 use std::rc::Rc;
 use std::str::FromStr;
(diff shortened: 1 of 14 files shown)


apps/src/lib/wallet/store.rs
```diff
@@ -49,6 +49,8 @@ pub struct ValidatorData {
 
 #[derive(Serialize, Deserialize, Debug, Default)]
 pub struct Store {
+    /// The validator's address, if the wallet belongs to a validator
```
batconjurer, line 52: In general, I really don't like types that allow non-meaningful instances. We want the address and keys to both be Some or both be None and nothing in type system enforces that now. Do we really need to do this because we want to be able to set things incrementally?

tzemanovic: For genesis setup, validators use their keys, but the addresses are generated only once the genesis is being finalized, but we could use a different backing store for this "pre-genesis" wallet (I had it in my history, but I didn't get to finish it):

```rust
/// Validator pre-genesis includes all the required keys for genesis.
#[derive(Serialize, Deserialize, Debug)]
pub struct ValidatorPreGenesis {
    /// Cryptographic keypair for validator account key
    account_key: wallet::StoredKeypair,
    /// Cryptographic keypair for consensus key
    consensus_key: wallet::StoredKeypair,
    /// Cryptographic keypair for rewards key
    rewards_key: wallet::StoredKeypair,
    /// Cryptographic keypair for Tendermint node key
    tendermint_node_key: wallet::StoredKeypair,
    /// Special validator keys
    pub(crate) validator_keys: wallet::ValidatorKeys,
}
```

It's a good point and it will actually simplify error handling for the genesis validator setup a lot, so I'm in favor of fixing this before merging

tzemanovic: Thanks for this suggestion! I dropped the commit that split up the validator address from validator keys and added the new pre-genesis wallet in 415a579 and eabe7e6

batconjurer (approved):

PR #1067, opened 2022-04-26
Fix building wasm_for_tests

james-chf:
Running `cargo check` or `make` in `wasm_for_tests/wasm_source` was failing with errors about BorshSchema, this was fixed by updating `wasm_for_tests/wasm_source` to use the same patched borsh-rs crates as we do in our other crates.

Some other small changes had to be made to get things to compile, also updated some Makefiles.

`.wasm` blobs were built in a container using `cd wasm_for_tests/wasm_source && make && make opt-wasm` though ideally we won't have these in the repo at some point


tzemanovic (approved):
LGTM, thanks!

PR #1070, opened 2022-04-27
PoS spec fixes & updates

tzemanovic:


tzemanovic:
let's remove the pos spec from here and instead link to https://specs.anoma.net/master/architecture/economics/proof-of-stake.html - we'll still keep the integration spec here

tzemanovic:
moved to https://github.com/anoma/namada/pull/282

PR #1075, opened 2022-04-29
PoS spec - add new rewards and commissions system

tzemanovic:
depends and based on #1070


tzemanovic:
moved to https://github.com/anoma/namada/pull/283

PR #1090, opened 2022-05-06
PoS tx tests

### LINKED ISSUE #462: PoS transaction tests
Add tests for the PoS transaction WASMs using the `tests` crate, which provides access to tx host env and wasm tx_prelude.

With the tests, we can remove or replace the `log_string` calls in `tx_bond`, `tx_unbond` and `tx_withdraw` with a `log_string`, which will be compiled away in release built WASM.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201757896515792/1201763033477401) by [Unito](https://www.unito.io)

tzemanovic:
closes anoma/namada#22, fixes anoma/namada#164 ([5d65759](https://github.com/anoma/anoma/pull/1090/commits/5d65759db9170345a50ae95e30663852339319e4) added the same seed to confirm)


tzemanovic:
pls update wasm

tzemanovic:
I think there's an issue with the bonds in the client caused by the split of the Bond `pos_deltas` and `neg_deltas` in [bf0b686](https://github.com/anoma/anoma/pull/1090/commits/bf0b6868d45bfe707c28268b38c23b50573e3803) - the client RPC only uses the `pos_deltas`. Ideally, we should hide the implementation details somewhere behind a nicer API in PoS

tzemanovic:
> I think there's an issue with the bonds in the client caused by the split of the Bond `pos_deltas` and `neg_deltas` in [bf0b686](https://github.com/anoma/anoma/pull/1090/commits/bf0b6868d45bfe707c28268b38c23b50573e3803) - the client RPC only uses the `pos_deltas`. Ideally, we should hide the implementation details somewhere behind a nicer API in PoS

I fixed this in a rebase, for now manually in place

proof_of_stake/src/epoched.rs
```diff
@@ -544,6 +584,27 @@ where
         }
     }
```
brentstone, line 586: The function `rev_update_while`, which is replaced by the `rev_while` added here, does not seem to be used anymore. Should it be removed?

tzemanovic: good point! yeah, let's remove it

tzemanovic:
moved to https://github.com/anoma/namada/pull/254

PR #1093, opened 2022-05-10
unify native and wasm vp

tzemanovic:
closes anoma/namada#110

depends and based on #1243

To facilitate code re-use between native and WASM VPs, this PR adds `trait VpEnv` that is implemented in both. Similarly, it adds `trait TxEnv` for transaction, currently only implemented in WASM.

Because the native VPs use methods on `Ctx`, but in WASM the "context" is implicit (WASM source itself doesn't have a direct access to it and the host env functions are resolved by the runtime), we're adding a "fake" `Ctx` to both `vp_prelude` and `tx_prelude` to be able to fit them under the same interface. 

Another key difference is that in native VPs, gas handling is explicit and the env method's `Result` type is used to return early on out of gas error, whereas in WASM VPs, gas handling is not included in the source, but rather injected into the WASM before it's executed on the ledger. To unify their host env interfaces, the methods in `VpEnv` (and in `TxEnv`) use an associated `Error` type. In native VPs, the `Error` is instantiated to include out of gas error, however, in WASM the host environment functions are essentially infallible at type level (`std::convert::Infallible`) and out of gas error (or any other runtime errors for that matter) instead panic the WASM execution. However, it presented an opportunity to add more seamless error-handling for txs and VPs, so in WASM the `EnvResult` can be extended with user defined errors and messages.

The `EnvResult`'s `Error` type can be used to add a static message to errors and/or wrap any other error `E where E: std::error::Error` (<https://github.com/anoma/anoma/pull/1093/commits/bd3b901ae74758d3f8a61c1047482660d89d5378>) and the errors can be printed in debug build via `debug_log!` macro (the WASM debug build recipes are added in #1243). Transactions' and VPs' results are handled in the macros (<https://github.com/anoma/anoma/pull/1093/commits/562267deffd1cb41c257797dc3f3eafc6ca90f44>).

The `debug_log!` macro is also improve here to just `println!` to stdout in non-WASM target, e.g. unit test (<https://github.com/anoma/anoma/pull/1093/commits/76e1458a0fb79d464ba7376adb2d78327abf3327>.

This is a breaking change for WASM transactions and VPs (<https://github.com/anoma/anoma/pull/1093/commits/e90da7cba99a26b549e5afe34edf5f5fb70cf627> for the changes in the current tx and VPs), where 
- transactions now receive `ctx: &mut Ctx` as their first arg and return `TxResult` (`TxResult = EnvResult<()>`)
- VPs receive `ctx: &Ctx` as their first arg return `VpResult` (`VpResult` = `EnvResult<bool>`)

Additionally, the tx and VPs library code from `vm_env` crate has been moved to the `tx_prelude`/`vp_prelude` crates, so that only thing left in `vm_env` is the low-level host env functions API and a small helper for 2-step buffer read for  dynamically-sized data.

For WASM testing helpers, a `vp_host_env::ctx()` and `tx_host_env::ctx()` is provided (<https://github.com/anoma/anoma/pull/1093/commits/d9f0c0b4c1d68f50b83bad06daf95bc31581e1fa>).


shared/src/ledger/vp_env.rs
```diff
@@ -12,8 +13,134 @@ use crate::ledger::storage::write_log::WriteLog;
...
+/// Validity predicate's environment is available for native VPs and WASM VPs
+pub trait VpEnv {
+    /// Storage read prefix iterator
+    type PrefixIter;
+
+    /// Host functions possible error.
+    ///
+    /// In a native VP this may be out-of-gas error, however, because WASM VP is
+    /// sandboxed and gas accounting is injected by and handled in the host,
+    /// in WASM VP this error is [`std::convert::Infallible`].
```
tzemanovic, line 28:
```suggestion
    /// Host functions possible errors, extensible with custom user errors.
```

tzemanovic:
moved to https://github.com/anoma/namada/pull/256

PR #1066, opened 2022-04-25
Add EthBridge internal address

james-chf:
Relates to https://github.com/anoma/anoma/issues/1102

This branch is based off latest `master` (integrating with some changes since last release), will eventually rebase off of the next release (e.g. `v0.6.0`) before it merges

This PR adds
- `InternalAddress::EthBridge` with a VP which (currently) rejects any changes to its storage
- E2E test which verifies the EthBridge VP rejects an attempt to write to its `/queue`
- test wasm `tx_write_storage_key.wasm` used by the E2E test

### Testing
There is an e2e test to assert that no transaction can currently modify the EthBridge storage.

```shell
RUST_BACKTRACE=none ANOMA_E2E_KEEP_TEMP=true ANOMA_E2E_DEBUG=true cargo test \
    --features anoma_apps/ABCI \
    --package 'anoma_apps' \
    --package 'anoma_tests' \
    ethbridge -- \
        --test-threads=1 \
        --nocapture
```

PR #1094, opened 2022-05-11
EthBridge storage should only be modifiable by protocol transactions

james-chf:
Relates to https://github.com/anoma/anoma/issues/1102 - Depends on #1066 - this branch is based off of `james/1059/eth-bridge-vp` and this PR is targeting that branch for the time being to get a better diff.

- add a new type of protocol transaction - `EthereumBridgeUpdate(Tx)` - and allow the wrapped `Tx` to be executed
- add `tx_enqueue_eth_transfer.wasm` which modifies the `/queue` key under the Ethereum bridge account's storage - currently it just writes the data of the transaction directly to it, but in a future update it will manipulate the queue structure there properly
- make the Ethereum bridge validity predicate reject any transaction which does not have its data signed with an active validator's protocol key
- add an end-to-end test exercising a successful protocol transaction
- some end-to-end test helpers (may be moved into a separate PR)

Still to come after this PR
- protocol transactions themselves will be verified against active validator's protocol keys rather than a bundled public key as is currently the case
- transactions that modify Ethereum bridge account storage (and perhaps all protocol transactions in general) may be restricted to have to be signed by the block proposer rather than any active validator

### Testing

To run end-to-end tests, assuming you already have `wasm/`s built except for the new one(s) added in this PR, run:

```shell
export RUST_BACKTRACE=short
export ANOMA_E2E_KEEP_TEMP=true
export ANOMA_E2E_DEBUG=true
make -C wasm/wasm_source tx_enqueue_eth_transfer
make checksum-wasm
```

- [x] using ABCI

```shell
cargo test \
  eth_bridge -- \
    --test-threads=1 \
    --nocapture
```

- [ ] using ABCI++ (currently the Tendermint client fails when trying to submit a transaction with an error like "Internal error: timed out waiting for tx to be included in a block (code: -32603)")

```shell
# ensure tendermint++ is built from this commit - https://github.com/tendermint/tendermint/commit/f40884911
export TENDERMINT=${HOME}/bin/tendermint++

cargo test \
 --no-default-features \
 --features "ABCI-plus-plus" \
 eth_bridge -- \
     --test-threads=1 \
     --nocapture
```


james-chf:
pls update wasm

james-chf:
pls update wasm

james-chf:
pls update wasm

shared/src/ledger/eth_bridge/mod.rs
```diff
@@ -44,16 +49,106 @@ where
...
-            keys_changed_len = _keys_changed.len(),
-            verifiers_len = _verifiers.len(),
+        tracing::info!(
+            tx_data_len = tx_data.len(),
+            keys_changed_len = keys_changed.len(),
+            verifiers_len = verifiers.len(),
             "Validity predicate triggered",
         );
+        let signed: Signed<Vec<u8>> =
+            match Signed::<Vec<u8>>::try_from_slice(tx_data) {
```
james-chf, line 63: If submitting a protocol transaction, `tx_data` ends up being the `tx.data` field of the innermost `Tx` (i.e. the one wrapped by `EthereumBridgeUpdate(Tx)`), which should be a `Signed { data, sig }` if the protocol transaction is constructed correctly. 

In the case of submitting a regular transaction, it seems to end up being a `SignedTxData`, which I think must be the `tx.data` field of the outermost `Tx` that actually gets submitted to Tendermint RPC. 

I haven't looked closely yet at why there is this difference, it could be just to do with the way this branch is coded currently, but the VP will reject as it can't deserialize that to a `Signed`, if it gets that far.

wasm/wasm_source/src/tx_enqueue_eth_transfer.rs
```diff
@@ -0,0 +1,41 @@
...
+
+fn fatal(msg: &str, err: impl std::error::Error) -> ! {
+    log(&format!("ERROR: {} - {:?}", msg, err));
+    panic!()
+}
+
+fn fatal_msg(msg: &str) -> ! {
+    log(msg);
+    panic!()
+}
```
james-chf, line 21: I had to put a fair bit of logging while developing and unsure how much to remove, I guess it would add to gas cost

james-chf (commented):
I have a helper utility for submitting protocol transactions I can publish in another repo

james-chf:
pls update wasm

james-chf:
pls update wasm

PR #1099, opened 2022-05-13
read_wasm: return a Result instead of exiting

james-chf:
Relates to anoma/namada#104

This PR refactors `wasm_loader::read_wasm` to return a `Result` rather than exit the process if it can't read a wasm file - usages of `wasm_loader::read_wasm` have been updated to explicitly `.unwrap()` so in effect we `panic!` where before we exited the process.

When running `make test-unit` without having wasms present, the test runner can now finish and we get logs like the below, which is slightly more helpful (although doesn't explicitly indicate that wasms need to be present in order to run unit tests).

```
Location:
    apps/src/lib/wasm_loader/mod.rs:282:41', apps/src/lib/node/ledger/shell/init_chain.rs:98:26

---- node::ledger::shell::process_proposal::test_process_proposal::test_undecryptable stdout ----
thread 'node::ledger::shell::process_proposal::test_process_proposal::test_undecryptable' panicked at 'called `Result::unwrap()` on an `Err` value: Failed to read WASM from /opt/workspace/heliax/repos/anoma/tertiary/wasm/vp_user.41eb8fb2c109b6ec520b6a8bff213447d1a52daeb8057ace18dfddbbdd861ac2.wasm

Caused by:
    No such file or directory (os error 2)

Location:
    apps/src/lib/wasm_loader/mod.rs:282:41', apps/src/lib/node/ledger/shell/init_chain.rs:98:26

---- node::ledger::shell::process_proposal::test_process_proposal::test_wrapper_insufficient_balance_address stdout ----
thread 'node::ledger::shell::process_proposal::test_process_proposal::test_wrapper_insufficient_balance_address' panicked at 'called `Result::unwrap()` on an `Err` value: Failed to read WASM from /opt/workspace/heliax/repos/anoma/tertiary/wasm/vp_user.41eb8fb2c109b6ec520b6a8bff213447d1a52daeb8057ace18dfddbbdd861ac2.wasm

Caused by:
    No such file or directory (os error 2)

Location:
    apps/src/lib/wasm_loader/mod.rs:282:41', apps/src/lib/node/ledger/shell/init_chain.rs:98:26


failures:
    node::ledger::shell::finalize_block::test_finalize_block::test_mixed_txs_queued_in_correct_order
    node::ledger::shell::finalize_block::test_finalize_block::test_process_proposal_rejected_decrypted_tx
    node::ledger::shell::finalize_block::test_finalize_block::test_process_proposal_rejected_wrapper_tx
    node::ledger::shell::finalize_block::test_finalize_block::test_undecryptable_returns_error_code
    node::ledger::shell::process_proposal::test_process_proposal::test_invalid_hash_commitment
    node::ledger::shell::process_proposal::test_process_proposal::test_undecryptable
    node::ledger::shell::process_proposal::test_process_proposal::test_wrapper_insufficient_balance_address

test result: FAILED. 24 passed; 7 failed; 0 ignored; 0 measured; 0 filtered out; finished in 2.07s

error: test failed, to rerun pass '-p anoma_apps --lib'
make: *** [test-unit] Error 101

```

### DIFF
--- a/apps/src/lib/cli/context.rs
+++ b/apps/src/lib/cli/context.rs
@@ -166,7 +166,13 @@ impl Context {
 
     /// Read the given WASM file from the WASM directory or an absolute path.
     pub fn read_wasm(&self, file_name: impl AsRef<Path>) -> Vec<u8> {
-        wasm_loader::read_wasm(self.wasm_dir(), file_name).unwrap()
+        match wasm_loader::read_wasm(self.wasm_dir(), file_name) {
+            Ok(wasm) => wasm,
+            Err(err) => {
+                eprintln!("Error reading wasm: {}", err);
+                safe_exit(1);
+            }
+        }
     }
 }
 
--- a/apps/src/lib/cli/context.rs
+++ b/apps/src/lib/cli/context.rs
@@ -166,7 +166,7 @@ impl Context {
 
     /// Read the given WASM file from the WASM directory or an absolute path.
     pub fn read_wasm(&self, file_name: impl AsRef<Path>) -> Vec<u8> {
-        wasm_loader::read_wasm(self.wasm_dir(), file_name)
+        wasm_loader::read_wasm(self.wasm_dir(), file_name).unwrap()
     }
 }
 
--- a/apps/src/lib/cli/context.rs
+++ b/apps/src/lib/cli/context.rs
@@ -166,13 +166,7 @@ impl Context {
 
     /// Read the given WASM file from the WASM directory or an absolute path.
     pub fn read_wasm(&self, file_name: impl AsRef<Path>) -> Vec<u8> {
-        match wasm_loader::read_wasm(self.wasm_dir(), file_name) {
-            Ok(wasm) => wasm,
-            Err(err) => {
-                eprintln!("Error reading wasm: {}", err);
-                safe_exit(1);
-            }
-        }
+        wasm_loader::read_wasm_or_exit(self.wasm_dir(), file_name)
     }
 }
 
--- a/apps/src/lib/node/ledger/shell/init_chain.rs
+++ b/apps/src/lib/node/ledger/shell/init_chain.rs
@@ -92,9 +92,10 @@ where
             storage,
         } in genesis.established_accounts
         {
-            let vp_code = vp_code_cache
-                .get_or_insert_with(vp_code_path.clone(), || {
+            let vp_code =
+                vp_code_cache.get_or_insert_with(vp_code_path.clone(), || {
                     wasm_loader::read_wasm(&self.wasm_dir, &vp_code_path)
+                        .unwrap()
                 });
 
             // In dev, we don't check the hash
@@ -147,9 +148,10 @@ where
             balances,
         } in genesis.token_accounts
         {
-            let vp_code = vp_code_cache
-                .get_or_insert_with(vp_code_path.clone(), || {
+            let vp_code =
+                vp_code_cache.get_or_insert_with(vp_code_path.clone(), || {
                     wasm_loader::read_wasm(&self.wasm_dir, &vp_code_path)
+                        .unwrap()
                 });
 
             // In dev, we don't check the hash
@@ -191,6 +193,7 @@ where
                         &self.wasm_dir,
                         &validator.validator_vp_code_path,
                     )
+                    .unwrap()
                 },
             );
 
--- a/apps/src/lib/node/ledger/shell/init_chain.rs
+++ b/apps/src/lib/node/ledger/shell/init_chain.rs
@@ -92,11 +92,16 @@ where
             storage,
         } in genesis.established_accounts
         {
-            let vp_code =
-                vp_code_cache.get_or_insert_with(vp_code_path.clone(), || {
-                    wasm_loader::read_wasm(&self.wasm_dir, &vp_code_path)
-                        .unwrap()
-                });
+            let vp_code = match vp_code_cache.get(&vp_code_path).cloned() {
+                Some(vp_code) => vp_code,
+                None => {
+                    let wasm =
+                        wasm_loader::read_wasm(&self.wasm_dir, &vp_code_path)
+                            .map_err(Error::ReadingWasm)?;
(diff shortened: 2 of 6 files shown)


apps/src/lib/wasm_loader/mod.rs
```diff
@@ -260,67 +261,36 @@ pub async fn pre_fetch_wasm(wasm_directory: impl AsRef<Path>) {
...
-                        ) {
-                            Ok(bytes) => {
-                                return bytes;
-                            }
-                            Err(_) => {
-                                eprintln!(
-                                    "Could not read file {}. ",
-                                    file_path.as_ref().to_string_lossy()
-                                );
-                                safe_exit(1);
```
james-chf, line 298: I saw safe_exit does some flushing of stdout/stderr, in what contexts do we need this? I think any flushing could be better handled up the stack where necessary (e.g. in the case of running tests we shouldn't need it)

apps/src/lib/cli/context.rs
```diff
@@ -166,7 +166,7 @@ impl Context {
 
     /// Read the given WASM file from the WASM directory or an absolute path.
     pub fn read_wasm(&self, file_name: impl AsRef<Path>) -> Vec<u8> {
-        wasm_loader::read_wasm(self.wasm_dir(), file_name)
```
tzemanovic, line 169: I think in here we could turn it into a more user-friendly error instead of panic

apps/src/lib/node/matchmaker.rs
```diff
@@ -142,7 +142,7 @@ impl Runner {
         // Prepare a client for intent gossiper node connection
         let (listener, dialer) = ClientListener::new_pair(intent_gossiper_addr);
 
-        let tx_code = wasm_loader::read_wasm(&wasm_dir, tx_code_path);
+        let tx_code = wasm_loader::read_wasm(&wasm_dir, tx_code_path).unwrap();
```
tzemanovic, line 145: in here it's fine - it's going to change anyway

apps/src/lib/wasm_loader/mod.rs
```diff
@@ -260,67 +261,36 @@ pub async fn pre_fetch_wasm(wasm_directory: impl AsRef<Path>) {
 pub fn read_wasm(
```
tzemanovic, line 261: maybe this could be `try_read_wasm` and we could provide another `read_wasm` from it that handles the error and exits

apps/src/lib/node/ledger/shell/init_chain.rs
```diff
@@ -92,9 +92,10 @@ where
             storage,
         } in genesis.established_accounts
         {
-            let vp_code = vp_code_cache
-                .get_or_insert_with(vp_code_path.clone(), || {
+            let vp_code =
+                vp_code_cache.get_or_insert_with(vp_code_path.clone(), || {
                     wasm_loader::read_wasm(&self.wasm_dir, &vp_code_path)
+                        .unwrap()
```
tzemanovic, line 98: I guess in here the error should really propagate up to the `init_chain`'s Result

apps/src/lib/wasm_loader/mod.rs
james-chf: Yeah it'd be good to consolidate our calls to `safe_exit` to somewhere further up the stack where it's less surprising, at some point could take a look at this across our codebase

apps/src/lib/node/ledger/shell/init_chain.rs
james-chf: Yeah it would be good if this could propagate up, I had a go at this though it was more involved than I thought at a glance
- changed the code to reading the wasm outside of the closure so we can return the `Result` if there is an error
- had to make a new `shell::Error::ReadingWasm` variant to wrap the `eyre::Error`
There are a couple more occurrences in `init_chain` I could update to use the same but could be worth doing all these changes in a separate PR (I think a `CodeCache` component wrapping `HashMap<String, Vec<u8>>` and handling wasm loading could be useful)

apps/src/lib/wasm_loader/mod.rs
james-chf: I'd be more inclined to use "read_wasm" for this here returning a `Result` then "read_wasm_or_exit" for the one which exits, what do you think?

At the moment we have many places in the code where the process can terminate, most of the places where this happens I don't think should be the concern of that component e.g. should the `wasm_loader` module ever be deciding to terminate the process or not?  So eventually we could have most calls returning `Result<T>` rather than returning `T` or maybe exiting altogether, in that case we could want to make explicit where exiting might happen with a suffix/prefix in the name rather than the other way round, if we followed an idiom like: returning a `Result<T>` means this function won't suddenly exit. Returning a `T` it shouldn't suddenly exit either unless it has a string in the name like `or_exit` (I'm not sure how to encode this in the type signature)

apps/src/lib/cli/context.rs
james-chf: There are 20 calls to this method so for now I've changed it back to the previous behaviour here (`eprintln!` + `safe_exit`), `make test-unit` still finishes so guess it wasn't a call to this that was causing it to terminate

james-chf:
Thanks for reviewing, have acted on some of the comments

apps/src/lib/node/ledger/shell/init_chain.rs
tzemanovic: Thanks! Yeah, let's refactor the rest of them separately. Btw, I didn't know about the `#[from]` macro, very convenient!

james-chf:
I just rebased this onto v0.6.0

apps/src/lib/wasm_loader/mod.rs
james-chf: I've added `read_wasm_or_exit` and used it in a couple of places where we were previously panicking or exiting. Ultimately this should become something like `read_wasm_or_panic` I think (or go away altogether) but that can happen in another PR at some point

tzemanovic:
moved to https://github.com/anoma/namada/pull/259

PR #1103, opened 2022-05-19
rust 1.60.0

yito88:
I want to update Rust toolchain because `ibc-rs` v0.14.0 requires it


rust-nightly-version
```diff
@@ -1 +1 @@
-nightly-2022-01-27
+nightly-2022-04-13
```
james-chf, line 1: how do we choose which nightly version to use?

james-chf (commented):
[1.61](https://blog.rust-lang.org/2022/05/19/Rust-1.61.0.html) came out today if we are feeling brave!

rust-nightly-version
yito88: Good question! I chose the latest version at that time when I tried updating Rust toolchain.

james-chf (commented):
there is also `wasm/rust-toolchain.toml`

rust-nightly-version
james-chf: should we use a more recent one when this PR lands? eg 2022-05-28. I will check it has components for aarch64-apple-darwin as well

tzemanovic: I'm updating this to 1.61.0. I also tried later version of nightly, but ran into this https://github.com/rust-lang/rust-clippy/issues/8867 with some PoS types with generics. `2022-05-20` seems to be the last ok one

tzemanovic:
merged in 83b5ff25f26483bec0f3099406152613c89b7f7f

Issue #1065, opened 2022-04-22
IBC acknowledgement and FungibleTokenData are incompatible with ibc-go


yito88:
Gaia couldn't unmarshal them made by IBC handler

PR #1106, opened 2022-05-20
Update ibc-rs to v0.14.0

yito88:
depends on #1065


shared/src/ledger/ibc/vp/channel.rs
```diff
@@ -236,6 +241,13 @@ where
...
     ) -> Result<()> {
         let prev_channel = self.channel_end_pre(port_channel_id)?;
+        let conn_id = channel.connection_hops().get(0).ok_or_else(|| {
+            Error::InvalidChannel(format!(
+                "No connection for the channel: Port/Channel {}",
+                port_channel_id,
+            ))
+        })?;
+        let counterparty = channel.counterparty();
         match channel.state() {
```
batconjurer, line 251: There is a lot of repeated code here. Can you instead do
`let msg = match channel.state() { ...`
and then after the match do 
```
let event = make_open_ack_channel_event(
                   &msg,
                   conn_id,
                   counterparty,
              );
 self.check_emitted_event(event)
       .map_err(|e| Error::IbcEvent(e.to_string()))
 ```

File: shared/src/ledger/ibc/handler.rs [hunk 2]
```diff
@@ -457,15 +455,23 @@ pub trait IbcActions {
...
+        let conn_id = channel.connection_hops().get(0).ok_or_else(|| {
+            Error::Channel(format!(
+                "No connection for the channel: Port/Channel {}",
+                port_channel_id,
+            ))
+        })?;
+        let counterparty = channel.counterparty();
+        let event = make_open_ack_channel_event(msg, conn_id, counterparty)
+            .try_into()
+            .unwrap();
```
batconjurer, line 474: This code block gets used a lot. Can we split if off into a separate function?

batconjurer (changes requested):

shared/src/ledger/ibc/vp/channel.rs
yito88: Thank you! I've tried cleaning the match block to check the message then the state changes.

batconjurer (approved):

tzemanovic:
moved to https://github.com/anoma/namada/pull/260

PR #1111, opened 2022-05-26
Improve governance

Fraccaman:
- Fix an overflow bug when computing voting power tally
- Add validation of proposal .json when creating a new proposal
- Fix a bug when computing total amount of nay/yay
- Improving CLI UI


grarco:
Also fixes #1113 by using [tempfile](https://crates.io/crates/tempfile)

.gitignore
```diff
@@ -27,3 +27,8 @@ wasm/*.wasm
 
 # app version string file
 apps/version.rs
+
+# Governance artifacts
```
tzemanovic, line 31: let's not do this - any files created by test(s) should be inside a temp dir

grarco: Yes, it's actually what we are now doing with 147cf381fc9fef7a05e03ddfff1de031aee99f4c, I'll revert the gitignore, thanks!

Fraccaman:
pls update wasm

Fraccaman:
pls restart ci

tzemanovic:
I'm thinking we could maybe speed up the e2e proposal_submission test - in the setup (`let test = setup::network(|genesis| genesis, None)?;`), we could modify the genesis to have shorter `min_proposal_period` and `min_proposal_grace_epochs`

apps/src/lib/client/tx.rs
```diff
@@ -532,7 +531,59 @@ pub async fn submit_init_proposal(mut ctx: Context, args: args::InitProposal) {
...
+                == 0
+        );
+        eprintln!(
+            "Invalid proposal start epoch: {} must be greater than current \
+             epoch {} and a multiple of {}",
+            proposal.voting_start_epoch,
+            current_epoch,
+            goverance_parameters.min_proposal_period
+        );
+        safe_exit(1)
```
tzemanovic, line 562: let's skip the `safe_exit` call `if args.tx.force` is set (same at the other calls below)

apps/src/lib/client/tx.rs
```diff
@@ -675,6 +727,14 @@ pub async fn submit_vote_proposal(mut ctx: Context, args: args::VoteProposal) {
 
         match proposal_start_epoch {
             Some(epoch) => {
+                if current_epoch < epoch {
+                    eprintln!(
+                        "Current epoch {} is not greater than proposal start \
+                         epoch {}",
+                        current_epoch, epoch
+                    );
+                    safe_exit(1)
```
tzemanovic, line 736: also `force` check here and below

apps/src/lib/node/ledger/shell/finalize_block.rs
```diff
@@ -60,177 +51,8 @@ where
...
-                self.storage.transfer(
-                    &m1t(),
-                    funds,
-                    &gov_address,
-                    &transfer_address,
-                );
-            }
-        }
+        let _proposals_result =
+            execute_governance_proposals(self, new_epoch, &mut response)?;
```
tzemanovic, line 55: nice, thank you for moving this out!

tzemanovic: I will do the same with PoS slashing

tzemanovic (approved):

tzemanovic (approved):
LGTM, just couple minor changes. Thanks!

grarco (approved):

apps/src/lib/node/ledger/shell/governance.rs
```diff
@@ -0,0 +1,195 @@
...
+            .read_storage_key::<Epoch>(&proposal_start_epoch_key)
+            .ok_or_else(|| {
+                Error::BadProposal(
+                    id,
+                    "Invalid proposal start_epoch.".to_string(),
+                )
+            })?;
+
+        let votes =
+            get_proposal_votes(&shell.storage, proposal_start_epoch, id);
```
tzemanovic, line 60: should the epoch given to `get_proposal_votes` and `compute_tally` (here and other places where we're using these) be end epoch? (https://specs.anoma.net/master/architecture/governance/signaling.html#tally says it should be `end_epoch`)

tzemanovic:
In `apps/src/lib/client/rpc.rs`, the offline tally `get_proposal_offline_votes` is missing slashes for delegator's bonds. I think we need something like this: https://github.com/anoma/anoma/commit/dad75398b30539e2da8e15db970bdcc2e32ee094. I also refactored the last bit to update the maps with https://doc.rust-lang.org/std/collections/hash_map/struct.HashMap.html#method.entry

Ideally, this should really be hidden behind PoS API as the data from PoS is hard to query atm. I think we can solve it in future with some common interface for storage direct access and client RPC queries

Issue #472, opened 2021-09-16
wallet address look-ups in client results


tzemanovic:
For client commands that print out some outputs that include bech32m encoded address(es), it would be nice to try to reverse look-up these from the wallet's address book and show the aliases instead when available. For this, we might want to switch the address book in the `wallet::store` from regular `HashMap` to a bijective map (e.g. https://crates.io/crates/bimap)

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201757896515792/1201763115228805) by [Unito](https://www.unito.io)

PR #1138, opened 2022-06-08
Changed hash-map to BiHashMap 

bengtlofgren:
In the wallet in order to be able to retrieve alias from address.

I think this is a good idea for UI, since atm we are spitting out hashed addresses that we don't know correspond to any account

### DIFF
--- a/.changelog/unreleased/improvements/1138-change-wallet-bihashmap.md
+++ b/.changelog/unreleased/improvements/1138-change-wallet-bihashmap.md
@@ -1,0 +1,4 @@
+- Allows simple retrival of aliases from addresses in the wallet without
+  the need for multiple hashmaps. This is the first step to improving the
+  UI if one wants to show aliases when fetching addresses from anoma wallet
+  ([#1138](https://github.com/anoma/anoma/pull/1138))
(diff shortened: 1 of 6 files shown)


tzemanovic:
nice! there's #472 for this :)

brentstone:
I'm wondering if changing to BiHashMap is even needed... is there a way to convert the `address` into a `PublicKeyHash` and then use `Store::pkhs` HashMap to get the `alias`?

james-chf:
> I'm wondering if changing to BiHashMap is even needed... is there a way to convert the `address` into a `PublicKeyHash` and then use `Store::pkhs` HashMap to get the `alias`?

Commenting here based on discussion earlier - the bech32m address (e.g. `atest1v4ehgw36g56ngwpk8ppnzsf4xqeyvsf3xq6nxde5gseyys3nxgenvvfex5cnyd2rx9zrzwfctgx7sp`) is derived from the "hash" of an account, for an implicit account this "hash" is the `PublicKeyHash` but for established accounts it's something random, I think using the `BiHashMap` here should be the most straightforward option

tzemanovic:
I rebased the branch onto v0.6.1 to check CI

tzemanovic:
moved to https://github.com/anoma/namada/pull/264

PR #1145, opened 2022-06-09
V0.6.0 rc2

tzemanovic:


tzemanovic:
there're still some e2e tests failing (see #1144)

PR #1151, opened 2022-06-10
Temp: rebase #1088 - update ABCI++ 

tzemanovic:
this is temp to check rebase of #1088


tzemanovic:
pls update wasm

tzemanovic:
There seems to be an issue in e2e node connectivity in ABCI++ with the selected tendermint go commit. It looks like in this commit, the p2p config is missing addrbook-strict-mode, which needs to be off for this test. It looks like it’s later been re-introduced https://github.com/tendermint/tendermint/blob/abci%2B%2B/config/config.go#L639

Issue #993, opened 2022-03-22
`anoma wallet address find` should provide a hint to the user of the missing `--alias` flag


gnosed:
When executing `anoma wallet address find` the following is output:
```bash
gnosed@mastoc:~/anoma$ anoma wallet address find
The application panicked (crashed).
Message:  called `Option::unwrap()` on a `None` value
Location: apps/src/lib/cli/utils.rs:173

Backtrace omitted.
Run with RUST_BACKTRACE=1 environment variable to display it.
Run with RUST_BACKTRACE=full to include source snippets.
```
 This message doesn't give any indication about the source of error. Should output a hint about the missing `--alias` flag which is required here.

PR #1095, opened 2022-05-11
e2e tests: switch from rexpect to expectrl

tzemanovic:
closes #686

depends and based on #1071

PR #1161, opened 2022-06-17
Fix output of `anoma wallet address find`

brentstone:
Fixes [#993](https://github.com/anoma/namada/issues/91). Based and depends on anoma/anoma#1138 

 The output is now:

image

### DIFF
--- a/apps/src/lib/cli/utils.rs
+++ b/apps/src/lib/cli/utils.rs
@@ -17,8 +17,16 @@ pub trait Cmd: Sized {
     fn parse(matches: &ArgMatches) -> Option<Self>;
 
     fn parse_or_print_help(app: App) -> (Self, Context) {
+        println!("Inside parse_or_print_help(app)");
         let mut app = Self::add_sub(app);
+        println!("Added subs again");
+
+        println!("{:?}",app);
+        
+        dbg!(app.clone().get_matches());
+
         let matches = app.clone().get_matches();
+        println!("Parse or print help");
         match Self::parse(&matches) {
             Some(cmd) => {
                 let global_args = args::Global::parse(&matches);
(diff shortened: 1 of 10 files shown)


Fraccaman:
I think `--alias` should be optional as we might introduce other arguments (like --address, which would do the opposite)

brentstone:
> I think `--alias` should be optional as we might introduce other arguments (like --address, which would do the opposite). I think this should be `at least one arg` (using clap `ArgGroup`)

In this particular case, would we want `--address` to return the `alias`, or should searching for aliases be its own new subcommand?

Fraccaman:
I think `--address` should return the alias

apps/src/lib/cli/utils.rs
```diff
@@ -170,6 +170,7 @@ where
     <T as FromStr>::Err: Debug,
 {
     pub fn parse(&self, matches: &ArgMatches) -> T {
+        println!("I'm parsing!");
```
Fraccaman, line 173: I think we can remove this log

Fraccaman, line 173: I think we can remove this log

Fraccaman (dismissed):

Fraccaman:
Looking great :) requires changelog

brentstone:
CI build is now failing on e2e tests after only removing a println! statement and adding the changelog. Is it possible something else changed in the CI? Not sure how the last two commits could have changed the behavior of the build.

tzemanovic:
> CI build is now failing on e2e tests after only removing a println! statement and adding the changelog. Is it possible something else changed in the CI? Not sure how the last two commits could have changed the behavior of the build.

Yeah, there seem to be some flakiness with e2e tests (possibly caused by expectrl switch #1095) and occasionally sccache timeouts :/

tzemanovic:
moved to https://github.com/anoma/namada/pull/276

Issue #162, opened 2021-04-29
add `secp256k1` support


tzemanovic:
add `secp256k1` crypto support to the wasm host env

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201757896515792/1201762871688692) by [Unito](https://www.unito.io)

tzemanovic:
an older wip version is in https://github.com/anoma/anoma/commits/nat/secp256k1-integration

tzemanovic:
woops, this issue shouldn't have been linked to #905 anymore

PR #1163, opened 2022-06-17
Adding support for secp256k1 keys

brentstone:
Closes [#162](https://github.com/anoma/namada/issues/4), currently based off of v0.6.0. This issue was also addressed in anoma/anoma#914, but that PR is based off an older version of the code and seems to have stalled.


brentstone:
Should there be additional tests written to validate the use of these keys elsewhere in the code? @murisi @tzemanovic

apps/src/lib/client/utils.rs
```diff
@@ -357,11 +358,12 @@ pub async fn join_network(
 const TENDERMINT_NODE_ID_LENGTH: usize = 20;
 
 /// Derive Tendermint node ID from public key
-fn id_from_pk(pk: &ed25519::PublicKey) -> TendermintNodeId {
-    let digest = Sha256::digest(pk.try_to_vec().unwrap().as_slice());
+fn id_from_pk(pk: &common::PublicKey) -> Result<TendermintNodeId, ParsePublicKeyError> {
```
tzemanovic, line 361: I'm not sure if we need this `Result` layer since the error is not used

apps/src/lib/client/utils.rs
```diff
@@ -263,11 +264,11 @@ pub async fn join_network(
     if let Some((validator_alias, pre_genesis_wallet)) =
         validator_alias_and_pre_genesis_wallet
     {
-        let tendermint_node_key: ed25519::SecretKey = pre_genesis_wallet
+        let tendermint_node_key: common::SecretKey = pre_genesis_wallet
             .tendermint_node_key
             .try_to_sk()
             .unwrap_or_else(|_err| {
-                eprintln!("Tendermint node key must be ed25519");
+                eprintln!("Tendermint node key must be common (need to change?)");
```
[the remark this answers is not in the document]
tzemanovic: ah, it might actually work with secp256k1, but we'll need to test that (we just cannot use the `common` signing scheme directly with tendermint as it has different encoding)

apps/src/lib/client/utils.rs
tzemanovic, line 361: the `common::PublicKey` is not going work with Tendermint, we need to convert it to the underlying scheme before we encode it (`common::PublicKey` encoding has an extra byte prefix for the enum case)

shared/src/types/key/secp256k1.rs
```diff
@@ -0,0 +1,488 @@
...
+        S: Serializer,
+    {
+        // not sure if this is how I should be doing this!
+        // https://serde.rs/impl-serialize.html!
+        let arr = self.0.serialize();
+        let mut seq = serializer.serialize_tuple(arr.len())?;
+        for elem in &arr[..] {
+            seq.serialize_element(elem)?;
+        }
+        seq.end()
```
tzemanovic, line 175: I think once you have the `arr`, you can use its `serde::Serialize` implementation with e.g.:

``
```suggestion
        serde::Serialize::serialize(&arr, serializer)
```

shared/src/types/key/secp256k1.rs
```diff
@@ -0,0 +1,488 @@
...
+                #[allow(clippy::needless_range_loop)]
+                for i in 0..libsecp256k1::util::SECRET_KEY_SIZE {
+                    arr[i] = seq.next_element()?
+                        .ok_or_else(|| Error::invalid_length(i, &self))?;
+                }
+                Ok(arr)
+            }
+        }
+
+        let arr_res = deserializer.deserialize_tuple(libsecp256k1::util::SECRET_KEY_SIZE, ByteArrayVisitor)?;
```
tzemanovic, line 207: Then you can deserialize this back from `[u8; 32]` without the visitor with e.g.:
```suggestion
        let arr_res: [u8; libsecp256k1::util::SECRET_KEY_SIZE] = serde::Deserialize::deserialize(deserializer)?;
```

tzemanovic: similar for the `Signature` below

apps/src/lib/client/utils.rs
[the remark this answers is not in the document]
brentstone: Discussed yesterday with @tzemanovic : going to leave `tendermint_node_key` as a `common::SecretKey` and then instead change to a particular variant inside `fn id_from_pk(pk: &common::PublicKey)` before returning the `TenderminNodeId`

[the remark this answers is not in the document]
brentstone: Actually, why do we have here `tendermint_node_key` as a `SecretKey` but then we load it into `id_from_pk()` which takes a `PublicKey`? Are there foreseeable issues with this?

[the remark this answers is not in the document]
murisi: Isn't `tendermint_node_key` as `SecretKey` because it is used as a parameter to `write_tendermint_node_key` on line 305 of `utils.rs`? I believe the `ref_to` (implemented in each signing scheme) call safely handles the conversion from `SecretKey` to `PublicKey`...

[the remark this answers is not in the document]
brentstone: Ah of course, I overlooked that. Thx

apps/src/lib/client/utils.rs
```diff
@@ -357,7 +358,7 @@ pub async fn join_network(
 const TENDERMINT_NODE_ID_LENGTH: usize = 20;
 
 /// Derive Tendermint node ID from public key
-fn id_from_pk(pk: &ed25519::PublicKey) -> TendermintNodeId {
+fn id_from_pk(pk: &common::PublicKey) -> TendermintNodeId {
```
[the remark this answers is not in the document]
murisi: See here the code that validates Ed25519 keys in Tendermint: https://github.com/tendermint/tendermint/blob/master/crypto/ed25519/ed25519.go#L102 . Note how it is checking if the public key bytes are concatenated...

apps/src/lib/client/utils.rs
```diff
@@ -359,9 +359,20 @@ const TENDERMINT_NODE_ID_LENGTH: usize = 20;
...
+            let _pk: ed25519::PublicKey = pk.try_to_pk().unwrap();
+            let digest = Sha256::digest(_pk.try_to_vec().unwrap().as_slice());
+            bytes.copy_from_slice(&digest[..TENDERMINT_NODE_ID_LENGTH]);
+        },
+        common::PublicKey::Secp256k1(_) => {
+            let _pk: secp256k1::PublicKey = pk.try_to_pk().unwrap();
+            let digest = Sha256::digest(_pk.try_to_vec().unwrap().as_slice());
+            bytes.copy_from_slice(&digest[..TENDERMINT_NODE_ID_LENGTH]);
+        },
+    }
```
brentstone, line 375: Is there a more succinct way to write this code for the conversion from common to one of its variants?

murisi: Could the following work?:
```
    match pk {
        common::PublicKey::Ed25519(_pk) => {
            let digest = Sha256::digest(_pk.try_to_vec().unwrap().as_slice());
            bytes.copy_from_slice(&digest[..TENDERMINT_NODE_ID_LENGTH]);
        },
        common::PublicKey::Secp256k1(_pk) => {
            let digest = Sha256::digest(_pk.try_to_vec().unwrap().as_slice());
            bytes.copy_from_slice(&digest[..TENDERMINT_NODE_ID_LENGTH]);
        },
    }
```
I.e. Somehow use pattern matching to obtain the contents you need instead of using `try_` functions?

murisi: And if the above works and the borrow checker allows, you could try:
```
    let raw_bytes = match pk {
        common::PublicKey::Ed25519(_pk) => _pk.try_to_vec(),
        common::PublicKey::Secp256k1(_pk) => _pk.try_to_vec(),
    }
    let digest = Sha256::digest(raw_bytes.unwrap().as_slice());
    bytes.copy_from_slice(&digest[..TENDERMINT_NODE_ID_LENGTH]);
```

apps/src/lib/client/utils.rs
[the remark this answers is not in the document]
brentstone: Thanks, this doesn't look like it is required in https://github.com/tendermint/tendermint/blob/master/crypto/secp256k1/secp256k1.go, which seems consistent with `PrivKeySize` in https://pkg.go.dev/github.com/tendermint/tendermint/crypto/secp256k1#pkg-constants, so I will pass the secret key itself without concatenation in the `Secp256k1` case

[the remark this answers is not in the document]
brentstone: Resolved with [1522aca](https://github.com/anoma/anoma/pull/1163/commits/1522aca03a3d20ac80ad24d782914f68f4bb9f25)

apps/src/lib/client/utils.rs
```diff
@@ -1142,16 +1170,26 @@ fn network_configs_url_prefix(chain_id: &ChainId) -> String {
...
+    node_sk: common::SecretKey,
+) -> common::PublicKey {
+    let node_pk: common::PublicKey = node_sk.ref_to();
+
+    // Convert and write the keypair into Tendermint node_key.json file.
+    // Tendermint requires concatenating the private-public keys for ed25519
+    // but does not for secp256k1.
+    let (node_keypair, key_str) = match node_sk {
+        common::SecretKey::Ed25519(_) => {
+            ([node_sk.try_to_vec().unwrap(), node_pk.try_to_vec().unwrap()].concat(),
```
murisi, line 1182: I think `node_sk.try_to_vec()` would produce a vector with a prefix byte indicating the type of key which Tendermint would not accept. Maybe something along the following lines would be better:
```
    let (node_keypair, key_str) = match node_sk {
        common::SecretKey::Ed25519(node_sk) => {
            ([node_sk.try_to_vec().unwrap(), node_sk.ref_to().try_to_vec().unwrap()].concat(),
            "Ed25519")
        },
        common::SecretKey::Secp256k1(node_sk) => {
            (node_sk.try_to_vec().unwrap(), "Secp256k1")
        },
    };
```

brentstone: Do you mean `node_pk.try_to_vec()`? The concatenation line you suggested concatenates `node_sk` to itself. Just wondering if the `node_sk` or `node_pk` should have a `.ref_to()`.

murisi: Sorry, I keep missing the mistake that I made in the concatenation line... I guess the first part `node_sk.try_to_vec().unwrap()` represents the  byte encoding of an `ed25519::SecretKey` and the second part `node_sk.ref_to().try_to_vec().unwrap()` represents the byte encoding of `node_sk.ref_to()` which I guess is an `ed25519::PublicKey`. No?

brentstone: Ahh ok sorry, I understand what is happening here now in your suggestion. Yes, you are correct, I'll try to check if what you say about `node_sk.try_to_vec()` having an prefix byte from `common` is true. Thx for the suggestion

brentstone:
pls update wasm

tzemanovic:
moved to https://github.com/anoma/namada/pull/278

PR #1166, opened 2022-06-23
Zeroize secret keys upon drop

brentstone:
Closes issue anoma/namada#82. It seems the best way to get the memory of the `SecretKey` byte array to be reset to `0`s is to wrap the external secret key structs (`ed25519_consensus::SigningKey` for example) in a `Box` pointer when initializing our own `SecretKey` structs.

I found some inspiration from [this blog post](https://benma.github.io/2020/10/16/rust-zeroize-move.html), which described behavior I was observing while experimenting with this code.


brentstone:
pls update wasm

tzemanovic (approved):
LGTM, verified tests locally - the last CI run was flakey

tzemanovic:
moved to https://github.com/anoma/namada/pull/277

Issue #1171, opened 2022-06-24
Add shared types and functions to Ethereum bridge integration branch


james-chf:
Based on changes being made in https://github.com/anoma/specs/pull/104. Any types or functions that will be used in both `PrepareProposal` and `FinalizeBlock` can be put in the integration branch now and maintained there, to be used in WIP branches off of that. Some of these types or types approximating them already exist in various branches, where not they should be defined in the PR and the spec updated. The types will include at least

- `EthereumEvent` type (see `bat/full-eth-node` - at least `EthereumEvent::TransfersToNamada` variant))
- `ProtocolTransaction` that includes Ethereum events from vote extensions (something similar to `ProtocolTxType::EthereumHeaders` in `bat/eth-header-vp`).

Some groundwork to start accepting protocol transactions in Anoma in general may also be needed (as a separate PR).

Issue #1101, opened 2022-05-16
Be able to submit protocol transactions in our end-to-end test


james-chf:
`anomac` which we use for submitting transactions in our e2e tests is not able to send protocol transactions and probably it doesn't make sense for it to have that functionality, since in practice those should always be broadcast by validators running `anoman`. It would be useful to have a way to submit protocol transactions in another way for testing.

We can use `TendermintWebsocketClient` (although it is being deprecated) to build a protocol transaction in memory and submit it to the Tendermint RPC endpoint. We should already have the e2e test node validators' protocol keys available to us in the e2e tests.

This issue is motivated by #1094 which needs this sort of functionality for testing.

james-chf:
I put some functionality in https://github.com/anoma/devtool for submitting protocol transactions manually to a chain. End-to-end tests for submitting a protocol transaction to Tendermint are working for ABCI in https://github.com/anoma/anoma/pull/1094. Submitting protocol transactions this way to modify the Ethereum bridge may not be applicable anymore in which case the generic functionality for submitting a protocol transaction within an e2e test can be factored out of that PR into a new PR if needed. There is a new Tendermint client coming in https://github.com/anoma/anoma/pull/1088 that should be used in such tests instead of `TendermintWebsocketClient`

Issue #1167, opened 2022-06-23
Internal transaction to mint wrapped Ethereum assets


james-chf:
- should work similarly to `tx_mint_multitoken` in https://github.com/anoma/anoma-wasm-multitoken
- the transaction will be crafted internally by the ledger, rather than submitted via Tendermint

PR #1172, opened 2022-06-24
Start accepting EthereumEvents protocol transaction

james-chf:
Relates to https://github.com/anoma/anoma/issues/1171. Adapting code from https://github.com/anoma/anoma/tree/james/1059/eth-bridge-allow-queue-changes and https://github.com/anoma/anoma/tree/bat/eth-header-vp

After this PR, we can implement https://github.com/anoma/anoma/issues/1101 at some point

### DIFF
--- a/apps/src/lib/node/ledger/protocol/mod.rs
+++ b/apps/src/lib/node/ledger/protocol/mod.rs
@@ -15,6 +15,7 @@ use anoma::ledger::treasury::TreasuryVp;
 use anoma::proto::{self, Tx};
 use anoma::types::address::{Address, InternalAddress};
 use anoma::types::storage;
+use anoma::types::transaction::protocol::{ProtocolTx, ProtocolTxType};
 use anoma::types::transaction::{DecryptedTx, TxResult, TxType, VpsResult};
 use anoma::vm::wasm::{TxCache, VpCache};
 use anoma::vm::{self, wasm, WasmCacheAccess};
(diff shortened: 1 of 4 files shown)


apps/src/lib/node/ledger/protocol/mod.rs
```diff
@@ -120,7 +119,24 @@ where
...
+                .finalize_transaction()
+                .map_err(Error::GasError)?;
+            Ok(TxResult {
+                gas_used,
+                ..Default::default()
+            })
+        }
+        TxType::Protocol(_)
+        | TxType::Wrapper(_)
+        | TxType::Decrypted(DecryptedTx::Undecryptable(_)) => {
```
james-chf, line 139: I expanded the `_` here to make explicit what we are executing this match arm for, it looks like a sort of noop but I'm not sure - the previous docstring mentioned the only tx types that should be input to this fn are `Decrypted` and `Wrapper`, so should we be erroring here instead?

shared/src/types/transaction/protocol.rs
```diff
@@ -76,9 +76,8 @@ mod protocol_txs {
         DKG(DkgMessage),
         /// Tx requesting a new DKG session keypair
         NewDkgKeypair(Tx),
-        /// Aggregation of Ethereum state changes
-        /// voted on by validators in last block
-        EthereumStateUpdate(Tx),
+        /// Ethereum events contained in vote extensions
+        EthereumEventVoteExtensions,
```
james-chf, line 80: NB: this `ProtocolTxType` is guarded by the `ferveo-tpke` feature flag. I think we should consider removing that flag from this file altogether (or at least refactoring the module so that more generic protocol transaction types are not guarded by it).

james-chf:
I've updated this PR to just go into the `eth-bridge-integration` branch. I will start building #1178 and #1167 off of this

batconjurer (approved):

james-chf:
pls update wasm

PR #1176, opened 2022-06-27
open_genesis_config should return a human-friendly error

james-chf:
Give a slightly better error message when `anomac` or `anomaw` commands are run, but a chain hasn't been joined. Before we would just get `No such file or directory (os error 2)`, now we should indicate that specifically we can't read a genesis config file.

```
mbp> target/debug/anomac epoch
2022-07-07T16:58:56.136364Z  INFO anoma_apps::cli::context: Chain ID: anoma-internal.000000000000000
Error: 
   0: couldn't read genesis config file from .anoma/anoma-internal.000000000000000.toml
   1: No such file or directory (os error 2)

Location:
   apps/src/lib/config/genesis.rs:601

Backtrace omitted.
Run with RUST_BACKTRACE=1 environment variable to display it.
Run with RUST_BACKTRACE=full to include source snippets.
[1] - 2022-07-07 17:58:56
```

This PR also makes some changes such that we can propagate an error all the way back up to `main()` of `anomac`, `anoman`, `anomaw`, so that it will be easier in the future to propagate errors like this. (Relates to anoma/anoma#1049)

To close anoma/namada#97, we should catch the error and translate it to a friendly message such as one indicating to the user to join a network first, rather than panicking with a backtrace.

### DIFF
--- a/.changelog/unreleased/improvements/1176-genesis-config-error.md
+++ b/.changelog/unreleased/improvements/1176-genesis-config-error.md
@@ -1,0 +1,2 @@
+- Improve the error message that is displayed when anoma binaries are run without
+  having joined a chain ([#1176](https://github.com/anoma/anoma/pull/1176))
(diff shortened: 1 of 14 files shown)


tzemanovic:
moved to https://github.com/anoma/namada/pull/291

PR #1182, opened 2022-06-28
Fix a bug in PoS votes_per_token parameter validation

### LINKED ISSUE #1181: PoS params update validation, wrong condition on votes_per_token
the validation for PoS params

```rust
        // Check that there is no more than 1 vote per token
        if self.votes_per_token >= BasisPoints::new(10_000) {
```

should be `if self.votes_per_token > BasisPoints::new(10_000) {`

tzemanovic:
fixes #1181


tzemanovic:
merged in 8f0e57e78f0749183be921233db01a7519c97b25

PR #1186, opened 2022-06-29
Updates to the `eth-bridge-integration` branch

james-chf:
Relates to #1171. Based on changes to the spec incoming in https://github.com/anoma/specs/pull/104

- [x] `EthereumEvent::hash` method
- [x] Remove enum variant for wrapped native ETH
- [x] Ensure `EthereumEvent` type has a nonce (so that we can always derive a unique hash per `EthereumEvent`)
- [x] Update `MultiSignedEthEvent` to use a better voting power type

### DIFF
--- a/shared/src/types/ethereum_events.rs
+++ b/shared/src/types/ethereum_events.rs
@@ -65,7 +65,13 @@ pub enum EthereumAsset {
 #[derive(
     Clone, PartialEq, Eq, Debug, BorshSerialize, BorshDeserialize, BorshSchema,
 )]
-struct FractionalVotingPower(fraction::Fraction);
+pub struct FractionalVotingPower(u64, u64);
+
+impl From<FractionalVotingPower> for fraction::Fraction {
+    fn from(fvp: FractionalVotingPower) -> Self {
+        fraction::Fraction::new(fvp.0, fvp.1)
+    }
+}
 
 /// This is created by the block proposer based on the Ethereum events included
 /// in the vote extensions of the previous Tendermint round
--- a/shared/src/types/ethereum_events.rs
+++ b/shared/src/types/ethereum_events.rs
@@ -12,7 +12,7 @@ use crate::types::token::Amount;
 #[derive(
     Debug, PartialEq, Eq, Clone, BorshSerialize, BorshDeserialize, BorshSchema,
 )]
-pub struct EthereumEventNonce(u64);
+pub struct Nonce(u64);
 
 /// An Ethereum event to be processed by the Anoma ledger
 #[derive(
@@ -21,7 +21,7 @@ pub struct EthereumEventNonce(u64);
 pub enum EthereumEvent {
     /// Event transferring batches of ether from Ethereum to wrapped ETH on
     /// Anoma
-    TransfersToNamada(Vec<TransferToNamada>, EthereumEventNonce),
+    TransfersToNamada(Vec<TransferToNamada>, Nonce),
 }
 
 impl EthereumEvent {
@@ -89,7 +89,7 @@ mod tests {
 
     #[test]
     fn test_ethereum_event_hash() {
-        let nonce = EthereumEventNonce(123);
+        let nonce = Nonce(123);
         let event = EthereumEvent::TransfersToNamada(vec![], nonce);
 
         let hash = event.hash().unwrap();
--- a/shared/src/types/ethereum_events.rs
+++ b/shared/src/types/ethereum_events.rs
@@ -19,6 +19,7 @@ pub enum EthereumEvent {
 }
 
 impl EthereumEvent {
+    /// SHA256 of the Borsh serialization of the [`EthereumEvent`]
     fn hash(&self) -> Result<[u8; 32], std::io::Error> {
         let bytes = self.try_to_vec()?;
         let mut hasher = Sha256::new();
--- a/shared/src/types/ethereum_events.rs
+++ b/shared/src/types/ethereum_events.rs
@@ -2,6 +2,7 @@
 use std::fmt::Debug;
 
 use borsh::{BorshDeserialize, BorshSchema, BorshSerialize};
+use eyre::{eyre, Result};
 use num_rational::Ratio;
 
 use crate::proto::MultiSigned;
@@ -77,6 +78,23 @@ pub enum EthereumAsset {
 /// fraction that is between zero and one inclusive.
 #[derive(Clone, PartialOrd, Ord, PartialEq, Eq, Debug)]
 pub struct FractionalVotingPower(Ratio<u64>);
+
+impl FractionalVotingPower {
+    /// Create a new FractionalVotingPower. It must be between zero and one
+    /// inclusive.
+    pub fn new(numer: u64, denom: u64) -> Result<Self> {
+        if denom == 0 {
+            return Err(eyre!("denominator can't be zero"));
+        }
+        let ratio: Ratio<u64> = (numer, denom).into();
+        if ratio > 1.into() {
+            return Err(eyre!(
+                "fractional voting power cannot be greater than one"
+            ));
+        }
+        Ok(Self(ratio))
+    }
+}
 
 impl From<&FractionalVotingPower> for (u64, u64) {
     fn from(ratio: &FractionalVotingPower) -> Self {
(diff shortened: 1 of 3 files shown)


shared/src/types/ethereum_events.rs
```diff
@@ -33,20 +52,49 @@ pub struct TransferToNamada {
...
+)]
 pub enum EthereumAsset {
-    /// Native ETH
-    Eth,
     /// An ERC20 token and the address of its contract
     ERC20(EthAddress),
 }
 
+/// A fraction of the total voting power. This should always be a reduced
+/// fraction that is between zero and one inclusive.
```
james-chf, line 64: `between zero and one inclusive` - or should this be exclusive of 0 and 1?

shared/src/types/ethereum_events.rs
```diff
@@ -5,24 +5,43 @@ use borsh::{BorshDeserialize, BorshSchema, BorshSerialize};
...
 /// An Ethereum event to be processed by the Anoma ledger
-#[derive(Debug, Clone, BorshSerialize, BorshDeserialize, BorshSchema)]
+#[derive(
+    Debug, PartialEq, Eq, Clone, BorshSerialize, BorshDeserialize, BorshSchema,
+)]
 pub enum EthereumEvent {
     /// Event transferring batches of ether from Ethereum to wrapped ETH on
     /// Anoma
-    TransfersToNamada(Vec<TransferToNamada>),
+    TransfersToNamada(Vec<TransferToNamada>, EthereumEventNonce),
```
james-chf, line 24: I'm not sure how best to include the nonce

shared/src/types/ethereum_events.rs
```diff
@@ -5,24 +5,43 @@ use borsh::{BorshDeserialize, BorshSchema, BorshSerialize};
...
 use crate::proto::MultiSigned;
 use crate::types::address::Address;
+use crate::types::hash::Hash;
 use crate::types::token::Amount;
 
+/// Each event has a nonce set by the Ethereum smart contract that emitted it.
+#[derive(
+    Debug, PartialEq, Eq, Clone, BorshSerialize, BorshDeserialize, BorshSchema,
+)]
+pub struct EthereumEventNonce(u64);
```
batconjurer, line 15: Do we need such a long type name? Could we just get away with `Nonce`?. The type `ethereum_events::Nonce` should make the type clear enough.

shared/src/types/ethereum_events.rs
batconjurer: I think we should refactor the type something like
```
pub enum RawEvent {
    TransfersToNamada(Vec<TransferToNamada>),
    // all other variants which also have a nonce
    ...
}

pub struct EthereumEvent {
    pub nonce: Nonce,
    pub event: RawEvent,
}
```
Although I don't particularly like the name `RawEvent`.

shared/src/types/ethereum_events.rs
```diff
@@ -33,20 +52,55 @@ pub struct TransferToNamada {
...
     /// An ERC20 token and the address of its contract
     ERC20(EthAddress),
 }
 
+/// A fraction of the total voting power. This should always be a reduced
+/// fraction that is between zero and one inclusive.
+#[derive(
+    Clone, PartialEq, Eq, Debug, BorshSerialize, BorshDeserialize, BorshSchema,
+)]
+pub struct FractionalVotingPower(u64, u64);
```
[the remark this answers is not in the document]
batconjurer: I think we should go back to `FractionalVotingPower(fraction::Fraction)` and write our own borsh serialization.  Something like
```
impl TryFrom<&FractionalVotingPower> for (u64, u64) {
    type Error = std::io::Error;
    fn try_from(power: &FractionalVotingPower) -> Result<Self, Self::Error> {
        (
            self.numer()
                  .ok_or_else(|| std::io::Error::new(
                                               ErrorKind::InvalidData,
                                               "Could not serialize FractionalVotingPower, missing numerator"
                                      ) )?,
              self.denom()
                    .ok_or_else(|| std::io::Error::new(
                                                  ErrorKind::InvalidData,
                                                   "Could not serialize FractionalVotingPower, missing denominator"
                                       ) )?
         )
    }
}

impl BorshSerialize for FractionalVotingPower {
    fn serialize<W: Write>(&self, writer: &mut W) -> std::io::Result<()> {
         let (numer, denom) = (u64, u64)::try_from(&self)?;
         (u64, u64).serialize(&(numer, denom), writer)
    }
}

impl BorshDeserialize for FractionalVotingPower {
    fn deserialize(buf: &mut &[u8]) -> std::io::Result<Self> {
       let (numer, denom): (u64, u64) = BorshDeserialize::deserialize(buf)?;
       Ok(FractionalVotingPower(Fraction::new(numer, denom)))
    }
}

impl BorshSchema for FractionalVotingPower {
    fn add_definitions_recursively(
        definitions: &mut HashMap<Declaration, Definition>,
    ) {
        let fields = borsh::schema::Fields::UnnamedFields(borsh::maybestd::vec![
            u64::declaration(),
           u64::declaration()
        ]);
        let definition = borsh::schema::Definition::Struct { fields };
        Self::add_definition(Self::declaration(), definition, definitions);
    }

    fn declaration() -> borsh::schema::Declaration {
        "FractionalVotingPower".into()
    }
}

**Review by batconjurer — CHANGES_REQUESTED** [review]

File: shared/src/types/ethereum_events.rs (hunk 3 above)
Comment (james-chf) [reply shared/src/types/ethereum_events.rs]: Yeah `Nonce` is better. Especially considering we can explicitly import as `ethereum_events::Nonce` if we need the extra context in the name.

File: shared/src/types/ethereum_events.rs (hunk 2 above)
Comment (james-chf) [reply shared/src/types/ethereum_events.rs]: :+1: I've updated it to be like in your comment. I'm not sure about the name `RawEvent` either but we can always change it later

File: shared/src/types/ethereum_events.rs (hunk 4 above)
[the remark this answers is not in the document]
Comment (james-chf) [reply shared/src/types/ethereum_events.rs]: Ah nice, I changed it back to be `FractionalVotingPower(fraction::Fraction)` and used a solution like yours. Then I was wondering why we need to implement `TryFrom` ie why should a `fraction::Fraction` ever possibly not have a numerator or denominator? It looked like the underlying type was representing also NaN and +/-infinity

```rust
pub enum GenericFraction<T>
where
    T: Clone + Integer,
{
    Rational(Sign, Ratio<T>),
    Infinity(Sign),
    NaN,
}
```

so I changed to using the `num-rational` `Ratio<u64>` type directly that `fraction::Fraction` was using under the hood

shared/src/types/ethereum_events.rs
```diff
@@ -33,20 +66,149 @@ pub struct TransferToNamada {
...
+    }
+}
+
 /// This is created by the block proposer based on the Ethereum events included
 /// in the vote extensions of the previous Tendermint round
 #[derive(Debug, Clone, BorshSerialize, BorshDeserialize, BorshSchema)]
 pub struct MultiSignedEthEvent {
     /// Address and voting power of the signing validators
-    pub signers: Vec<(Address, u64)>,
+    pub signers: Vec<(Address, FractionalVotingPower)>,
```
james-chf, line 159: Re. discussion before, I still think it should be `EstablishedAddress` rather than `Address`, as only the `Address::Established` variant will make sense here (validators should all be established accounts). Although we may only handle these addresses opaquely, we can encode in the type what we expect the type of addresses to be

james-chf:
Thanks for reviewing, have acted on comments

shared/src/types/ethereum_events.rs
```diff
@@ -33,20 +66,149 @@ pub struct TransferToNamada {
...
 
+/// A fraction of the total voting power. This should always be a reduced
+/// fraction that is between zero and one inclusive.
+#[derive(Clone, PartialOrd, Ord, PartialEq, Eq, Debug)]
+pub struct FractionalVotingPower(Ratio<u64>);
+
+impl FractionalVotingPower {
+    /// Create a new FractionalVotingPower. It must be between zero and one
+    /// inclusive.
+    pub fn new(numer: u64, denom: u64) -> Result<Self> {
```
james-chf, line 85: I used `eyre` here instead of `thiserror` I wasn't sure when the caller would need to discriminate between an error when denominator is 0, and when the fraction is > 1, I think we will only care if there is an error or not

PR #1195, opened 2022-07-01
Murisi/masp history

murisi:
This branch implements a feature allowing users to view historical shielded and transparent transactions given the necessary keys. It also allows one to filter transactions by the addresses they pertain to or the tokens that they use. This is done using Tendermint transaction events and queries.

# Invasive Changes
* bit-vector to indicate which block transactions were accepted by VPs 
* creating special RPC query that aggregates transaction acceptance bit-vectors over range
* adding a `get_tx_index` to transaction execution environment to help order shielded transactions
* decrypting Tx objects and conducting inspection so that correct Tendermint events are attached

# Limitations and Considerations
* this implementation of histories is incompatable with DKG as it uses information that will be encrypted in future
* block events were considered instead of tx events, but queries for them will download the entire block using bandwidth
* this implementation of histories can only parse the built-in transaction formats, not arbitrary WASM transactions
* indexers were considered and it is hoped that we will migrate to them once they are completed

# Other Changes
* shielded transactions are now stored in the DB by index rather than as a linked list to simplify client fetching code and reduce space requirements
* shielded transactions are now stored with their epoch, height, and index to facilitate history ordering and correct rewards computations


shared/src/types/storage.rs
```diff
@@ -43,6 +43,125 @@ pub const VP_KEY_PREFIX: char = '?';
...
+    PartialEq,
+    Eq,
+    PartialOrd,
+    Ord,
+    Hash,
+    Debug,
+    Serialize,
+    Deserialize,
+)]
+pub struct BlockResults(pub Vec<u8>);
```
tzemanovic, line 98: let's maybe use https://crates.io/crates/bit-vec here

apps/src/lib/node/ledger/shims/abcipp_shim.rs
```diff
@@ -95,14 +97,20 @@ impl AbcippShim {
                         .map_err(Error::from)
                         .and_then(|res| match res {
                             Response::ProcessProposal(resp) => {
+                                #[cfg(feature = "ABCI")]
```
tzemanovic, line 100:
```suggestion
                                // The transaction's data is used in the response to add
                                // transfer event via `impl From<Tx> for ResponseDeliverTx` conversion.
                                #[cfg(feature = "ABCI")]
```

tzemanovic:
I think the general approach is fine for short-term, but we'll need remain aware that this only works for tx_transfer. There are other types of transactions that may move balance (e.g. PoS tx_bond and tx_withdraw, but for withdrawal, we'd actually need to determine the amount from the storage, not from tx.data, because of slashing that may be applied in the transaction when it's executed)

PR #1200, opened 2022-07-04
multitoken transfer

yito88:
closes anoma/namada#132


shared/src/types/token.rs
```diff
@@ -279,20 +296,58 @@ pub fn is_any_token_balance_key(key: &Key) -> Option<&Address> {
...
+                        segments: key.segments[1..(len - 2)].to_vec(),
+                    };
+                    Some((sub_prefix, owner))
+                }
+                _ => None,
+            }
+        }
         _ => None,
     }
 }
```
james-chf, line 353: Can `is_multitoken_balance_key`/`is_any_multitoken_balance_key` share some code? These would be useful to have tests for as well (asserting what multitoken keys look like)

vm_env/src/token.rs
```diff
@@ -73,38 +78,102 @@ pub mod tx {
...
+            None => token::balance_key(token, src),
+        };
+        let dest_key = match dest_sub_prefix {
+            Some(sub_preifx) => {
+                let prefix =
+                    token::multitoken_balance_prefix(token, &sub_preifx);
+                token::multitoken_balance_key(&prefix, dest)
+            }
+            None => token::balance_key(token, dest),
+        };
```
james-chf, line 100: typo - `sub_preifx` should be `sub_prefix`

james-chf (commented):
I haven't looked at IBC-related changes, wanted to test against `vp_multitoken` from `https://github.com/anoma/anoma-wasm-multitoken` and ABCI++ but was getting an error when doing a regular (non-multitoken) transfer from the faucet which might be to do with my setup, so still need to test this out.

It'd be nice if we could split the PR into `anomac balance` and `anomac transfer` and target a `main` branch for it, but I'm guessing some of the logic is tied in with IBC.

wasm_for_tests/wasm_source/src/lib.rs
```diff
@@ -140,6 +140,8 @@ pub mod main {
             source: _,
             target,
             token,
+            source_sub_prefix: _,
+            target_sub_prefix: _,
             amount,
```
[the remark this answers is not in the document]
yito88: Thanks! That looks better and I've tried that. However, clippy said to use `_`.

tzemanovic:
moved to https://github.com/anoma/namada/pull/297

PR #1217, opened 2022-07-06
Run CI for `eth-bridge-integration` and diff against `master`

james-chf:
PR for running CI of this integration branch against `master`
- cargo fmt/clippy checks
- catch conflicts with `master` early
- see how the diff looks against `master` at any time


james-chf:
Maybe there is a better way to run full CI for every `eth-bridge-integration` merge (eg directly via ci.heliax.dev)

PR #1219, opened 2022-07-06
temporarily switch back to rexpect from expectrl

tzemanovic:


tzemanovic:
there are still flakey e2e tests with rexpect, so closing this to stay with expectrl

PR #1222, opened 2022-07-07
CI: bump sccache bucket

### LINKED ISSUE #1204: sccache timeouts in CI
eg https://ci.heliax.dev/anoma/anoma/4214/3/4

```
+ sccache --start-server
sccache: Starting the server...
sccache: error: Timed out waiting for server startup
```

This only happens occasionally. We could probably workaround by allowing more time for sccache to start up.

tzemanovic:
this seems to fix #1204


tzemanovic:
the existing sscache has been cleared so this is no longer needed

PR #1228, opened 2022-07-08
Add stub OpenAPI spec

james-chf:
Relates to https://github.com/anoma/namada/issues/149

This is a stub OpenAPI spec that could be imported with Insomnia or Postman, with only one type of operation - making a call via [`abci_query`](https://docs.tendermint.com/master/rpc/#/ABCI/abci_query) of Tendermint RPC. The default server is the ledger running on your localhost (`http://127.0.0.1:26657`). Many more examples could be added, e.g. specific `value/<path>` queries for common VPs such as `vp_token` (or alternatively we could have VP-specific OpenAPI files?).

Unfortunately only the first example is imported into tools such as Insomnia or Postman. Maintaining a direct Postman collection of examples as well as/instead of this OpenAPI spec may be better. Unsure where the most appropriate place for this OpenAPI spec file is (maybe `specs/` or `docs/` somewhere?).

### DIFF
--- a/specs/openapi.yml
+++ b/specs/openapi.yml
@@ -1,0 +1,130 @@
+openapi: 3.0.3
+info:
+  title: Anoma
+  description: Interacting with an Anoma blockchain via Tendermint RPC
+  version: 0.6.1
+servers:
+  - url: http://127.0.0.1:26657
+    description: Tendermint RPC endpoint for an Anoma ledger
+paths:
+  /:
+    post:
+      summary: Interact with the Anoma blockchain via Tendermint RPC
+      operationId: abci_query
+      requestBody:
+        required: true
+        content:
+          application/json:
+            schema:
+              type: object
+              properties:
+                id:
+                  description: Should be unique between requests
+                  type: integer
+                  example: 58392
+                method:
+                  description: The Tendermint RPC method being called which in this case should always be abci_query
+                  type: string
+                  enum:
+                    - "abci_query"
+                params:
+                  type: object
+                  required:
+                    - path
+                  properties:
+                    path:
+                      description: Path as will be recognized by the ledger
+                      oneOf:
+                        - type: string
+                          enum:
+                            - "epoch"
+                            - "dry_run_tx"
+                          description: >
+                            * `epoch` - Epoch at the given block height
+                            * `dry_run_tx` - Dry run a transaction
+                        - type: string
+                          description: Read a storage value with exact storage key
+                          pattern: r"^value\/(\w|\/)+$"
+                        - type: string
+                          description: Read a range of storage values with a matching key prefix
+                          pattern: r"^prefix\/(\w|\/)+$"
+                        - type: string
+                          description: Check if the given storage key exists
+                          pattern: r"^has_key\/(\w|\/)+$"
+                    data:
+                      description: Optional data to go along with the query (base64-encoded if necessary)
+                      type: string
+                      example: "abcd"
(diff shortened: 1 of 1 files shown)


specs/openapi.yml
```diff
@@ -0,0 +1,130 @@
...
+                    height:
+                      description: Height as a base64 encoded integer (0 means latest)
+                      type: string
+                      example: "1"
+                      default: "0"
+                    prove:
+                      description: Include proofs of the transaction's inclusion in the block
+                      type: boolean
+                      example: true
+                      default: false
```
james-chf, line 68: This "prove" parameter is in Tendermint docs but not sure if it makes sense for any of our RPC query paths so maybe should be removed?

james-chf: Ah interesting! I added an example for getting an account's public key also using this "prove" parameter.

specs/openapi.yml
```diff
@@ -0,0 +1,199 @@
...
+                        - type: string
+                          enum:
+                            - "epoch"
+                            - "dry_run_tx"
+                          description: >
+                            * `epoch` - Epoch at the given block height
+                            * `dry_run_tx` - Dry run a transaction
+                        - type: string
+                          description: Read a storage value with exact storage key
+                          pattern: r"^value\/([\x00-\x7F]|\/)+$"
```
james-chf, line 47: I'm not sure what pattern to allow here for a valid storage key, `[\x00-\x7F]` covers ASCII bytes (0-127), so this pattern can accept things like `value/#atest1v4ehgw36g4pyg3j9x3qnjd3cxgmyz3fk8qcrys3hxdp5xwfnx3zyxsj9xgunxsfjg5u5xvzyzrrqtn/public_key` with a `#` character in it, but maybe this is too wide (or too narrow?).

specs/openapi.yml
```diff
@@ -0,0 +1,199 @@
...
+                  type: object
+                  required:
+                    - path
+                  properties:
+                    path:
+                      description: Path as will be recognized by the ledger
+                      oneOf:
+                        - type: string
+                          enum:
+                            - "epoch"
```
james-chf, line 40: Docstrings for these values taken from https://github.com/anoma/anoma/blob/9475cb338489b37aa44971ef4d768e2a57b20aa5/apps/src/lib/node/ledger/rpc.rs#L16 , although for "epoch" I think we can query the epoch for any given block height rather than just the latest block, when going via `?abci_query`.

tzemanovic: I've opened https://github.com/anoma/namada/issues/172

specs/openapi.yml
```diff
@@ -0,0 +1,199 @@
...
+                            - "dry_run_tx"
+                          description: >
+                            * `epoch` - Epoch at the given block height
```
tzemanovic, line 43:
```suggestion
                            * `epoch` - Get the epoch of the last block (the height argument is not yet supported <https://github.com/anoma/namada/issues/172>)
```

specs/openapi.yml
tzemanovic: currently we allow utf-8

james-chf: I've changed these regexes to permit any pattern for storage key (I guess it's too permissive as it would allow non-UTF-8 string).

tzemanovic (approved):
LGTM, thanks!

tzemanovic:
moved to https://github.com/anoma/namada/pull/322

PR #1229, opened 2022-07-08
Ethereum bridge VP and e2e test clean up

james-chf:
This is a PR to the mainline branch to clear up some code relating to the old Ethereum bridge design, which involved maintaining a queue of transfers. Also make the Ethereum bridge end-to-end test slightly clearer.

At the same time, trying to stop this test flaking in CI (https://github.com/anoma/namada/issues/161), though we probably shouldn't close the linked issue even after this PR lands for a while until we have not seen the flake. We need better logging to investigate why it is happening (I have seen it a few times).


tests/src/e2e/eth_bridge_tests.rs
```diff
@@ -6,22 +6,23 @@ use crate::e2e::setup::constants::{
...
     format!("#{ETH_BRIDGE_ADDRESS}/{}", path)
 }
 
 #[test]
-fn everything() {
-    const LEDGER_STARTUP_TIMEOUT_SECONDS: u64 = 30;
-    const CLIENT_COMMAND_TIMEOUT_SECONDS: u64 = 30;
+fn unauthorized_tx_cannot_write_storage() {
+    const LEDGER_STARTUP_TIMEOUT_SECONDS: u64 = 60;
+    const CLIENT_COMMAND_TIMEOUT_SECONDS: u64 = 60;
```
james-chf, line 25: I *suspect* the flakes may have been to do with the timeouts being 30s (rather than the 40s we seem to use elsewhere), but clutching at straws a bit. We can try increasing this timeout and monitor for flakes going forward. A flaky error can happen at the first client assertion we make e.g.

```rust
anomac_tx.exp_string("Transaction accepted").unwrap();
```
Message:  called `Result::unwrap()` on an `Err` value: 
   0: Reached a timeout for expect type of command
       Needle: Transaction accepted
```

It's possible the ledger is shutting down for some reason before the client makes its call, or the client is making its call too quickly?

Separately from this PR, I wonder if this timeout value should be some tuneable configurable value, maybe from an environment variable. We use `Some(40)` seconds in most places but perhaps it should be a bit higher everywhere.

File: tests/src/e2e/eth_bridge_tests.rs [hunk 2]
```diff
@@ -61,34 +67,25 @@ fn everything() {
...
-        // TODO: we should check here explicitly with the ledger via a
-        //  Tendermint RPC call that the path `value/#EthBridge/queue`
-        //  is unchanged rather than relying solely  on looking at anomac
-        //  stdout.
-        anomac_tx.exp_string("Transaction is invalid").unwrap();
-        anomac_tx
-            .exp_string(&format!("Rejected: {}", ETH_BRIDGE_ADDRESS))
-            .unwrap();
-        anomac_tx.assert_success();
+    if !cfg!(feature = "ABCI") {
```
james-chf, line 78: We really need to catch the flake with logs of what exactly is going on in the ledger (and the client). A flaky error can happen at the first client assertion we make e.g.

```rust
anomac_tx.exp_string("Transaction accepted").unwrap();
```
```
Message:  called `Result::unwrap()` on an `Err` value: 
   0: Reached a timeout for expect type of command
       Needle: Transaction accepted
```

I think it's possible the ledger is shutting down for some reason before the client makes its call. As we wait for the ledger to have committed at least one block, Tendermint is presumably healthy for at least a bit, so not sure why this is happening (if it is).

james-chf:
pls update wasm

tests/src/e2e/eth_bridge_tests.rs
batconjurer: That's entirely possible. But catching it and logging it propoerly has been our white whale

batconjurer (approved):

tests/src/e2e/eth_bridge_tests.rs
james-chf: after https://github.com/anoma/anoma/pull/1202 merges in we will have full logs for all the `anoma*` commands we run, then we could try making the logs available from CI somehow. Maybe making the logs available for download, or just spewing them out to stdout when there is a failure

tzemanovic:
moved to https://github.com/anoma/namada/pull/323

PR #1231, opened 2022-07-11
Refactor ledger startup code

sug0:
I have broken down the `run_aux` function into several different logical units, which should make the ledger startup code more modular and easier to reason about.

This PR will help with the changes done to the eth bridge, since the latter will require even more startup logic code to manage `geth`. For the same reason, it is in conflict with #1196, so this will need to be handled.


sug0:
CI timed out :\
```
+ sccache --start-server
sccache: Starting the server...
sccache: error: Timed out waiting for server startup
```

apps/src/lib/node/ledger/mod.rs
```diff
@@ -282,94 +374,71 @@ async fn run_aux(config: config::Ledger, wasm_dir: PathBuf) {
...
-    let tendermint_dir = config.tendermint_dir();
-    let ledger_address = config.shell.ledger_address.to_string();
+    RunAuxSetup {
+        vp_wasm_compilation_cache,
+        tx_wasm_compilation_cache,
+        db_block_cache_size_bytes,
+    }
+}
+
+fn start_abci_broadcaster_shell(
```
batconjurer, line 388: Can I get some docstrings here?

batconjurer (changes requested):
Looks great to me. Just a couple of tiny things

apps/src/lib/node/ledger/mod.rs
sug0: done

tzemanovic:
I tried to check the windows compilation, but unfortunately, we have a unix only dependency in the wallet `file-lock`, so it's not working atm. We don't officially support windows yet, so it's fine to go ahead - visually it seems fine to me :)

tzemanovic:
if you could just pls add a changelog (https://github.com/anoma/anoma/blob/master/CONTRIBUTING.md#opening-a-pull-request)

sug0:
done! added a changelog entry

juped:
Moved to anoma/namada#270 anoma/namada#271

PR #1236, opened 2022-07-13
fix header hash

yito88:
Fix the hash (merkle tree's root) of a header

IBC proof verification for consensus state failed due to the wrong hash.


tzemanovic:
moved to https://github.com/anoma/namada/pull/298

PR #1237, opened 2022-07-14
Update to transaction hashing and encryption.

IdaTucker:
main points:
Instead of signing the hash of the whole transaction:
Sign(Hash( (code || data || timestamp))),
one now signs the hash of the hash of the code, concatenated to the data and timestamp:
Sign(Hash( (Hash(code) || data || timestamp))),

This is so that we can pass the code hash to the device instead of the whole code field.

We also encrypt in parts, so that we can work with unencrypted data


batconjurer (changes requested):
Also, make sure to hit this with cargo fmt.

PR #1239, opened 2022-07-15
print tendermint log to stdout only if `ANOMA_TM_STDOUT=true`

tzemanovic:
trying to hide the TM stdout to see if it maybe helps a bit with flakiness in our e2e tests.

Repeatedly running the CI, we got 19 e2e successes and 1 failure.


tzemanovic:
merged via https://github.com/anoma/namada/pull/222

PR #1241, opened 2022-07-18
Uploading wasms for eth-bridge-integration from Namada

james-chf:
This PR is only for uploading wasms to S3 via CI for the `eth-bridge-integration` branch from Namada


james-chf:
pls update wasm

PR #1243, opened 2022-07-21
make: add recipes to build wasm in debug

tzemanovic:
closes anoma/namada#51, closes anoma/namada#52

This PR adds recipe `make debug-wasm-scripts` in the root that can be used to enable `debug_log!` expressions in WASM. In the `wasm/wasm_source`, one can debug build an individual tx or VP with e.g. `make debug_tx_transfer`.

Additionally, the `debug_log!` get printed to stdout in non-WASM build, which is handy for testing.


tzemanovic:
moved to https://github.com/anoma/namada/pull/255

PR #1246, opened 2022-07-22
fix validator raw hash

tzemanovic:
closes anoma/namada#200

depends and based on #1202, depends on #1248 (merged on top of #1202, before this PR's commits)


tzemanovic:
I'm trying to add an e2e test in https://github.com/anoma/anoma/pull/1246/commits/50577c8532edf71193de9c37daf5558583d2799a, but it's not yet working - I run the same validator node twice, but it's not double-signing

tzemanovic:
pls update wasm

tzemanovic:
pls update wasm

PR #1248, opened 2022-07-25
filter out expired evidence in the ledger instead of tendermint

tzemanovic:
closes anoma/namada#154


tzemanovic:
moved to https://github.com/anoma/namada/pull/325

PR #1249, opened 2022-07-26
ledger: fix last_epoch to only change when committing block

tzemanovic:
fixes anoma/namada#218


tzemanovic:
moved to https://github.com/anoma/namada/pull/327

PR #1325, opened 2024-10-09
Evaluate submitted transaction function with order

mariari:
When submitting a transaction function using rm-submit we extract the commitments and nullifers from the computed transaction to display to the user.

The transaction function must be evaluated in the same way as is done in the worker.

This does not affect the transaction passed to the worker, the original submitted transaction continues to be passed.


mariari:
Yes superseded

@paulcadman closing.

PR #1327, opened 2024-10-09
Glossary

mariari:
This adds a `Glossary` module to generate.. a glossary. 

You can define new items in any module by `use Glossary`. 

I took the liberty of adding `Anoma.Glossary`, but maybe others have other opinions about this. 

Some stuff I haven't figured out yet: 

How to deal with spaces in the names? You can have functions with spaces in the name, but I can't manage to link to them properly. Two options: either auto replace spaces with `_`, or figure out how to link back to the functions with spaces in the name.  For example, calling `Anoma.Glossary."some function"()` is valid syntax, but the doc system doesn't pick it up.

### DIFF
--- a/lib/anoma_glossary.ex
+++ b/lib/anoma_glossary.ex
@@ -1,0 +1,7 @@
+defmodule Anoma.Glossary do
+  use Glossary
+
+  define "transaction_candidate" do
+    "the bees knees"
+  end
+end
--- a/lib/anoma_glossary.ex
+++ b/lib/anoma_glossary.ex
@@ -1,7 +1,18 @@
-defmodule Anoma.Glossary do
-  use Glossary
+defmodule Glossary do
+  use GlossaryBuilder
 
-  define "transaction_candidate" do
-    "the bees knees"
+  define transaction(candidate) do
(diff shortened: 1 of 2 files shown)


mariari:
I managed to figure out how to link back to definitions with spaces in the name. It's not pretty, but it works. 

So if you have a definition of `transaction candidate`, you can generate a link using plain html links:

``[transaction candidate](Anoma.Glossary#Anoma.Glossary.html#transaction candidate/0)``

instead of the more useful generated link: 

``[transaction candidate](`Anoma.Glossary.html#transaction candidate/0`)``

Only downside is that the first approach does not result in a compiler warning if the definition does not exist.

apps/anoma_lib/lib/anoma_glossary.ex
```diff
@@ -0,0 +1,18 @@
+defmodule Glossary do
+  use GlossaryBuilder
+
+  define transaction(candidate) do
+    """
```
mariari, line 5: let's keep this as `define transaction_candidate` to avoid the ()'s and make it easier to link

PR #1330, opened 2024-10-09
Remove dialyzer suppression attributes

mariari:
The removed suppression attributes have no effect with OTP 27 / Elixir 1.17.0

Closes https://github.com/anoma/anoma-archive/issues/1267

This needs to be refactored for the defactored codebase


mariari:
@juped we should merge this in `v0.25.0`, since we are enforcing versions greater than Elixir 1.17.0.

PR #1331, opened 2024-10-09
Use the standard rocksdb dependency

mariari:
We fix the current spot in master which addresses the reason why we had this fork to begin with:

aeternity/mnesia_rocksdb#51

Works on my machine at the very least


mariari:
Yeah probably, we should probably bump the rocksdb further potentially

`9cea41cb9a6dbc1cef016e92f7a824d369d0bb28`

https://github.com/aeternity/mnesia_rocksdb/commits/master/

there are three patches since then.

juped:
#1798

Issue #1332, opened 2024-10-09
Introduce Node ID Info for Every Anoma Event


agureev:
customer: @mariari 
performer:
deadline:
estimated:
started:
actual:
completed:
confirmed:
dependencies:
- [ ]
supersedes:
-
---

Currently every Anoma actor broadcasts generic messages and subscribes to generic filters. This interacts with the specified semantics of e.g. Executor, which expects completion messages only when it goes into a receive loop. Instead, it now can listed to completion messages from workers in other nodes, breaking this functionality.

We should make each anoma-specific message have a node-id attached to it, so that every engine can only listen to information from its own node.

PR #1333, opened 2024-10-09
Add Node Id Filter

agureev:
Adds a wrapper around all the Anoma Node events with a field specifying the ID of the node the message came from.

Allows for proper multi-homing.


agureev:
Closes #1332

Issue #1295, opened 2024-09-25
Add Replay Functionality


agureev:
customer: @mariari 
performer:
deadline:
estimated:
started:
actual:
completed:
confirmed:
dependencies:
- [ ]
supersedes:
-
---

- [x]  Add Replay Logic
- [x] Integrate Replay Logic with Node Supervisor

At Node application startup, we should check whether there are any block or tx-level events which have not been properly committed by comparing the Logging Event and Block table persistent information. 

If any discrepancies are found, we relaunch stored transactions and consensus events, if present.

Note that for future-proofing the system we need to expect that replays do not always succeed. E.g. something might raise and error during specific transaction execution of a block which will make every replay fail. In this sense, we should first launch a mock Anoma Node with Event Broker as a separate task, run it to see whether replays succeed there, and then replay (or store core information in order to save computing power and restart with it) for the core system.

PR #1337, opened 2024-10-10
Add Replay Functionality

agureev:
Adds a replay functionality for Anoma Node.

The user-facing API is `restart_with_replay(node_id)` with the following semantics:

The user calls `restart_with_replay(node_id)` manually after noticing that their node has been stopped. We assume that the tables with appropriate names have been started and in-sync with regards to their content in the sense that they come from the same node which was launched successfully.

Given the node info, we sync the Event table with the Blocks table, removing any consensus lists that might have been missed by the Logger and launched a mock node with a mock node id. This launches all pending transactions as recorded by the Events table after sync, alongside calling `execute` on any remaining orders. If the launch has been successful, i.e. nothing errors on Node start and we receive all messages regarding launched consensus, we launch the node with our id and same parameters.

Otherwise, we launch a node with the only non-default starting parameters being round for Mempool, next height for Ordering, and uncommitted height for Storage.


agureev:
Closes #1295

PR #1344, opened 2024-10-11
M1dnight/multi homing

mariari:


apps/anoma_node/lib/node/utility/consensus.ex
```diff
@@ -3,54 +3,75 @@ defmodule Anoma.Node.Utility.Consensus do
...
 
+    table =
+      String.to_atom(
+        "#{Anoma.Node.Logging.Events}_#{:erlang.phash2(args[:node_id])}"
+      )
+
     :mnesia.transaction(fn ->
-      for cons <- :mnesia.read(Anoma.Node.Logging.Events, :consensus) do
+      for cons <- :mnesia.read(table, :consensus) do
+        IO.puts("wait for block")
```
agureev, line 40:
```suggestion
```

agureev (changes requested):

apps/anoma_node/lib/node/utility/consensus.ex
agureev: Don't fix and re-roll everything just yet though. I have an already prepared branch with the fix we can make into a PR on top of everything. Might be easier

m1dnight: Alright, I'll keep it in then.

Issue #1345, opened 2024-10-11
Support random bits generation in anoma stdlib for Resource nonce values


paulcadman:
customer: @paulcadman 
performer:
deadline:
estimated:
started:
actual:
completed:
confirmed:
dependencies:
- [ ]
supersedes:
-
---

Random bit generation was added in:

* https://github.com/anoma/anoma/commit/b33771e50bcbf4b85a5360c09042f3d9bab28ec6

but we now realise that this does not cover the requirements for nonce generation in Anoma application resources.

Application authors are required to populate the [`nonce`](https://github.com/anoma/anoma/blob/7a551df0a1204decd864cd0a60e7be2fa69f968c/hoon/resource-machine.hoon#L25) field for each resource present in the transaction. We are proposing to support this by providing the user with a function to generate random bytes. 

The transaction is constructed in a transaction function called from a proving context by the client. The proving context provides exactly one random seed.

A transaction may contain more than one resource, and the number of resources required may be computed in the transaction function.

* The random number generation facility we provide must be able to generate a different nonce for each resource from a single initial seed. 
* It must be possible to pass the state of the rng through different code paths such that all code paths generate different random values. In haskell you need to use something like [`split`](https://hackage.haskell.org/package/random-1.2.1.2/docs/System-Random.html#v:split) for this. I'm not sure if we require this.

We are proposing use the nock function [`raws`](https://docs.urbit.org/language/hoon/reference/stdlib/3d#rawsog) for the random number generation facility.

PR #1348, opened 2024-10-11
Add the Rest of og Door

agureev:
Adds `rad`, `raws`, `rads` gates to the standard library.


agureev:
@paulcadman 

Please let us know whether that addressed #1345

PR #1350, opened 2024-10-13
Fill in Verify Logic For Resource Neo

mariari:
Each commit of this topic should be considered as a correct abstraction ontop of the Resource Neo.

Thus this topic is "ready", however all the functions may not be ready at any particular commit, and new commits will follow that will fill in the gaps.


mariari:
@agureev One thing I noticed that is wrong with my topic

```hoon
+$  logic-proof
  [resource=resource inputs=[public-inputs private-inputs]]
+$  compliance-proof  %compliance
+$  proof  ?(compliance-proof logic-proof)
```

I don't take in account that a proof can be a `compliance-proof`.

I may need to update this topic to account for those

PR #1352, opened 2024-10-13
Resource Neo, from and to Noun

mariari:
We create from and to Nouns, these will be very important for a few things:

1. Making dialyzer pass on the reosurce-neo branch
2. writing ResourceLogic.verify

The first commit of this topic creates enough to do `2.` but not enough to do `1.`, the second commit in the series of commits will allow `1.`  in #1350 to actually be fulfilled


mariari:
A lot of this logic is manual as I was avoiding the great `to_noun` and `from_noun`, as the choice around it are not liked, here are some nice uses of it that would streamline this work

```elixir
iex(anoma@YU-NO)22> Noun.Nounable.to_noun(%{1 => 2, 2 => 3})
[[1 | 2], [2 | 3] | 0]
iex(anoma@YU-NO)23> Noun.Nounable.Map.from_noun(v(22))
{:ok, %{1 => 2, 2 => 3}}

```

mariari:
@agureev One thing I noticed that is wrong with my topic

```hoon
+$  logic-proof
  [resource=resource inputs=[public-inputs private-inputs]]
+$  compliance-proof  %compliance
+$  proof  ?(compliance-proof logic-proof)
```

I don't take in account that a proof can be a `compliance-proof`.

I may need to update this topic to account for those

PR #1400, opened 2024-10-17
Make IntentPool Handle New Nullifiers

agureev:
- Adds NullifierEvent for the workers
- Subscribes IntentPool to the new nullifiers and removes those intents which use the nullified resources.


apps/anoma_node/lib/node/transaction/intent_pool.ex
```diff
@@ -116,7 +139,11 @@ defmodule Anoma.Node.Transaction.IntentPool do
       {:ok, :already_present, state}
     else
       Logger.debug("new intent added #{inspect(intent)}")
-      EventBroker.event(Event.new_with_body({:intent_added, intent}), Broker)
+
+      EventBroker.event(
+        Node.Event.new_with_body(state.node_id, {:intent_added, intent}),
```
m1dnight, line 144: You could just make this Event since you aliased Node.Event

apps/anoma_node/lib/node/transaction/intent_pool.ex
```diff
@@ -155,4 +182,39 @@ defmodule Anoma.Node.Transaction.IntentPool do
...
   end
+
+  @spec handle_new_nullifiers(t(), MapSet.t(binary())) :: t()
+  defp handle_new_nullifiers(state, nlfs_set) do
+    set_of_txs = state.intents
+
+    new_intents =
+      Enum.reject(set_of_txs, fn x ->
+        Stream.map(x.actions, fn x -> x.proofs end)
+        |> Stream.map(fn x ->
```
m1dnight, line 193: Stream makes no sense here since you will call Enum afterwards which will resolve your entire stream anyway

apps/anoma_node/lib/node/transaction/intent_pool.ex
agureev: ? what do you mean?

agureev: oh, no, we should not do this

let us distinguish between these two separate macros

apps/anoma_node/lib/node/transaction/intent_pool.ex
agureev: we call stream map twice, so that is ok

however I do now see that we can use Stream.reject as well

agureev: We Stream.map twice, so one less list gets created

PR #1404, opened 2024-10-17
Log Verification Errors

agureev:
Adds a logging functionality to the backend while processing
transparent RM. If it returns any meaningful value, it will be logged
on the `debug` level.


mariari:
Gonna start integrating this now.

mariari:
I've now rerolled it and it should be ready

PR #1406, opened 2024-10-17
Add RocksDB Option During Table Initialization

### LINKED ISSUE #1303: renable rocksdb on our tables
customer: @mariari 
performer:
deadline:
estimated:
started:
actual:
completed:
confirmed:
dependencies:
- [ ]
supersedes:
-
---

After the defactor changes we don't use rocksdb for tables.

We should amend this by making them rocksdb backed.

agureev:
Closes #1303


mariari:
The default should be true for rocks.

We should revive the configuration locations so the database gets setup there.

mariari:
Still needs environmental logic.

PR #1394, opened 2024-10-16
compile_protoc: remove old .pb.ex outputs

juped:
This is in case a message type is removed, or something similar. They
will dangle as Elixir modules locally and possibly cause confusion or
issues if they're referenced.

PR #1417, opened 2024-10-18
Update protobuf files only when really changed

mariuszzak:
I noticed that ElixirLS keeps rebuilding the app, causing my laptop to overheat. I looked into it and realized that it's happening because `mix compile` rebuilds the protobuf files, which are then recognized as updated, triggering another rebuild and causing a loop. I’ve prepared a quick fix for this.

With this change the task prints either:

```
Updated protobuf files:
apps/anoma_protobuf/lib/anoma/protobuf/prove.pb.ex
apps/anoma_protobuf/lib/anoma/protobuf/prove/response.pb.ex
apps/anoma_protobuf/lib/anoma/protobuf/intent.pb.ex
```

or

```
Protobuf files are up to date
```

Let me know if this makes sense.

### DIFF
--- a/apps/compile_protoc/lib/mix/tasks/compile/protoc.ex
+++ b/apps/compile_protoc/lib/mix/tasks/compile/protoc.ex
@@ -1,5 +1,15 @@
 defmodule Mix.Tasks.Compile.Protoc do
-  @moduledoc "I compile the protobuf files before compiling the project."
+  @moduledoc """
+  I compile the protobuf files before compiling the project.
+
+  In order to avoid unnecessary rebuilds and prevent tools like Elixir LS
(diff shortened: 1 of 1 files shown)


mariari:
Hello Thank you for your pull request.

We actually have a patch for that, that's currently in next and should work:

https://github.com/anoma/anoma/pull/1389

Along with another patch that that fixes an issue of dangling pb.ex files

https://github.com/anoma/anoma/pull/1394

Would these patches work for you?

mariuszzak:
Ah, sorry, I wasn't aware that there is an open PR solving this.
Your patch is even better since it's much simpler and completely prevents recompilation. 🎉

Thanks for solving this, it works for me!

PR #1419, opened 2024-10-21
Remove sleep from EConsensus.execution_continues/0

mariari:
Sleeping is bad, but even worse, Supversior.stop is synchronous so this achieves nothing


agureev:
This example design was generally bad and was onlya work-around until multi-homing was introduced

Sp this issue is priperly solged in #1410

PR #1455, opened 2024-10-24
Fix for some credo errors

m1dnight:


apps/anoma_node/lib/examples/eintent_pool.ex
```diff
@@ -53,7 +53,7 @@ defmodule Anoma.Node.Examples.EIntentPool do
     IntentPool.remove_intent(enode.node_id, intent)
 
     # the intent will be removed from the mapset.
-    assert Enum.count(IntentPool.intents(enode.node_id)) == 0
+    assert IntentPool.intents(enode.node_id) == []
```
mariari, line 56: What if intents are a set? This to me doesn't seem equivalent?

mariari (changes requested):

PR #1460, opened 2024-10-24
separate config for logging levels in dev

m1dnight:
I added the option to purge the logging from releases below the set level so there's no overhead.


mariari:
Please reroll this a bit so when running tests, the terminal isn't spammed

m1dnight:
@mariari this has been rerolled

PR #1465, opened 2024-10-25
Mariari/style guide refactoring

mariari:


TheMorpheus407:
## Research Summary

**Issue**: Parallelize sync_service.dart cleanup operations with correct error handling

### Current Behavior

Two methods in  execute cleanup operations **sequentially** when they could run in parallel:

1. **** (lines 505-552): Runs  then  sequentially
2. **** (lines 554-598): Same sequential pattern

Each method uses try-catch blocks to collect errors from both operations for DLQ reporting.

### Required Change

Use  to parallelize these two independent operations while preserving error semantics (both errors must be collected if both operations fail).

### Impacted Files

| File | Lines |
|------|-------|
|  | 505-552, 554-598 |
|  | 2390+ (existing tests) |
|  | 140, 467+ (DLQ tests) |

### Key Constraints

1. **Error semantics must be preserved**: Both errors collected when both operations fail
2. **Existing patterns**:  already used in this file at line 420 and in  line 49
3. **Tests must pass**: Existing tests verify sequential error collection behavior

### Test Ideas

1. Both operations succeed - no DLQ entry
2. First fails, second succeeds - DLQ with first error
3. First succeeds, second fails - DLQ with second error  
4. Both fail - DLQ with **both** errors collected
5. Verify concurrent execution via mock delays

PR #1471, opened 2024-10-25
Mariari/artem roots fix

mariari:
@agureev should reroll


agureev:
re-rolled
https://github.com/anoma/anoma/pull/1424

PR #1475, opened 2024-10-28
Aliasing for Development IEx Sessions

agureev:
Modifies the .iex file to run aliasing of core Anoma modules upon detecting the development environment.


m1dnight:
I think this belongs in your own `~/.iex.exs` file, rather than in the repo?

nvm, did not see you closed it.

PR #1476, opened 2024-10-28
Run run_nock/2 instead of prove/2 in EClient

mariuszzak:
I noticed a small discrepancy. It seems the intention was to test the `run_nock/2` function. Is it correct?

### DIFF
--- a/apps/anoma_client/lib/examples/e_client.ex
+++ b/apps/anoma_client/lib/examples/e_client.ex
@@ -252,8 +252,7 @@ defmodule Anoma.Client.Examples.EClient do
 
     assert {:ok,
             %RunNock.Response{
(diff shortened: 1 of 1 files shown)


m1dnight:
Hi @mariuszzak  thanks for the PR! 

A bit annoying, but I saw the bug as well and it has been fixed in a refactor in the meanwhile so this merge is no longer necessary. 

Just wanted to say we do appreciate your PRs, so thanks so much for those!

mariuszzak:
Yeah, there is indeed another PR touching this part, so I’ll go ahead and close this one.

Thanks a lot 🙂 I would love to work on something more useful, so if you have any suggestions, please let me know. I'm also available on Discord.

Issue #1495, opened 2024-10-30
With Subscription Macro


mariari:
customer: @mariari 
performer:
deadline:
estimated:
started:
actual:
completed:
confirmed:
dependencies:
- [ ]
supersedes:
-
---

Currently we have some nice functions like `EventBrorker.subscribe_me/1` and `EventBroker.unsubscribe_me/1`

These functions are great, however, for testing, we often want the ability to say `EventBroker.with_subscription/2`

#### BNF
```ebnf
with_subscription ::= with_subscription <matched_expression> <do_expr>
```
#### Abstract Notation

```elixir
with_subscription [filter1...filtern] do
   ...
end
```
#### Usage

```elixir
  def resubmit_trivial_swap_again(node_id \\ Node.example_random_id()) do
    EventBroker.with_subscription [] do
      submit_successful_trivial_swap(node_id)
      code = trivial_transparent_transaction()

      Mempool.tx(node_id, code, "id 2")
      Mempool.execute(node_id, Mempool.tx_dump(node_id))

      recieve_logger_failure(node_id, "nullifier already")
    end
  end
```

### Macro Semantics
If we call the macro as the following

```elixir
subscription_list = ...
EventBroker.with_subscription subscription_list do ... end
```
Then it should roughly do the following 

`EventBroker.subscribe_me(subscription_list)` 
do the work inside the `do` block, then call 
`EventBroker.unsubscribe_me(subscription_list)`

### Extra Functionality

What would be nice is to capture the previous subscription, and resubscribe the pid to it.

I.E. if we in our agent had:

```elixir
EventBroker.subscribe_me([filter1, filter2])
EventBroker.subscribe_me([filter1])

EventBroker.with_subscription [filter1] do
  ...
end
```
I'd expect we'd still be subscribed to the output of:

```elixir
EventBroker.subscribe_me([filter1, filter2])
EventBroker.subscribe_me([filter1])
```

as the `with_subscription` should restore the state of the subscriptions to before it was called.

This can be done in a lot of ways, but finding a way to restore subscriptions would be great.

If we can have multiple kinds of subscriptions at once, I suspect this can be done by simply checking if the subscription already exists and using that, if not then a query may need to happen.

I am unfamiliar with the specifics of the event-broker subscription system, so these are merely ideas.

PR #1498, opened 2024-10-30
Create a with_subscription Macro 

agureev:
Make a macro spawning a new worker and asynchronously subscribing to
specified filters, executing the logic, unsubscribing, and then
removing the result.

Changes Logging examples to use the macro.


agureev:
Closes #1495

Issue #1490, opened 2024-10-29
Macro For Defining Events and Filters for Them Automatically in a Node


agureev:
customer: @agureev 
performer:
deadline:
estimated:
started:
actual:
completed:
confirmed:
dependencies:
- [ ]
supersedes:
-
---

When we define Event structs for our Node, we frequently define Filters for trivially checking whether that exact events get received or not.

Example: 

```elixir
typedstruct module: TxEvent do
    field(:id, binary())
    field(:tx, Mempool.Tx.t())
end
  
deffilter TxFilter do
    %EventBroker.Event{body: %Node.Event{body: %Mempool.TxEvent{}}} ->
      true

    _ ->
      false
end
```

Even if this is not used for core functionality, this is frequently important for proper example environment.

We might consider getting a macro over `typedsctuct` using `deffilter` (`eventstruct` or something?) which both creates the typedstruct and defines its trivial filter.

PR #1501, opened 2024-10-30
#1490 - Implement defevent macro

mariuszzak:
This PR is aiming to address https://github.com/anoma/anoma/issues/1490 issue. 

It implements `defevent` macro.

With this change we could for example refactor  `Anoma.Node.Transaction.Backends` module this way:

```
  typedstruct enforce: true, module: CompleteEvent do
    field(:tx_id, integer())
    field(:tx_result, {:ok, any()} | :error)
  end
  
  ...
  
    deffilter CompleteFilter do
    %EventBroker.Event{body: %Node.Event{body: %CompleteEvent{}}} ->
      true

    _ ->
      false
  end
  
```

=>

```
  use Node.Event.DefEvent

  defevent CompleteEvent, filter: CompleteFilter do
    field(:tx_id, integer())
    field(:tx_result, {:ok, any()} | :error)
  end
```

Please let me know if this meets your expectations. I'm happy to adjust or expand it if needed. I’ve tried to follow all the conventions I’m aware of, but I’m not certain everything is correct.


agureev:
This looks very nice, thank you!

One thing I haven't thought about before filing an issue: having docs or typed structs is nice. Can you add an extra argument to the macro (with default `nil` e.g.) which then puts a `@typedoc` in the struct?

mariuszzak:
The entire body of the `do ... end block` is passed into `typedstruct`, which means you can specify `@typespec` in the same way as with the `typedstruct` macro. For example:

```elixir
defevent CompleteEvent, filter: CompleteFilter do
  @typedoc """
  Foo bar
  """

  field(:tx_id, integer())
  field(:tx_result, {:ok, any()} | :error)
end
```

However, there is currently no way to specify `@typedoc` for the Filter module. If it is needed, do you have any ideas on what kind of syntax would make the most sense?

agureev:
> The entire body of the `do ... end` block is passed into `typedstruct`, which means you can specify `@typespec` in the same way as with the `typedstruct` macro. 

Oh, I did not notice that, thanks for pointing it out! That's all that I needed

agureev:
@mariari asked whether you could by chance also provide an example not only of the macro but of the macro usage? That is, an example of:
- making some arbitrary Event with filter using the macro, then
-  subscribing to it in the EventBroker, then 
-  sending the event of a specified format to the EventBroker
-  send a message of a different format
-  check whether the message has been succesfully received and the other has not

You can use my `with_subscription` macro if you'd want to: https://github.com/anoma/anoma/pull/1498

mariuszzak:
Sure, I'm on it 👍

mariuszzak:
Added the the extra example. Does it make sense?

I didn't use `with_subscription` since it hasn't been merged yet. I could rebase this branch on top of `artem/with-sub` or refactor it later on. Which approach would you prefer?

mariari:
> Added the the extra example. Does it make sense?
> 
> I didn't use `with_subscription` since it hasn't been merged yet. I could rebase this branch on top of `artem/with-sub` or refactor it later on. Which approach would you prefer?

If you want to use it, you can just merge the topic (branch) into yours. It's fine either way though no more changes needed I don't think

mariuszzak:
Ok, let's keep it as it is then for now. 👍

mariari:
Rerolling to `48443e1d69da89e562f7595bda4200b7b92a3589` to fix some credo errors and make the history a bit cleaner.

PR #1513, opened 2024-11-01
Update promise graph task template

cwgoes:
This will fix the template so it no longer looks super weird on new issues.

### DIFF
--- a/.github/ISSUE_TEMPLATE/1-tasks.md
+++ b/.github/ISSUE_TEMPLATE/1-tasks.md
@@ -13,10 +13,7 @@ started:
 actual:
 completed:
 confirmed:
-dependencies:
-- [ ]
-supersedes:
--
+
 ---
 
 <definition of done>


.github/ISSUE_TEMPLATE/1-tasks.md
```diff
@@ -13,10 +13,7 @@ started:
 actual:
 completed:
 confirmed:
-dependencies:
-- [ ]
-supersedes:
--
+
```
mariari, line 16: Would users have to remember `dependencies:` is a valid field?

cwgoes: Yes, but I think that's easier than remembering to delete it every time if there are no dependencies.

mariari: Probably better to just delete supersedes then, as dependencies are common. Which would fix the rendering issues as it mainly affects the section after rendering due to the empty text

mariari:
Closing due to no updates

PR #1515, opened 2024-11-01
Disable Logging During Tests

agureev:
Previously the Logging test environment was set to `:error` which
caused unnecessary messages to pop up on user's screen after running
tests. These were unneeded since the test were running exactly to test
for errors occuring.

Generally, logging is availiable to us interactively through examples,
so it is better to keep the terminal clear when running tests instead.

Hence this commit turns of logging at the test level.


mariari:
Please reroll we should do this properly rather than disable it.

PR #1518, opened 2024-11-02
Mariari/example meaningful return

mariari:
Based on V0.26.0-rc1 needs to be rebased on the next release


mariari:
This was completed in v0.28.0 under a different set of commits

PR #1533, opened 2024-11-05
Fix copy-paste error in Nock example comment

rokopt:
A trivial fix for what appears to be a copy/paste error in a header comment to a Nock example.


agureev:
Per the contributor's guide, such individual bug-fix commits are to be based on the commit introducing the bug.

mariari:
In the future please follow what @agureev mentioned. However since this is a documentation bug I'll let it slip this time.

PR #1537, opened 2024-11-05
Add examples for proving/running Nock programs without any arguments

m1dnight:


mariari:
@m1dnight this relies on the testnet which I believe is incorrect given what these commits are

Issue #1534, opened 2024-11-05
Intermittent failure in `anoma_node` tests `Examples.ETransaction.bluf_transaction_errors` and `Examples.ETransaction.bluff_txs_write_nothing`


rokopt:
While testing something else, I encountered a `mix test` failure in `Examples.ETransaction.bluff_txs_write_nothing`.  Unfortunately it was intermittent; using the same seed on a rerun didn't reproduce the failure.  I even tried forcing the random `node_id` to the same one that had produced the original failure, and the test passed.

However, I was eventually able to reproduce it just by running it in a loop without forcing any seeds or node IDs, against the current stock `base`.  In my most recent attempt, for example, it failed on the 14th iteration.  That failure was in the same group of tests but a different specific one -- `bluf_transaction_errors` [sic] instead of `bluff_txs_write_nothing`.

```
==> anoma_node
Running ExUnit with seed: 145026, max_cases: 48

.............

  1) test example Anoma.Node.Examples.ETransaction.bluf_transaction_errors/0 (TransactionTest)
     apps/anoma_node/test/transaction_test.exs:3
     ** (MatchError) no match of right hand side value: []
     code: use TestHelper.GenerateExampleTests, for: Anoma.Node.Examples.ETransaction
     stacktrace:
       (anoma_node 0.25.0) lib/examples/e_transaction.ex:424: Anoma.Node.Examples.ETransaction.bluf_transaction_errors/1       test/transaction_test.exs:3: (test)

...............................................
Finished in 2.0 seconds (0.1s async, 1.9s sync)
61 tests, 1 failure
```

```
==> anoma_node
Running ExUnit with seed: 346920, max_cases: 48

................
05:07:28.208 [error] node=example_7E5B94037E2DEAEC8A9B43B3FE97A8E8


  1) test example Anoma.Node.Examples.ETransaction.bluff_txs_write_nothing/0 (TransactionTest)
     apps/anoma_node/test/transaction_test.exs:3
     ** (MatchError) no match of right hand side value: []
     code: use TestHelper.GenerateExampleTests, for: Anoma.Node.Examples.ETransaction
     stacktrace:
       (anoma_node 0.25.0) lib/examples/e_transaction.ex:425: Anoma.Node.Examples.ETransaction.bluf_transaction_errors/1
       (anoma_node 0.25.0) lib/examples/e_transaction.ex:451: Anoma.Node.Examples.ETransaction.bluff_txs_write_nothing/1
[...]

PR #1539, opened 2024-11-05
Replace transaction examples' dirty reads with transactional reads

rokopt:
@agureev worked out that the cause of #1534 was probably that the transaction example was doing a dirty read rather than a transactional read.  He provided 58312434bebbd6de01525969bab5d848781b6516 to illustrate how to replace the dirty read with a transactional read.

Several other examples have the same problem, but his system doesn't reproduce it so he asked me to try fixing the others and testing the fixes.

In this PR I try to apply the same style of change to all the `dirty_read` calls in the transaction examples.  So far my tests have passed a couple hundred iterations without failures, which is more than they ever did before the changes.  I'll continue letting them run for a while, but I think they're probably worth a review by now.


agureev (approved):
Looks good to me

Issue #1547, opened 2024-11-06
Wrong format for Storage in the Backends


mariari:
customer:
performer:
deadline:
estimated:
started:
actual:
completed:
confirmed:
dependencies:
- [ ]
supersedes:
-
---
In `Anoma.Node.Transaction.Backends`

We have the following code:

```elixir
      Ordering.add(
        node_id,
        {id,
         %{
           append: [
             {:nullifiers, map.nullifiers},
             {:commitments, map.commitments}
           ],
           write: [{:anchor, anchor}, {:ct, ct_new}]
         }}
      )
```

These names `:nullifiers`, `:commitments` etc etc should be a string path something akin to `["anoma", "nullifiers", ...]` rather than an atom.

PR #1553, opened 2024-11-06
Storage to Use Keyspaces

agureev:
Before we stored commitments/nullifiers etc by their bare
atoms. Instead, we need to provide them as existing in the anoma keyspace.

Generally, make Storage types accept only keyspaces.


agureev:
Closes #1547

PR #1562, opened 2024-11-08
cairo backend

XuyangSong:


apps/anoma_node/lib/node/transaction/backends.ex
```diff
@@ -381,4 +392,89 @@ defmodule Anoma.Node.Transaction.Backends do
...
+  @spec cairo_resource_tx(String.t(), binary(), Noun.t()) ::
+          :ok | :error
+  defp cairo_resource_tx(node_id, id, result) do
+    with {:ok, tx} <- CTransaction.from_noun(result),
+         true <- Anoma.RM.Transaction.verify(tx),
+         true <- root_existence_check(tx, node_id, id),
+         # No need to check the commitment existence
+         true <- nullifier_existence_check(tx, node_id, id) do
+      {ct, append_roots} =
+        case Ordering.read(node_id, {id, :ct}) do
```
agureev, line 405: Are we using the same commitment tree/nullifier and commitment set as the transparent RM?

apps/anoma_node/lib/node/transaction/backends.ex
```diff
@@ -381,4 +392,89 @@ defmodule Anoma.Node.Transaction.Backends do
...
+      {ct_new, anchor} =
+        CommitmentTree.add(ct, tx.commitments)
+
+      Ordering.add(
+        node_id,
+        {id,
+         %{
+           append: [
+             {:nullifiers, MapSet.new(tx.nullifiers)},
+             {:commitments, MapSet.new(tx.commitments)},
```
agureev, line 423: Do you do any commitment checks? If not, why do you need to add to the commitment set here? Asking this because one way or another I don't think we should have this set around and should avoid it whenever possible.

agureev (commented):
Due to the fact that currently things are hacky without a proper commitment tree, I am unsure of some of the storage choices here and wanted to asked some questions.

@XuyangSong thank you beforehand for answering!

apps/anoma_node/lib/node/transaction/backends.ex
XuyangSong: ah, you're right. We don't need the commitment set for shielded resources. I'll remove it.

apps/anoma_node/lib/node/transaction/backends.ex
XuyangSong: Assuming they go to different commitment trees and nullifier sets for now, things may change after interoperability discussion.
Would using the `id` here result in different storage? Or should I use a separate indicator like `:cairo_nullifiers` to distinguish the nullifier sets?

agureev: So yes, you would have to use different keys for things of different sorts. If you want to keep transparent and shielded resources separate, use `:your_name_for_commitment_tree` as the key at which you read and write/append.

`id` here is just the id of the transaction, it does not result in different storage, every transaction fed into a node X will use one storage that node X provides, yet different resource machines can use different keys.

agureev: Actually, here is a good idea: write a test where the same node processes a cairo and transparent transaction. Do all the checks pass?

agureev, line 405:
```suggestion
        case Ordering.read(node_id, {id, :cairo_ct}) do
```

apps/anoma_node/lib/node/transaction/backends.ex
```diff
@@ -381,4 +392,89 @@ defmodule Anoma.Node.Transaction.Backends do
...
+
+      {ct_new, anchor} =
+        CommitmentTree.add(ct, tx.commitments)
+
+      Ordering.add(
+        node_id,
+        {id,
+         %{
+           append: [
+             {:nullifiers, MapSet.new(tx.nullifiers)},
```
agureev, line 422:
```suggestion
             {:cairo_nullifiers, MapSet.new(tx.nullifiers)},
```

apps/anoma_node/lib/node/transaction/backends.ex
```diff
@@ -381,4 +392,89 @@ defmodule Anoma.Node.Transaction.Backends do
...
+             {:commitments, MapSet.new(tx.commitments)},
+             {:roots, MapSet.put(append_roots, anchor)}
+           ],
+           write: [{:anchor, anchor}, {:ct, ct_new}]
```
agureev, line 426:
```suggestion
           write: [{:cairo_anchor, anchor}, {:cairo_ct, ct_new}]
```

apps/anoma_node/lib/examples/e_shielded_transaction.ex
```diff
@@ -0,0 +1,61 @@
...
+    {tree, anchor} =
+      Examples.ECommitmentTree.memory_backed_ct_with_trivial_cairo_tx()
+
+    {_ct, _merkle_proof, old_root} = Examples.ECommitmentTree.a_merkle_proof()
+
+    assert {:ok, MapSet.new([old_root, anchor])} ==
+             Storage.read(node_id, {1, :roots})
+
+    assert {:ok, tree} == Storage.read(node_id, {1, :ct})
+    assert {:ok, anchor} == Storage.read(node_id, {1, :anchor})
```
agureev, line 45:
```suggestion
    assert {:ok, tree} == Storage.read(node_id, {1, :cairo_ct})
    assert {:ok, anchor} == Storage.read(node_id, {1, :cairo_anchor})
```

apps/anoma_node/lib/examples/e_shielded_transaction.ex
```diff
@@ -0,0 +1,61 @@
...
+
+    output_commitment =
+      ESResource.a_fixed_output_resource()
+      |> Anoma.CairoResource.Resource.commitment()
+
+    assert {:ok, MapSet.new([input_nullifier])} ==
+             Storage.read(node_id, {1, :nullifiers})
+
+    assert {:ok, MapSet.new([output_commitment])} ==
+             Storage.read(node_id, {1, :commitments})
```
agureev, line 34:
```suggestion
             Storage.read(node_id, {1, :cairo_ommitments})
```

apps/anoma_node/lib/examples/e_shielded_transaction.ex
```diff
@@ -0,0 +1,61 @@
...
+
+    # Generate the nf and cm from fixed resources
+    input_nullifier = ESResource.a_resource_nullifier()
+
+    output_commitment =
+      ESResource.a_fixed_output_resource()
+      |> Anoma.CairoResource.Resource.commitment()
+
+    assert {:ok, MapSet.new([input_nullifier])} ==
+             Storage.read(node_id, {1, :nullifiers})
```
agureev, line 31:
```suggestion
             Storage.read(node_id, {1, :cairo_nullifiers})
```

apps/anoma_node/lib/node/transaction/backends.ex
```diff
@@ -381,4 +392,89 @@ defmodule Anoma.Node.Transaction.Backends do
...
+
+        :error
+    end
+  end
+
+  @spec nullifier_existence_check(CTransaction.t(), String.t(), binary()) ::
+          true | {:error, String.t()}
+  def nullifier_existence_check(transaction, node_id, id) do
+    with {:ok, stored_nullifiers} <-
+           Ordering.read(node_id, {id, :nullifiers}) do
```
agureev, line 453:
```suggestion
           Ordering.read(node_id, {id, :cairo_nullifiers}) do
```

apps/anoma_node/lib/node/transaction/backends.ex
agureev: also: why are you storing an achor here? we store it in the TRM to check that commitments were included at a specific timestamp. you seem to be storing roots separately and not using the anchor directly anywhere

agureev (changes requested):
Given previous discussion, it was revealed that the Cairo RM should currently not use the same nullifier set and merkle tree.

Due to that, I made some suggestions below with name changes, showcasing how to change the keys at which information gets stored.

Hopefully this helps @XuyangSong 

NB: The `:roots` key

apps/anoma_node/lib/node/transaction/backends.ex
XuyangSong: removed the anchor

XuyangSong:
> Given previous discussion, it was revealed that the Cairo RM should currently not use the same nullifier set and merkle tree.
> 
> Due to that, I made some suggestions below with name changes, showcasing how to change the keys at which information gets stored.
> 
> Hopefully this helps @XuyangSong
> 
> NB: The `:roots` key is unchanged since the TRM does not use that key. We store only the anchor. I did add a comment regarding that: you seem to not be using the anchor anywhere, i.e. you never read it in the logic it seems.

Thanks for reviewing. Using `:cairo_nullifiers`, `:cairo_ct` and `:cairo_roots` instead and removed `:anchor` and `:commitments` in cairo storage.

mariari (changes requested):
One more thing thanks to @agureev's topic

#1553 
https://github.com/anoma/anoma/pull/1553/commits/81ac7d06a3bf35d7ba44b45a8bb18ed79b87c817

Can we do this for this topic as well?

Namely using strings for the name within the anoma keyspace?

XuyangSong:
> One more thing thanks to @agureev's topic
> 
> #1553 [81ac7d0](https://github.com/anoma/anoma/commit/81ac7d06a3bf35d7ba44b45a8bb18ed79b87c817)
> 
> Can we do this for this topic as well?
> 
> Namely using strings for the name within the anoma keyspace?

Applied. 
Do we consider adding a prefix to transparent keys as well? @agureev

agureev:
> Applied. Do we consider adding a prefix to transparent keys as well? @agureev

I actually am not sure... Sounds like a good idea to me,  we can do even something better like `["anoma", "transparent", key]` for transparent and `["anoma", "cairo", key]` so that the transparent and other RM keys occupy actually different keyspaces rather than just keys.

However, we should consult @juped first. These keyspace formations will be important once we start doing sharding and he is the one with the best understanding of how that process will occur at this point.

XuyangSong:
fixed the conflicts and rebased

agureev (approved):
The backends logic seems fine to me!

Thank you for rebasing!

PR #1585, opened 2024-11-14
Remove broker from anoma_lib dependencies

karbyshev:


apps/anoma_lib/mix.exs
```diff
@@ -30,7 +30,6 @@ defmodule AnomaLib.MixProject do
   defp deps do
     [
       {:anoma_protobuf, in_umbrella: true},
```
agureev, line 32:
```suggestion
      {:anoma_protobuf, in_umbrella: true}
```

agureev (changes requested):

apps/anoma_lib/mix.exs
karbyshev: Thanks!

Issue #1578, opened 2024-11-13
Hash to Index Mapping in Storage


agureev:
customer: @mariari @XuyangSong 
performer:
deadline:
estimated:
started:
actual:
completed:
confirmed:
dependencies:
- [ ]
supersedes:
-
---

Our commitment tree using the poseidon hash needs the index of a leaf in order to provide a merkle path for it to prove the existence of said hash inside a tree. The user does not control at which time their transaction is processed, so they do not know at which index the commitment they need to check are initially. Hence we need to store the information about the index in Node storage.

In particular, we should store a map whose keys are commitments and values are their indices in a selected merkle tree.

PR #1587, opened 2024-11-14
Store Commitment Index for Cairo Transactions

agureev:
Store the commitments with their indices in storage for proof
generation and verification of the merkle tree paths.


agureev:
Closes #1578

agureev:
@mariari @XuyangSong

Let me know if this seems like an appropriate implementation.

Issue #1581, opened 2024-11-14
Refactor IntentPoolTest to use TestHelper.GenerateExampleTests macro


mariuszzak:
customer:
performer:
deadline:
estimated:
started:
actual:
completed:
confirmed:
dependencies:
- [ ]
supersedes:
-
---

Currently, the `IntentPoolTest` does not use the `TestHelper.GenerateExampleTests` macro because custom arguments are passed to each example. We can consider ways to simplify this.

PR #1592, opened 2024-11-15
Node Tests Use Examples Macro

agureev:
Previously Intent Pool and Solver tests did not use Examples macro generating tests automatically.

This gets rid of tests written by hand.


agureev:
Closes #1581

Issue #1582, opened 2024-11-14
Change Commitment Accumilators to Sets in TRM


agureev:
customer: @mariari 
performer:
deadline:
estimated:
started:
actual:
completed:
confirmed:
dependencies:
- [ ]
supersedes:
-
---

Make the Transparent Resource Machine use simple sets instead of commitments trees as accumilators.

PR #1597, opened 2024-11-18
Make Transparent Resource Machine use Sets as Commitment Accumulators 

agureev:
1. Gets rid of commitment trees in the TRM candidate processing, instead only needing to store the set and the hash of that set (for optimization purposes)

2. Make explicit the commitment accumulator API


agureev:
Closes #1582

mariari:
I find the placement of the new API odd, as it shouldn't be in backend, where is up for debate

PR #1616, opened 2024-11-22
dialyzer: enable -Wunmatched_returns

juped:
This warning checks for unmatched returns, i.e., for exhaustiveness on
structured return values (but not unions of just atoms or integers).


AJAXZAI:
🚀

Issue #1620, opened 2024-11-25
Make Solver Interact with Mempool


agureev:
customer: @mariari 
performer:
deadline:
estimated:
started:
actual:
completed:
confirmed:
dependencies:
- [ ]
supersedes:
-
---

Currently the solver just returns the solved transactions. Instead, it should wrap transactions into a transaction candidate core and submit it to the mempool once solved.

PR #1626, opened 2024-11-25
Make Solver Communicate with Mempool

agureev:
Given a solved set of intents, make solver have submit API. Given
appropriate intents structure, it is to be used to forward the intent
to the Mempool in an appropriate wrap. Currently this only supports
transparent transactions as intents.

Reformats the solver appropriately to suit this functionality.


agureev:
Closes #1620

PR #1630, opened 2024-11-26
Fix the type of Transaction.rseed

m1dnight:
Removes some nasty dialyzer errors


mariari (approved):

mariari:
I recommend in the future basing this on where the bug occurs instead of base

Issue #1568, opened 2024-11-11
Intent Pool Initially Filters Intents


agureev:
customer: @mariari 
performer:
deadline:
estimated:
started:
actual:
completed:
confirmed:
dependencies:
- [ ]
supersedes:
-
---

When an intent gets submitted, it should be initially filtered, i.e. if the intent is to commit/nullify anything already committed/nullified, the pool should reject such an intent.

The most evident way of doing this is storing the most recent nullifier set and commitment tree in the pool's state and checking against it using the same logic as when handling new nullifiers/intents.

PR #1633, opened 2024-11-26
Intent Pool Handles Commitments

agureev:
Intent Pool now stores commitments and does not accepts intents whose commitments are already in, also removing those which have commitments which were transmitted through the Event Broker.


agureev:
Closes #1568

PR #1648, opened 2024-11-28
GRPC should return intents as jammed nouns

m1dnight:


agureev (changes requested):
This should be based on base

apps/anoma_client/lib/api/mempool.ex
```diff
@@ -12,7 +12,6 @@ defmodule Anoma.Client.Api.Servers.Mempool do
   def add(request, _stream) do
     Logger.debug("GRPC #{inspect(__ENV__.function)}: #{inspect(request)}")
 
-    IO.inspect(request, label: "client")
```
mariari, line 15: Included in #1642

m1dnight: removed

m1dnight:
rebased

mariari (approved):

PR #1650, opened 2024-11-28
Add Delta Jets

agureev:
Jets all stubbed resource machine functionality in the standard library.


apps/anoma_lib/lib/anoma/transparent_resource/delta.ex
```diff
@@ -69,4 +69,12 @@ defmodule Anoma.TransparentResource.Delta do
...
     end
   end
+
+  @spec to_noun(t()) :: Noun.t()
+  def to_noun(delta) do
+    delta
+    |> Enum.map(fn {binary, integer} ->
+      [binary | Noun.encode_signed(integer)]
+    end)
+  end
```
mariari, line 79: I see no commit that uses this in `Anoma.TransparentResource.Transaction`, which it should be if we include this.

Issue #1567, opened 2024-11-11
Make Intent Pool Handle New Commitments


agureev:
customer: @mariari 
performer:
deadline:
estimated:
started:
actual:
completed:
confirmed:
dependencies:
- [x] #1566 
supersedes:
-
---

Currently the intent pool handles only new nullifiers, i.e. removes those transactions that use nullifiers already stored. However, any transactions which want to commit a resource which will be committed at some point should also be removed at the time that the resource has ineeed been committed.

Hence, `handle_new_nullifiers` should actually handle also new commitments.

This relies on commitment tree API.

PR #1655, opened 2024-11-29
A General-Purpose Sparse Merkle Tree

giddie:
Firstly, greetings from [Eiger](https://www.eiger.co/)! :wave: 

Secondly, I apologise if this is duplicating work that has already been done internally. This is intended to help with https://github.com/anoma/anoma/issues/1567. I'm aware of [Ray's work on this](https://github.com/anoma/anoma/tree/ray/sparse-merkle-tree), but as that branch is still WIP I'm afraid I found it easier to begin a fresh implementation.

I'm also aware that the tests should be moved into an examples module. I'll begin work on that in a moment. I just wanted to push something out to see if this is heading in the right direction for your needs.

I _am_ able to spend more time on this, so please let me know if you have any particular requirements, and I'm sure I can accommodate.


apps/anoma_lib/lib/sparse_merkle_tree.ex
```diff
@@ -0,0 +1,417 @@
...
+  end
+
+  # Refreshes the cached hashes in the given subtree.
+  @spec rehash_subtree(tree()) :: tree()
+  defp rehash_subtree({:no_hash, left, right}) do
+    new_left = rehash_subtree(left)
+    new_right = rehash_subtree(right)
+
+    {
+      calculate_subtree_hash({:no_hash, left, right}),
```
murisi, line 341: In the case where `left` and `right` do not have cached hashes, it seems like `calculate_subtree_hash` will recomputed them. No? Wouldn't it be more efficient to instead call `calculate_subtree_hash({:no_hash, new_left, new_right})` here?

murisi (commented):
Looks good to me, thanks! Just not sure about the efficiency of `rehash_subtree`...

apps/anoma_lib/lib/sparse_merkle_tree.ex
giddie: I think you're right; good catch :+1:

PR #1659, opened 2024-12-01
GRPC verify and compose endpoint

m1dnight:
This adds the verify and compose endpoints to the client grpc endpoint

### DIFF
--- a/apps/anoma_client/lib/api/intents.ex
+++ b/apps/anoma_client/lib/api/intents.ex
@@ -3,6 +3,10 @@ defmodule Anoma.Client.Api.Servers.Intents do
   alias Anoma.Protobuf.Intents.Add
   alias Anoma.Protobuf.Intents.List
   alias GRPC.Server.Stream
+  alias Anoma.Protobuf.Intents.Compose
+  alias Anoma.Protobuf.Intents.Intent
+  alias Anoma.Protobuf.Intents.Verify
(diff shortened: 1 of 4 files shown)


agureev (changes requested):
Note that this should also be based on base

apps/anoma_client/lib/api/intents.ex
```diff
@@ -64,6 +64,19 @@ defmodule Anoma.Client.Api.Servers.Intents do
...
+      |> Nock.Cue.cue!()
+      |> Transaction.from_noun()
+      |> elem(1)
+      |> Anoma.RM.Intent.verify()
+
+    bool =
+      case res do
+        true -> true
+        _ -> false
+      end
```
mariari, line 78: I feel we should have a general function that takes our error or true to a boolean in the module that has the weird signature on this. Maybe near the protocol module that will exist?

agureev: hm yeah that would be fine

however do we generally have a lot of places where we need this? this is the first one i can recall

mariari: One of the firsts, but it's a good helper to have

m1dnight:
@agureev  @juped  I rebased this on .30. Should be good to go.

m1dnight:
This is obsolete since the grpc client is no more. Nuking.

PR #1660, opened 2024-12-01
Fix double encoding error in the indexer

m1dnight:
The indexer GRPC endpoint double base64 encoded its outputs. This fixes that.


agureev:
Let's revive this branch, I do not see it merged yet and this is very important

agureev:
My mistake, this was fixed after rebasing the Indexer-GRPC connection

PR #1664, opened 2024-12-02
TRM Verification Functions Exclude Reads

agureev:
Previously the storage check for double-inserting has relied on the
reads done inside the logic. Instead, we now read outside and feed in
the commitment and nullifier sets separately.

Moreover, makes functions publicly usable.


agureev:
This is important for Client-side verification.

agureev:
However we may not need it if we have current time reads from the storage. However that might risk some desync as that might read from uncommitted storage.

PR #1667, opened 2024-12-02
List all commitments via GRPC

m1dnight:


agureev (changes requested):
This is based on e.g. `mempool_dump` which was rebased which now causes hash-conflicts

Should be rebased

agureev:
This ended up having minimal dependencies

Made it dependent on un-hardcoded examples just for ease

Issue #1638, opened 2024-11-27
Get Rid of Sleeps in Examples


agureev:
customer: @mariari 
performer:
deadline:
estimated:
started:
actual:
completed:
confirmed:
dependencies:
- [ ]
supersedes:
-
---

Some examples use `Process.sleep`. This should instead be changed to waiting for Broker messages.

PR #1669, opened 2024-12-03
Get rid of Process.sleep from examples in favour of events subscription

mariuszzak:
I have removed `Process.sleep/1` from the examples and replaced it with event subscriptions. 

Please let me know your thoughts on this approach.

---

It solves [#1638](https://github.com/anoma/anoma/issues/1638).

### DIFF
--- a/apps/anoma_node/lib/examples/esolver.ex
+++ b/apps/anoma_node/lib/examples/esolver.ex
@@ -13,6 +13,7 @@ defmodule Anoma.Node.Examples.ESolver do
   alias Anoma.Node.Transaction.Mempool
   alias Anoma.RM.DumbIntent
   alias Anoma.Node.Examples.ENode
+  alias Examples.ETransparent.ETransaction
 
   ############################################################
   #                           Scenarios                      #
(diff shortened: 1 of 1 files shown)


apps/anoma_node/lib/node/intents/solver.ex
```diff
@@ -187,20 +189,59 @@ defmodule Anoma.Node.Intents.Solver do
...
-    %{state | solved: solved, unsolved: unsolved}
+    {
+      :noreply,
+      new_state,
+      {:continue, {:emit_events, newly_solved, new_intent}}
+    }
+  end
+
+  @impl true
+  def handle_continue({:emit_events, newly_solved, new_intent}, state) do
```
agureev, line 221: Why do we want a continuation here?

agureev: I have read the PR description and now I understand better regarding the continuation use. In reality I do not think we need to do that.

What you actually probably want to do is after sending the transaction request to the Mempool, wait for the transaction even to be transmitted, at which point you do send your own events re solved intents.

I think this situation is about a compromise between situations:

1) X is solved, Solver sends event of X solved yet no transaction has been launched (i.e something failed)
vs
2) X is solved, a transaction is laucnhed yet something fails so the user didn't get to see an event re solved X

I think situation 2 is preferable. In situation 1 our replay system should catch the transaction and intentpool restart. Then the use should be able to see their intent removed from the intentpool on startup, signaling the appropriate sutuation.

agureev (commented):
Commented re continuation above

apps/anoma_node/lib/node/intents/solver.ex
mariuszzak: Hi @agureev, 

Thank you for reviewing the code! 🙏 I just want to make sure I understand your feedback correctly.

When you mentioned `Mempool`, did you mean `IntentPool`?

The flow in the current implementation of this PR is as follows:
1. A new intent is added to the `IntentPool`.
2. `IntentPool` emits an `:intent_added` event.
3. The `Solver` handles the `:intent_added` event (as it subscribes to this event), which triggers the solving logic. Once finished, it updates its state and enqueues continuation logic.
4. The `Solver` handles the continuation, which emits events: `{:intent_solved, intent}` and `{:unsolved_intent_added, intent}`.
5. In the `ESolver`, we subscribe to `:intent_solved` and `:unsolved_intent_added` to re-enable the example's execution on these events instead of relying on `Process.sleep`.

The continuation was intended to avoid a race condition:
- The `Solver` emits an event, such as `:intent_solved`.
- The `ESolver`, reacting to the event, calls `Solver.get_solved`, but the state might not be updated yet (because the `Solver` is still in the middle of the event emission process), resulting in an obsolete state being returned.

I believe this may be overkill, as the situation seems quite unrealistic.

---

I'm happy to align the code with your requirements, but I wanted to make sure we're on the same page first. 😅

agureev: Hi once again!

So there are a couple of things to remark:

I am unsure here about continuation race condition problem. Sorry if I am being silly here and not getting something but let me write out how I imagine this happening sans continuations:
  - solver gets a new intent in
  - solver solves some intents
  - if successful, it launches a transaction candidate to the Mempool
  - then emits the intents that went into the transaction candidate launched as Event messages
  - changes own state appropriately
  
  The above all happens in a one `handle_info` e.g. as a reaction to a new intent message
  
  Now suppose X heard that intent gets solved and immediately launches a message to query the appropriate Solver's state. Then it gets processed as the next message after the `handle_info` assuming an empty mailbox. At that point state gets changed modulo process crashing.
  
  So assuming that nothing crashes, no need for continuation due to Erlang semantics, no messages will be processed before the state change if all the above events happen in the same handle.
  
  So we have to worry only about restarts. Yet - you can trust me on this - restarts are fine as long as when the Solver sends a transaction candidate, it also listens to a `Mempool.Tx` event that an appropriate transaction got launched. We have a pretty cute system handling stuff like that for a Node-wide failure.
  
  So I did mean the Mempool when I said Mempool previously.
  
  Does that make things clearer? I know I am a bit vague regarding the restart system, but it would make this comment too large for a quick read :) Let me know if you want to hear more about it, however, we can have a chat somewhere about this.
  
Thanks again for all your help!

agureev: Right, yeah, sorry for the confusion! That PR was sorta just always assumed to exist and wasn't connected before we upgraded Transactions to intents. So that was a fault on my end for not explaining that properly!

mariuszzak: No worries at all, and thanks for clarifying! I appreciate your explanation—it definitely clears things up for me. 😊 I’ll make sure to pay closer attention to open PRs.

apps/anoma_node/lib/examples/esolver.ex
```diff
@@ -113,20 +115,15 @@ defmodule Anoma.Node.Examples.ESolver do
...
-      0 | 707
-    ]
-
     tx_filter = [Anoma.Node.Event.node_filter(node_id), %Mempool.TxFilter{}]
 
     with_subscription [tx_filter] do
-      Mempool.tx(
-        node_id,
-        {:transparent_resource, tx_candidate}
-      )
```
mariuszzak, line 129: I dropped it since it seems to be redundant. From what I understand the Solver is supposed to call it. Does it make sense?

ntnguyen79:
I checked again and it works fine. Thank you very much

Vào Th 4, 18 thg 12, 2024 vào lúc 21:56 Mariusz Żak <
***@***.***> đã viết:

apps/anoma_node/lib/examples/esolver.ex
agureev: Looks good to me!

Thanks once again for all your help!

agureev (approved):

mariari:
Merged in the latest release (not automatically closed as it's not against base)

Issue #1621, opened 2024-11-25
No parsing things to get port numbers / node id


juped:
customer:
performer:
deadline:
estimated:
started:
actual:
completed:
confirmed:
dependencies:
- [ ]
supersedes:
-
---

Users of the Anoma client require its gRPC port and associated node's node_id in order to call gRPC endpoints. We currently obtain these by parsing stdout from the REPL. 

We should have a way to obtain the gRPC port and node_id without parsing stdout. Especially as this will not be an option when using the binary.

### Details on how we currently parse

When testing with Juvix we start the client / node from the REPL and `puts` the `gRPC` port and `node_id`:

```
(
eclient = Anoma.Client.Examples.EClient.create_example_client
IO.puts("#{eclient.client.grpc_port} #{eclient.node.node_id}")
)
```

We then run this using the following mix command. 

NB: The `MIX_QUIET=1` is necessary to suppress Mix info logging.

```
$ MIX_QUIET=1 mix run --no-halt -e <cmd>
```

agureev:
@m1dnight is this still a relevant problem?

PR #1682, opened 2024-12-09
Store client configurations

mariuszzak:
I've been thinking about how to simplify the testing process described in [this issue](https://github.com/anoma/anoma/issues/1621). The idea is that when a client connects, it stores the port information in a temporary file, making it easily accessible to another application instance.

I'm not entirely sure if these functions fit within this module, but if you like the approach, I can relocate them to a more suitable module if needed.

Let me know if this is helpful, or if I might have misunderstood the issue.

We can also consider disabling it in prod env via config. 

---

Here is how it works:

```elixir
iex(1)> eclient = Anoma.Client.Examples.EClient.create_example_client()
%Anoma.Client.Examples.EClient{
  channel: nil,
  client: %Anoma.Client{
    grpc_port: 59755,
    supervisor: #PID<0.381.0>,
    type: :grpc
  },
  node: %Anoma.Node.Examples.ENode{
    grpc_port: 59753,
    tcp_ports: [],
    pid: #PID<0.332.0>,
    node_id: "94766831"
  },
  supervisor: nil
}
```

In another IEx:

```elixir
iex(1)> Anoma.Client.get_last_client_config()
{:ok, %{"grpc_port" => 59755, "node_id" => "94766831"}}
```

### DIFF
--- a/apps/anoma_client/lib/client.ex
+++ b/apps/anoma_client/lib/client.ex
@@ -11,6 +11,8 @@ defmodule Anoma.Client do
   alias Anoma.Client.Runner
   alias Anoma.Protobuf.Intents.Intent
 
+  @client_config_dir "./tmp/clients"
+
   typedstruct do
     field(:type, :grpc | :tcp)
     field(:supervisor, pid())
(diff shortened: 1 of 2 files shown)


mariari:
Client will get archived

PR #1691, opened 2024-12-10
A more verbose example of listing unspent resources 

m1dnight:
This example makes it clearer that unspent resources are commitments, and that those can be turned back into resources.


apps/anoma_node/lib/node/utility/indexer.ex
```diff
@@ -76,10 +76,33 @@ defmodule Anoma.Node.Utility.Indexer do
...
+  def unspent_resources(node_id) do
+    name = Registry.via(node_id, __MODULE__)
+    GenServer.call(name, :unspent_resources)
+  end
+
   ############################################################
   #                    Genserver Behavior                    #
   ############################################################
 
+  def handle_call(:unspent_resources, _from, state) do
```
agureev, line 92: Not sure what is happening here. We already have the `:resources` flag that does almost exactly this without needing to convert back to an elixir resource struct

agureev (changes requested):
We can add extra examples for checking the resources but I do not get what prompts us to duplicate the functionality we already have

PR #1694, opened 2024-12-12
RocksDB Integration

m1dnight:
Add RocksDB back into the node.


apps/anoma_lib/lib/tables.ex
```diff
@@ -0,0 +1,343 @@
...
+
+      _error ->
+        {:error, :failed_to_initialize_storage}
+    end
+  end
+
+  @doc """
+  Given an atom as table name, I create a node-specific name based on that.
+  """
+  @spec node_table_name(String.t(), atom()) :: atom()
```
agureev, line 77: This is more of a comment than anything, but that is reversed from what we originally used, where we do `String.to_atom("#{name}_#{node_id}")` basically

Not sure what's a better way of naming

apps/anoma_node/lib/node/transaction/storage.ex
```diff
@@ -165,21 +173,23 @@ defmodule Anoma.Node.Transaction.Storage do
...
-    keylist =
-      args
-      |> Keyword.validate!([
-        :node_id,
-        uncommitted_height: 0,
-        rocks: false
-      ])
-
-    node_id = keylist[:node_id]
+    args
```
agureev, line 176: I think you forgot to bind `args` here

Also it seems that the rocks option does not matter anymore, does it? Similarly to what you did in logging. Since all of this is done at the configuration level now.

apps/anoma_node/lib/node/logging.ex
```diff
@@ -589,14 +565,18 @@ defmodule Anoma.Node.Logging do
...
+  @spec init_table(String.t()) :: :ok
+  defp init_table(node_id) do
+    # initialize the tables
+    Tables.initialize_tables_for_node(node_id, @tables)
 
-  Given a Node ID, I create an appropriately named Event table for it.
-  """
+    # insert default record in the events table
+    table = Tables.node_table_name(node_id, @events_table)
+
```
agureev, line 575: We should still clear the table here due to replay semantics. Either that or we should clear the table at the supervision level before launching the Node.

However since we already do a round write here, it seems better to internalize all of this minor operations in one spot. That way all "unnatural" operations to be done for the Event table will be done in one spot.

agureev (changes requested):
Added a couple of comments.

apps/anoma_node/lib/node/transaction/storage.ex
m1dnight: No, the validate! will raise if its wrong, so there is no real need to bind it again. But I get the confusion. Ill change it.

agureev: Yes, but `uncommitted_height: 0` sets the height to 0 if none is provided.

yingqian520 (approved):

yingqian520 (commented):
IMG_0453
IMG_0308
IMG_0307
IMG_0306
IMG_0305
IMG_0280
IMG_0279
IMG_0174
IMG_0172
IMG_0171
IMG_0065

m1dnight:
Closing because there is a different PR than this one.

PR #1698, opened 2024-12-13
Capture the expected error logs

mariuszzak:
This PR mutes two error logs that are printed every time the full test suite is run. These error logs are expected though, so added assertions. I could not find `capture_log` in the codebase, so I'm unsure about your approach regarding this. WDYT?

```
==> anoma_node
Running ExUnit with seed: 502479, max_cases: 16
Excluding tags: [:zk]

...............
11:09:13.691 [error] Transaction verification failed. Reason: {:error, "Nullified resources are not committed at latest root"}
................................
11:09:13.733 [error] Transaction verification failed. Reason: {:error, "A submitted nullifier already exists in storage"}
.............................
Finished in 1.4 seconds (1.1s async, 0.3s sync)
76 tests, 0 failures
```


m1dnight:
This is a good fix! I tried it at some point, and the capture_log didnt want to work for me. I don't see anything different. Thanks a bunch for this though.

mariuszzak:
I just realised I forgot to add `assert` keyword. Now one of the examples randomly fails. I will convert the PR to draft for now. Will reopen when it's fixed.

PR #1709, opened 2024-12-18
Remove indexer logic from base

m1dnight:
There was indexer logic present on base. I do not believe that this belongs here.


m1dnight:
Rerolled on .30 @juped

Issue #1675, opened 2024-12-06
Persist shielded resource cyphertext


XuyangSong:
customer:
performer:
deadline:
estimated:
started:
actual:
completed:
confirmed:
dependencies:
- [ ]
supersedes:
-
---

<definition of done>

The verifiable encryption support in shielded resource logics is now available, along with an API for parsing logic instances and retrieving the ciphertext. Full nodes should store the ciphertext, which can be fetched and decrypted by users.

PR #1710, opened 2024-12-18
store cairo cipher texts

XuyangSong:
based on #1562
to solve #1675 

@agureev 
I've added an API to get ciphertexts from Cairo transactions. As discussed, could you help store the ciphertexts and provide an API for users to fetch them. I left a TODO [here](https://github.com/anoma/anoma/blob/dc3ea16777cab670cc2f766b79ecce42766d9ca6/apps/anoma_node/lib/node/transaction/backends.ex#L455) in the cairo backend.

### DIFF
--- a/apps/anoma_lib/lib/anoma/cairo_resource/logic_instance.ex
+++ b/apps/anoma_lib/lib/anoma/cairo_resource/logic_instance.ex
@@ -62,11 +62,19 @@ defmodule Anoma.CairoResource.LogicInstance do
     }
   end
 
-  @spec decrypt(list(binary()), binary()) :: list(binary())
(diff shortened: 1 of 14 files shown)


agureev (changes requested):
I submitted the logic to append the ciphertexts using the provided function along with some extra checks in current examples.

That is, the ciphertexts namespace now has a set of maps with a tag and ciphertext keys. 

All other stuff seems fine to me as well.

One thing I wanted to request once you get done with core logic: could you please move all the cairo example tests into a separate test folder and preface it with a `:zk` moduletag like we do for the cairo examples in the `anoma_lib` application?

This way the default option to run tests will always exclude any cairo stuff.

PR #1713, opened 2024-12-18
Change Computations for Jet Mugs

agureev:
Previously mug computations relied on logics cores and disallowed any compile-time mug computations due to 9 calls. Fixes this.


yingqian520 (approved):

yingqian520 (commented):
5510290137135249

Issue #1676, opened 2024-12-07
Anoma.Protobuf.IntentsService.Verify crashes if invalid intent is sent


paulcadman:
customer: @paulcadman 
performer:
deadline:
estimated:
started:
actual:
completed:
confirmed:
dependencies:
- [ ]
supersedes:
-
---

## Steps to reproduce:

1. Checkout anoma at `lukasz/juvix-integration-tracking` : `81601ab7ee8e7ceade7e7cb00bdcc5b65aef77c9`
2. Start a client / node
3. Submit an invalid intent to `Anoma.Protobuf.IntentsService.Verify`, e.g the base64 encoding of the string "invalid":

```
jq -n --arg id "" --arg intent $(base64 <<< "invalid") '{node_info: {node_id: $id}, intent: {intent: $intent}}' | grpcurl -plaintext -d @ localhost:49776 Anoma.Protobuf.IntentsService.Verify
```

### Expected:

The response to be:

```
{
  "valid": false
}
```

### Actual

```
ERROR:
  Code: Unknown
  Message: Internal Server Error
```

## Valid intents work as expected

If you submit the following valid intent, the client responds as expected:

[trans.nockma.gz](https://github.com/user-attachments/files/18048614/trans.nockma.gz)

```
$ jq -n --arg id "" --arg intent $(base64 < trans.nockma) '{node_info: {node_id: $id}, intent: {intent: $intent}}' | grpcurl -plaintext -d @ localhost:49776 Anoma.Protobuf.IntentsService.Verify
{
  "valid": true
}
```

agureev:
Why is returning false the desired semantics here?

paulcadman:
> Why is returning false the desired semantics here?

I'm not sure if it's desired but this is how it's defined in the corresponding GRPC definition:

https://github.com/anoma/anoma/blob/6b1e3cae5db718abe0eddc03c0a934b1b29346ba/apps/anoma_protobuf/priv/protobuf/intents/verify.proto#L11

i.e if the intent in the request is valid the response should be:

```
{"valid": true}
```

if the intent is invalid the response should be:

```
{"valid": false}
```

The response should certainly never be "Internal Server Error".

mariuszzak:
Hi there, I opened a [PR](https://github.com/anoma/anoma/pull/1718) fixing this issue. Please let me know if it works as expected. Thanks!

m1dnight:
[...]

PR #1718, opened 2024-12-19
Don't crash when invalid intent is passed to Intents.verify

mariuszzak:
Reproduced and fixed the issue described in https://github.com/anoma/anoma/issues/1676.

With the fix, `Anoma.Protobuf.IntentsService.Verify` with invalid intent does not crash.

### DIFF
--- a/apps/anoma_client/lib/api/intents.ex
+++ b/apps/anoma_client/lib/api/intents.ex
@@ -78,5 +78,12 @@ defmodule Anoma.Client.Api.Servers.Intents do
       end
 
     %Verify.Response{valid: bool}
+  rescue
+    _ in Noun.Jam.CueError ->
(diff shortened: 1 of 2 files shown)


m1dnight:
Hi Marius, 

Thanks for this PR. It looks good to me. However, I've been working on a more general approach for error handling, and I would like to incorporate this work into that branch. So please be patient until that happens. 

Appreciate the work a lot! 🙌

mariuszzak:
Hi @m1dnight,

Thanks for checking! Feel free to incorporate this as needed. Let me know if I can assist further.

mariari:
I don't see verify anymore. I plan to deprecate client after this. Also it may have been merged since the target of this is not base, it may have already made it into a previous release.

PR #1740, opened 2024-12-30
Make App Data a Map in RM Library

agureev:
1. Change Nock stdlib Action datatype to have maps instead of binary for app data
2. Implement set to_noun and from_noun functionality
3. Change the Elixir Actions datatype accoridngly


agureev:
IMPORTANT: We need to take care when merging this to coordinate with Juvix as this is potentially a breaking change.

Tnuwan99:
???

PR #1743, opened 2024-12-31
Update crypto.lisp

Mukzid:
I've added some comments for clarity and adjusted the verify-nullifier function to return the result of ironclad:verify-signature. Also, I've corrected the format of the create-nullifier function to properly handle the resource parameter

### DIFF
--- a/experiments/cl-resource/crypto.lisp
+++ b/experiments/cl-resource/crypto.lisp
@@ -1,28 +1,33 @@
+
 (in-package :resource-machine)
 
+;; Convert an ed25519 private key to an integer representation
 (-> nullifier-key (ed25519-private-key) integer)
 (defun nullifier-key (key)
   (values (octets->uint (ed25519-key-x key) 32)))
 
+;; Reconstruct an ed25519 private key from an integer
 (-> reconstruct-private (integer) ed25519-private-key)
 (defun reconstruct-private (number)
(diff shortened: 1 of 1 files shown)


experiments/cl-resource/crypto.lisp
```diff
@@ -1,28 +1,33 @@
...
-   (octets->uint (ironclad:sign-message key (int64->octets (nullifier-hash resource)))
-                 64)))
+   (octets->uint (ironclad:sign-message key (int64->octets (nullifier-hash resource))) 64)))
 
-;; Question when do even have the info to call verify-nullifier?
+;; Verify a nullifier using the public key, message, and signature
 (-> verify-nullifier (ed25519-public-key t  (simple-array (unsigned-byte 8) (64))) t)
 (defun verify-nullifier (key message signature)
-  (verify-signature key message signature))
+  (values (ironclad:verify-signature key message signature)))
```
mariari, line 28: Why qualify the function, I probably should have.

the `values` here I believe is pointless as they are only there to satisfy the CL type checker, however that part of the signature hasn't changed

mariari (commented):
The comments are mostly pointless, as the type signature covers what they do given the name.

Further the documentation strings aren't in the proper CL format for being documentation when querying information in the CL system

PR #1752, opened 2025-01-09
fix: enhance ID generation with validation and configuration options

crStiv:
Added improvements to ID generation:
- Prefix validation using regex pattern
- Configurable random bytes length
- Bang (!) version of generation function
- Proper error handling with {:ok, result} | {:error, reason} tuples
- Type specifications and improved documentation

Breaking changes:
- `prefix_random_id/1` now returns tuple instead of string

### DIFF
--- a/apps/anoma_node/lib/node.ex
+++ b/apps/anoma_node/lib/node.ex
@@ -6,13 +6,20 @@ defmodule Anoma.Node do
     Anoma.Supervisor.start_link(args)
   end
 
+  @default_bytes_length 16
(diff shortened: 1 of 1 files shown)


mariari:
Thank you for your contribution.

I do have a few notes on this topic:

1. what does this bit of customization give us? It seems to come at the cost of having our id generation return a `maybe`.
2. I believe our id generation is intentionally dumb and is left for a future improvement on an actual design for ID's. Until then I think we should focus on making it as simple and understandable as possible.

Issue #1654, opened 2024-11-29
Refactor the compliance and the trivial logic juvix circuits and encapsulate the common impls into juvix-lib


XuyangSong:
customer:
performer:
deadline:
estimated:
started:
actual:
completed:
confirmed:
dependencies:
- [ ]
supersedes:
-
---

<definition of done>

Abstract and encapsulate commonly used functions (commitment, nullifier, encryption, merkle path check, etc.) from compliance and trivial logic juvix circuits

discuss with @lukaszcz on integrating them into application-juvix-lib for easier application development

PR #1753, opened 2025-01-09
Use a Map for logic_proofs in the cairo action struct

### LINKED ISSUE #1646: Use a Map instead of a List for logic_proofs in the action of Cairo RM
customer:
performer:
deadline:
estimated:
started:
actual:
completed:
confirmed:
dependencies:
- [ ]
supersedes:
-
---

<definition of done>

The current implementation uses a list of ProofRecord for logic_proofs in each action. However, the order of logic_proofs must strictly match the order of resources in compliance proofs to ensure successful correspondence check between logics and compliances. It would be more efficient to use a map instead, using tags as keys, which would eliminate the need for a default order.

XuyangSong:
We no longer need to worry about the specific order of logic proofs in the Cairo action by using a Map, as it simplifies checks in actions and transactions. This change is already included in the latest Spec.

close #1646 
based on #1654

### DIFF
--- a/apps/anoma_lib/lib/anoma/cairo_resource/action.ex
+++ b/apps/anoma_lib/lib/anoma/cairo_resource/action.ex
@@ -20,8 +20,21 @@ defmodule Anoma.CairoResource.Action do
   alias Anoma.Constants
 
   typedstruct enforce: true do
-    field(:logic_proofs, list(ProofRecord.t()), default: [])
+    field(:logic_proofs, %{binary() => ProofRecord.t()}, default: {})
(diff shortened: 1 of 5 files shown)


XuyangSong:
> Introduces a dialyzer error, revisit after 0.30

Thank you for the feedback. @juped 
Could you provide more details on how to reproduce the error? I tried running `mix dialyzer` but couldn't encounter it. It seems the CI is not complaining. 
I only rebased it to 0.30 rn.

mariari:
Yeah I see no dialyzer errors myself, I've merged it into next

PR #1776, opened 2025-01-15
Artem/devnet timeout

agureev:


agureev:
Closed as now txs are fast enough

PR #1781, opened 2025-01-15
Artem/juvix integration v0.30

agureev:
Integration branch between 0.30 and 0.31 releases for Juvix use.


agureev:
Deprecated in favor of the final testnet-1 release

PR #1782, opened 2025-01-16
Mute the expected error log in examples

mariuszzak:
Similar to https://github.com/anoma/anoma/pull/1698

It mutes another error log:

```
[error] Transaction verification failed. Reason: {:error, "Nullified resources are not committed at latest root"}
```

With this change the output of `mix test` is nice and clean.

### DIFF
--- a/apps/anoma_node/lib/examples/eintent_pool.ex
+++ b/apps/anoma_node/lib/examples/eintent_pool.ex
@@ -5,6 +5,7 @@ defmodule Anoma.Node.Examples.EIntentPool do
 
   require ExUnit.Assertions
   import ExUnit.Assertions
+  import ExUnit.CaptureLog
 
   alias Anoma.Node.Intents.IntentPool
   alias Anoma.RM.DumbIntent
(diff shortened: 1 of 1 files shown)


mariari:
@mariuszzak I think I've considered capture logs before but they have issues when running in the repl

```elixir
iex(anoma@YU-NO)4> Anoma.Node.Examples.EIntentPool.add_intent_with_submitted_commitments
** (exit) exited in: GenServer.call(ExUnit.CaptureServer, {:log_capture_on, #PID<0.805.0>, #PID<0.1242.0>, []}, :infinity)
    ** (EXIT) no process: the process is not alive or there's no process currently associated with the given name, possibly because its application isn't started
    (elixir 1.17.3) lib/gen_server.ex:1121: GenServer.call/3
    (ex_unit 1.17.3) lib/ex_unit/capture_log.ex:110: ExUnit.CaptureLog.with_log/2
    (ex_unit 1.17.3) lib/ex_unit/capture_log.ex:75: ExUnit.CaptureLog.capture_log/2
    (anoma_node 0.30.0) lib/examples/eintent_pool.ex:134: Anoma.Node.Examples.EIntentPool.add_intent_with_submitted_commitments/1
    iex:4: (file)

```

they are sort of annoying to deal with, but overall I do like the idea of using these for these kinds of errors when running mix test

mariari:
I need to experiment with injecting

```elixir
iex(anoma@YU-NO)8> ExUnit.CaptureServer.start_link([])
{:ok, #PID<0.1364.0>}
```

on application startup to avoid the issues of trying to run examples

mariari:
I endedup making a topic here about it:

https://github.com/anoma/anoma/pull/1785

agureev:
@mariari  please exclude this topic from next

this is a bug, this should not error at all

PR #1783, opened 2025-01-16
Change Jet Checks

agureev:
Prior, the jet registry carried the mug of the parent to compare
against. However, mug computations for such large core can be
expensive in Elixir due to murmur3 limitations.

In order to speed up the process, we now store instead of mugs the
actual parents and can compare against them during the jet checks.


agureev:
On my machine Anoma Lib tests now run 2x times faster due to the change.

PR #1788, opened 2025-01-16
Only process execution if transaction is in mempool

m1dnight:
When the mempool receives an execution event for a transaction that is not in the mempool, it errored. 
This fixed it and prints a warning instead.


agureev:
Why is it a bug? A mempool has a unique track of the transactions it is concerned for. If a system somehow executes a transaction that is not account for it should panic. Mempool is the source of truth of all pending transactions.

m1dnight:
Closing this as per meeting 23/1/24. Will not be handling this error right now. Instead, figure out a more generic crash handling behavior in the node.

PR #1789, opened 2025-01-16
Launch the transaction before sending the transaction event

m1dnight:
This PR sends the transaction event *after* the transaction has been launched.
This ensures the supervisor has the child ready when a process handles the tx event.

### DIFF
--- a/apps/anoma_node/lib/node/transaction/mempool.ex
+++ b/apps/anoma_node/lib/node/transaction/mempool.ex
@@ -400,9 +400,9 @@ defmodule Anoma.Node.Transaction.Mempool do
     value = %Tx{backend: backend, code: code}
     node_id = state.node_id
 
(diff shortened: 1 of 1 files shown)


agureev (commented):
I need convincing. Why is this good?

The reason the event fired earlier is because it is in our interest to record the tx launch as soon as possible.

agureev:
> Because if you fire the event before you actually launch the transaction you can have a state where there is a transaction event, while there isnt actually a transaction.

This change isn't addressing this. Executor.launch is a cast.

The event is more about that the tx candidate has been submitted rather than anything else.

agureev:
in other words if you want to have a message that guarantees a tx is launched upon reception put it into a handle_cast in the Executor

TxEvent has a different purpose

m1dnight:
Closing this as it's wrong.

PR #1792, opened 2025-01-16
Replay logic

m1dnight:
This PR adds the replay logic to the node.


apps/anoma_node/lib/node/logging.ex
```diff
@@ -299,7 +299,7 @@ defmodule Anoma.Node.Logging do
 
       current_pending = match(:consensus, table)
       :mnesia.write({table, :consensus, tl(current_pending)})
-      :mnesia.write({table, :round, round})
+      :mnesia.write({table, :round, round + 1})
```
agureev, line 302: Why do you need to increase the round here?

PR #1806, opened 2025-01-20
Add GRPC Endpoint for Read-Only Transactions

agureev:
Adds a GRPC interface for submitting read-only transactions and getting a reply in a jammed format.


apps/anoma_client/lib/client/client_connection/grpc_proxy.ex
```diff
@@ -93,6 +95,10 @@ defmodule Anoma.Client.Connection.GRPCProxy do
     GenServer.call(__MODULE__, {:add_transaction, jammed_nock})
   end
 
+  def add_read_only_transaction(jammed_nock) do
```
m1dnight, line 98: Typespecs! (I noticed I forgot one too, so ill fix that :)

m1dnight:
Minor comments, lgtm

PR #1812, opened 2025-01-23
Validation of GRPC requests in the node

m1dnight:
This PR adds validation to the GRPC requests in the node. 

The summary as to why this is a thing: 
 - In protobuf everything is optional (see: https://protobuf.dev/programming-guides/proto3/#field-labels)
 - GRPC crashes agressively, and no interception of that is possible (afaict)
 - Validating requests allows us to return reasonable error messages to the user.


rafsyah22:
I don't know

PR #1813, opened 2025-01-23
Adjustments to Cairo RM Structs

### LINKED ISSUE #1277: Adjust commitments, nullifiers, and roots in TX struct
customer:
performer:
deadline:
estimated:
started:
actual:
completed:
confirmed:
dependencies:
- [x] anoma/green#715 
supersedes:
-
---

<definition of done>

Essentially, the commitments, nullifiers, and roots should be obtained from the TX struct instead of being included in TX itself. 

In shielded cases, we can retrieve them from the public inputs of circuit proofs. 
In transparent cases, they are recalculated in the partition_resources function. The reason is that when it comes to transactions, we can't really trust the data (like commitments, nullifiers, and roots) provided by users.

It is possible to return commitments, nullifiers, and roots from the verify function for both transparent and shielded transactions.

XuyangSong:
To sync with the updated RM Specs:
* use a set for roots in shielded tx
* move commitments and nullifiers from tx to action

based on #1753 
close #1277


apps/anoma_lib/lib/anoma/cairo_resource/action.ex
```diff
@@ -20,14 +20,43 @@ defmodule Anoma.CairoResource.Action do
   alias Anoma.Constants
 
   typedstruct enforce: true do
-    field(:logic_proofs, list(ProofRecord.t()), default: [])
-    field(:compliance_proofs, list(ProofRecord.t()), default: [])
+    field(:created_commitments, list(<<_::256>>), default: [])
+    field(:consumed_nullifiers, list(<<_::256>>), default: [])
+    field(:logic_proofs, %{binary() => ProofRecord.t()}, default: {})
```
agureev, line 25: So formally this is a little bit off. 

The resourceLogicProofs field is of type `binary() => {logicHash, ResourceLogic.Proof}`

apps/anoma_lib/lib/anoma/cairo_resource/action.ex
```diff
@@ -20,14 +20,43 @@ defmodule Anoma.CairoResource.Action do
...
+    field(:compliance_units, MapSet.t(ProofRecord.t()), default: MapSet.new())
+  end
```
agureev, line 27: We also need application data here of type

`%{binary() => list(binary(), bool())}`

apps/anoma_lib/lib/anoma/cairo_resource/action.ex
```diff
@@ -52,71 +92,108 @@ defmodule Anoma.CairoResource.Action do
 
   @spec verify(t()) :: boolean()
   def verify(action) do
```
agureev, line 94: Does this check that compliance units uniquely partition the action they are contained in?

agureev: This is not explicitly written in the spec but I think we care about that, do we not?

apps/anoma_lib/lib/anoma/cairo_resource/action.ex
agureev: the bool element is just whether you want to keep or get rid of the data in blob storage after tx verification

apps/anoma_lib/lib/anoma/cairo_resource/action.ex
```diff
@@ -127,8 +204,8 @@ defmodule Anoma.CairoResource.Action do
   end
 
   @spec verify_compliance_hash(list(ProofRecord.t())) :: boolean()
-  def verify_compliance_hash(compliance_proofs) do
-    compliance_proofs
+  defp verify_compliance_hash(compliance_units) do
```
agureev, line 207: What does this correspond to anything in the spec? Or is this Cairo-specific?

Just asking whether we should ask Yulia to add something like this or not.

apps/anoma_lib/lib/anoma/cairo_resource/action.ex
agureev, line 25: I think this is not the precise type from the specs

ProofRecord is a datatype with 2 elements. hile the compliance unit in the spec has three: proof, instance, and verification key

https://specs.anoma.net/latest/arch/system/state/resource_machine/data_structures/compliance_unit/compliance_unit.html

apps/anoma_lib/lib/anoma/cairo_resource/action.ex
```diff
@@ -52,71 +92,108 @@ defmodule Anoma.CairoResource.Action do
...
     else
       _ -> false
     end
   end
 
   @spec verify_proofs(list(ProofRecord.t()), binary() | nil) :: boolean()
-  defp verify_proofs(proofs, debug_msg \\ nil) do
+  defp verify_proofs(proofs, debug_msg) do
     Enum.reduce_while(proofs, true, fn proof_record, _acc ->
       res = ProofRecord.verify(proof_record)
```
agureev, line 194: When I look at the definition of `verify` for Prove Systems, I see that they need three arguments

I think it is a requirement we have something similar for all Proof Systems

agureev (changes requested):
I left a lot of comments. Note that the only thing I was noting is discrepancies with the specs. So if there is something specs asks for which is not good, just let me know and we will discuss it with Yulia.

The key discrepancies I currently see are:

- appdata missing
- ComplianceUnit missing as a separate datatype which it seemingly needs to be
- verify for proving system taking the wrong number of arguments

with other miscellaneous comments

agureev:
I also think we have some interfaces missing here, like explicit functions to create, compute deltas etc.

But I think these are minor. It would be a good thing to have nevertheless.

apps/anoma_lib/lib/examples/e_cairo/e_transaction.ex
```diff
@@ -161,17 +163,21 @@ defmodule Examples.ECairo.ETransaction do
...
+        delta_proof: priv_keys
       }
 
     composed_shielded_tx =
-      Anoma.RM.Transaction.compose(shielded_tx_1, shielded_tx_2)
+      %Transaction{
+        roots: MapSet.union(shielded_tx_1.roots, shielded_tx_2.roots),
+        actions: shielded_tx_1.actions ++ shielded_tx_2.actions,
+        delta_proof: shielded_tx_1.delta_proof <> shielded_tx_2.delta_proof
+      }
```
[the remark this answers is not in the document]
agureev: Oh do we not?

This is bad, specification insists on `compose` being the interface for transactons

We need this for solving as well

apps/anoma_lib/lib/anoma/cairo_resource/transaction.ex
```diff
@@ -67,145 +62,113 @@ defmodule Anoma.CairoResource.Transaction do
...
     end
 
-    # TODO: We can return roots, commitments, and nullifiers instead of just a
-    # boolean value so that we can get rid of them in the Transaction struct. We
-    # can apply the same improvement to the transparent Transaction.
     @impl true
     def verify(transaction = %Transaction{}) do
-      with true <- verify_proofs(transaction),
+      with true <- verify_actions(transaction),
            true <- verify_duplicate_nfs(transaction),
```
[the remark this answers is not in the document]
XuyangSong: checks on duplication of commitments?
We don't need it in shielded cases. The uniqueness of commitment is proved in the compliance proof.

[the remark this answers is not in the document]
agureev: should we discuss moving these around with Yulia?

apps/anoma_lib/lib/anoma/cairo_resource/action.ex
XuyangSong: the logicref is already in the `ProofRecord`.
I really don't want to encode and decode the cairo proof/verifying key/instance(`ProofRecord`) back and forth in Elixir and Juvix since the cairo proof is a complicated struct. 
IMP, the complex struct should remain hidden from Elixir and Juvix. I have provided APIs to access them.

XuyangSong: Well, I will create a new type to explicitly include the logicHash and ResourceLogic.Proof

apps/anoma_lib/lib/anoma/cairo_resource/action.ex
XuyangSong: The app_data will be in a separate PR as I mentioned in the meeting.

apps/anoma_lib/lib/anoma/cairo_resource/action.ex
XuyangSong: I don't fully understand your question. Could you provide more details?

agureev: Sure!

So each compliance unit contains information of created and consumed resources. The idea as that, as stated in the spec, 

`the sets of resources covered by any two compliance units cover don't intersect, and together the compliance units cover all of the resources in the action.`

Does this make sense?

apps/anoma_lib/lib/anoma/cairo_resource/action.ex
XuyangSong: The Cairo proving system provides only `proof` and `instance`, with the verification key included in the `instance`. 

A potential solution is to hack Cairo to parse and return three elements. Is this necessary?

agureev: @vveiln tagging you for reference

apps/anoma_lib/lib/examples/e_cairo/e_transaction.ex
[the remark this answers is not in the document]
XuyangSong: why? The `Transaction.compose` is still there.
I manually compose the example to avoid triggering the checks in `Transaction.compose`

apps/anoma_lib/lib/anoma/cairo_resource/action.ex
XuyangSong: Did I miss any checks?
The uniqueness check of commitments and nullifiers confirms nonintersection. It's obvious all resources are covered.

apps/anoma_lib/lib/anoma/cairo_resource/action.ex
XuyangSong: Sorry for misleading. I thought it's logic hash checks.
We need to check the compliance hash which is a fixed one

apps/anoma_lib/lib/anoma/cairo_resource/action.ex
agureev: Unless you check somewhere that compliance units do that and I have not noticed

If so, maybe you could explain it to me :)

apps/anoma_lib/lib/anoma/cairo_resource/action.ex
agureev: I see, could you elaborate what a compliance hash is in Cairo and how it may related to concepts in the spec?

i..e whether this corresponds to some check already listed in spec or whether it is general enought that we should ask to add it to specs

apps/anoma_lib/lib/anoma/cairo_resource/action.ex
XuyangSong: I am unsure of the origin of the concept of compliance units partitioning actions. 
I am sure that all resources are covered by compliance units, which do not intersect since each compliance unit produces unique outputs, namely commitment and nullifier.

@vveiln Do you know which specific checks you recently added?

agureev: `since each compliance unit produces unique outputs, namely commitment and nullifier`

@XuyangSong could you elaborate what you mean by "produces" here?

XuyangSong: The compliance unit creates and consumes some resources, then outputs/produces the corresponding commitments and nullifiers.

apps/anoma_lib/lib/anoma/cairo_resource/action.ex
XuyangSong: For transparent cases, compliance constraints are checked by the hardcoded program. 
For shielded cases, compliance proofs must be verified using the expected verifying key, so I check if the hash of the verifying key matches the hardcoded value.

apps/anoma_lib/lib/anoma/cairo_resource/action.ex
agureev: Hm, I am just asking because a compliance unit is a datatype in the spec, so it does not "do" anything. I guess you are referring to some interface which we are using?

Sorry if I am being stupid here :(

apps/anoma_lib/lib/anoma/cairo_resource/action.ex
agureev: Actually I do that as well for my case

The verifying key in the transparent case is also a hardcoded predicate encoding of checking compliance-related issues. So should we maybe raise an issue making it a general check if we do it either way?

Or is this maybe presupposed by type-checking `vk` field in CU maybe..?

apps/anoma_lib/lib/anoma/cairo_resource/action.ex
XuyangSong: yeah, you're right. Let me rephrase it.
The commitments and nullifiers are components of instances in a compliance unit. 
For zkVM, witnesses are often referred to as inputs and instances as outputs, which is more intuitive for a virtual machine.

agureev: Thanks so much! That makes perfect sense!

So returning then to this point:

> which do not intersect since each compliance unit produces unique outputs, namely commitment and nullifier.

So do we check that uniqueness of outputs anywhere? Maybe I missed that somewhere. I am sorry if I did.

apps/anoma_lib/lib/anoma/cairo_resource/action.ex
XuyangSong: It should be explicitly specified in the spec, IMP.

apps/anoma_lib/lib/anoma/cairo_resource/action.ex
XuyangSong: The uniqueness of commitments is guaranteed by shielded compliance constraints. I'll wait for Yulia's reply to confirm if a duplicate check is needed. 
The uniqueness of nullifiers is checked.

agureev: Are you referring to this function? Does this do the checking?

` defp verify_duplicate_nfs(tx) do`

XuyangSong: yes, the uniqueness in current tx is checked by this one, and historical uniqueness will be verified during storage.

agureev: So, note that this function checks that there are no duplications between actions. 

i.e. it calls Transaction.nullifiers which then goes through the nullifiers listed in the actions

this seems to me not to imply anything regarding compliance units

i might have just one action `A` in my tx `T`,  so that my nullifiers and commitments are always unique, so that check always passes

suppose that in `A` there is only one nullifier `[nf]` in its nullifier field. now suppose it also has two compliance units, both of which have that nullifier in their instance fields

that would break the fact that compliance units partition the action as stated in the spec, would it not?

XuyangSong: > this seems to me not to imply anything regarding compliance units

There is a check to ensure nfs in the list match the nfs in compliance instances and in logic_proofs indices (tags).

> suppose that in A there is only one nullifier [nf] in its nullifier field. now suppose it also has two compliance units, both of which have that nullifier in their instance fields

There are two nfs in compliance units, but only one in the list. They don't match.

XuyangSong: https://github.com/anoma/anoma/blob/5d0d3de168310758afcb4366372c1501999cf026/apps/anoma_lib/lib/anoma/cairo_resource/action.ex#L126

https://github.com/anoma/anoma/blob/5d0d3de168310758afcb4366372c1501999cf026/apps/anoma_lib/lib/anoma/cairo_resource/action.ex#L140

agureev: > There is a check to ensure nfs in the list match the nfs in compliance instances and in logic_proofs indices (tags).

Ah, I see it now, thanks for the pointers and apologies for such a hassle!

This I think indeed does exactly what I asked about.

But one last small question:

for https://github.com/anoma/anoma/blob/5d0d3de168310758afcb4366372c1501999cf026/apps/anoma_lib/lib/anoma/cairo_resource/action.ex#L112

wouldn't you want to first check that this does not have repeating elements?

Other than that, I have no questions and can consider this specific convo closed!

XuyangSong: Ideally, we should construct the compliance instance directly from the nf list instead of checking matches..
I don't want to hack cairo system and parse and reconstruct the compliance instance, which is inefficient and annoying. So there might be some duplicate data. 
I'll try to figure it out later. It seems we can get the list of nfs from compliance instances to avoid the correspondence check.

XuyangSong: > wouldn't you want to first check that this does not have repeating elements?

We still need to check the repeating elements for the tx scope, even with this check.
I can add it if you think it's needed.

apps/anoma_lib/lib/anoma/cairo_resource/action.ex
XuyangSong: fixed

apps/anoma_lib/lib/anoma/cairo_resource/action.ex
XuyangSong: I'll create a separate PR to modify the `verify` API to accept tree arguments, as this requires some changes in the aarm-cairo repo.
Since `vk` is already included in the `proof`, maybe I can set `vk` to empty to avoid unnecessary encoding and decoding the `proof`, similar to your approach in TRM for specs-compliance. Then I don't need to change the serialization and deserialization in the underlying cairo proving system. Wdyt?

XuyangSong:
replaced by #1915

PR #1816, opened 2025-01-24
Remove Executor Timeouts

agureev:
Previously if a transaction in a consensus would not finish in 5 seconds, the Executor would error causing a system crash.

This removes such a block. It should instead be replaced by gas timeouts.


agureev:
This is a good PR for devnet purposes and for the purposes of testing. 

However this should be soon followed by proper gas VM changes.

PR #1850, opened 2025-01-29
Make TRM Resource Logic to be a Hash

agureev:
Prior, resource logic was an actual Nockma core. Now, it is a sha256
of the jam of the core, making it specs-compliant.

Changes the verification for logic proofs appropriately.


mariari:
Dead end doesn't grow into anything and is suplanted by trvm2

PR #1851, opened 2025-01-29
Make TRM Resource Data Field be of Type Atom

agureev:
Was a mismatch before.


mariari:
Dead end doesn't grow into anything and is suplanted by trvm2

Issue #1285, opened 2024-08-20
Byte Array Representation


mariari:
customer:
performer:
deadline:
estimated:
started:
actual:
completed:
confirmed:
dependencies:
- [ ]
supersedes:
-
---

Everywhere where we use nock, we should update the interface of our binary from being just the binary in nock to being `[size binary]`

A good example of this at work is

```elixir
  def from_noun([logic, label, quantity, data, eph, nonce, npk | rseed]) do
    %Resource{
      logic: logic,
      label: Noun.atom_integer_to_binary(label),
      quantity: quantity,
      data: Noun.atom_integer_to_binary(data),
      eph:
        case eph do
          0 -> true
          1 -> false
        end,
      nonce: Noun.atom_integer_to_binary(nonce, 32),
      npk: Noun.atom_integer_to_binary(npk, 32),
      rseed: Noun.atom_integer_to_binary(rseed, 32)
    }
  end
```

Where we enforce the `nonce` `npk` and `rseed` to be 32 bytes long. 

The encoding should be that we assert that the size is `32` but wee expected the nock values to be `[size binary]`.

This makes the representation uniform, as this is how one has to handle all binary data of unknown quantity.

If we do not make this change this:

1. Makes it harder for the Juvix team, as they have to inconsistently encode their Arrays
2. May allow bugs in our implementation as we have to special case them in our implementation, rather than just asserting that our sizes are not violated

PR #1861, opened 2025-02-08
Pad binary data for verify_detached jet to required size

paulcadman:
A ed25519 signature is 64 bytes and a verifying key is 32 bytes. These
are passed to the verify_detached jet as atoms. These atoms must be
padded to the expected size otherwise the verification can fail.

For example, as part of the Kudos application test the following
signature was generated:

1e66a254ff3897aa6657e418258b9a8ebc59f165deb17858f18e56e48f1f4bc63519c4d3d0e776835bbcc1a955e6310c11df59de27d907612667063b09a6d0

The last byte of this signature is 0 so when its atom representation is
converted back to bytes by Noun.atom_integer_to_binary, the resulting
binary is only 63 bytes.

The final byte is missing and so the signature verification check fails.

The only relevant commit in this PR is https://github.com/anoma/anoma/pull/1861/commits/bd485d54ea350c692055e4e0bb619734f39e415c

The `message` argument to this jet has unknown size and so cannot be padded. This might cause hard to debug verification failures in the future. Perhaps we should consider the proposal to encode these arguments as a cell `[byte_size atom_payload]`, see: https://github.com/anoma/anoma/issues/1285.

Other jets, where the input is assumed to be a binary of an expected size, could also be affected by this issue.

### DIFF
--- a/apps/anoma_lib/lib/nock/jets.ex
+++ b/apps/anoma_lib/lib/nock/jets.ex
@@ -305,9 +305,9 @@ defmodule Nock.Jets do
       when is_noun_atom(a) and is_noun_atom(b) and is_noun_atom(c) ->
         try do
           if Sign.verify_detached(
-               Noun.atom_integer_to_binary(a),
+               Noun.atom_integer_to_binary(a, 64),
                Noun.atom_integer_to_binary(b),
-               Noun.atom_integer_to_binary(c)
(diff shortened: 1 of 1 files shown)


agureev (changes requested):
Needs to change ENock test as this changes sign semantics.

PR #1871, opened 2025-02-12
CairoRM Intents Support

agureev:
Add intent interface and appropriate functionality for IntentPool and Solver processing for CairoRM.


mariari:
For some reason I have issues with this topic when it's merged with the latest release candidate, the filtering of the mapset filter does not work.

Merged against: 3396e2ccf80a8bfeba17725c30b1bd7f4255893b

PR #1884, opened 2025-02-17
Adapt the Cairo workflow to use parsed JSON objects as input

lukaszcz:


agureev (approved):
Looks good to me! Thank you for your work!

I think with that once we get approval for CairoRM V2 I will be able to provide this function from the standard library.

PR #1885, opened 2025-02-17
Fix version comparison and improve compatibility

indmind:
I've fixed the version comparison logic to ensure it works correctly across different platforms. Replaced `tac` with `sort -Vr` for better compatibility. Also removed an unused `app` variable and confirmed that `EOF` properly closes the `cat` block. These changes should make the script more reliable and portable.


mariari:
Included in next for an inclusion in a Release Candidate

Issue #1880, opened 2025-02-14
Adjust Proof struct and APIs in cairo RM


XuyangSong:
customer:
performer:
deadline:
estimated:
started:
actual:
completed:
confirmed:
dependencies:
- [ ]
supersedes:
-
---

<definition of done>

In the Cairo RM's `ProofRecord` struct, we have `proof` and `public_inputs`. `public_inputs` includes the verifying key, instances, and other public information (such as public memory). Consider explicitly adding the verifying key and modifying the corresponding APIs (`prove` and `verify`) to align with the Specs definitions. Think about how to prevent hacking and rewriting the serialization and deserialization of `public_inputs` in the underlying cairo proving system lib.

Issue #1858, opened 2025-02-06
Move app_data from instance to action in cairo RM


XuyangSong:
customer:
performer:
deadline:
estimated:
started:
actual:
completed:
confirmed:
dependencies:
- [#1813]
supersedes:
-
---

<definition of done>

Issue #1815, opened 2025-01-23
Add the isConsumed flag to the resource logic in Cairo RM


XuyangSong:
customer:
performer:
deadline:
estimated:
started:
actual:
completed:
confirmed:
dependencies:
- [ ]
supersedes:
-
---

<definition of done>
As discussed, `isConsumed` is already included in the latest RM Specs. 
Remove the previous odd/even index flag for specifying `isConsumed` and update the corresponding checks.

Issue #1653, opened 2024-11-29
Offer a default padding shielded resource construction


XuyangSong:
customer:
performer:
deadline:
estimated:
started:
actual:
completed:
confirmed:
dependencies:
- [ ]
supersedes:
-
---

<definition of done>

What's a padding shielded resource?
The compliance unit currently consists of a pair of resources (one input and one output resource). It is common to have padding resources when the actions' resources are not in pairs. Padding resources must always have a quantity of 0 to maintain the balance.

Offer a default padding resource construction and user-friendly APIs to effortlessly complete an action. The ideal situation is for users and application developers to be unaware of padding resources.

Also, we can set a default logic for the padding resource so that we don't have to generate and verify the proofs for the padding resource logic.

PR #1897, opened 2025-02-20
refine cairo rm for V2

### LINKED ISSUE #1277: Adjust commitments, nullifiers, and roots in TX struct

XuyangSong:
address remaining issues with Specs V2 alignment following PR #1813

close #1277 #1646 #1815 #1858 #1880 #1653

Rebase once TRM V2 #1864 merges.


XuyangSong:
replaced by #1915

PR #1899, opened 2025-02-24
Fix Wrong From Noun for Lists

agureev:
Previously the from_noun for lists did not include the empty bitstring


agureev:
This repeats this:

https://github.com/anoma/anoma/pull/1832

but does it a bit better, so maybe worth keeping if 1832 is merged

PR #1910, opened 2025-02-26
Make Blocks Scryable

agureev:
Makes blocks be stored in the noun format with appropriate keyspace for further scry intergrations.


agureev:
@mariari where do you think we should place that Tx module?

PR #1915, opened 2025-02-28
cairo rm v31 integration

XuyangSong:
based on #1912

This PR can replace #1897 and #1813


apps/anoma_node/lib/node/transaction/backends.ex
```diff
@@ -445,6 +445,19 @@ defmodule Anoma.Node.Transaction.Backends do
 
       ciphertexts = tx |> CTransaction.get_cipher_texts() |> MapSet.new()
 
+      write_app_data =
+        tx.actions
+        |> Enum.flat_map(fn action ->
+          action.app_data
+          |> Enum.map(fn {key, value} ->
```
agureev: Blob storage is content-adressed, so we need to use the sha256 value as the key instead of the tag.

agureev (changes requested):
App data blob storage needs change.

apps/anoma_node/lib/node/transaction/backends.ex
XuyangSong: fixed

agureev (approved):

PR #1920, opened 2025-03-03
CairoRM Stdlib Additions

agureev:
Add CairoRM Datatypes to the standard library, alongside with the composition function jet


apps/anoma_lib/lib/nock/jets.ex
```diff
@@ -933,6 +933,45 @@ defmodule Nock.Jets do
...
+             res
+           end),
+         bin2_res <- Enum.map(bin2_list, &Noun.atom_integer_to_binary/1),
+         jason3_res <-
+           Enum.map(json3_list, fn x ->
+             {:ok, res} = Noun.Nounable.Jason.OrderedObject.from_noun(x)
+             res
+           end) do
+      {:ok,
+       CairoResource.Transaction.create_from_compliance_units(
```
lukaszcz: `create_from_compliance_units` returns either `{:ok, Transaction.t()}` or `{:error, term()}`. Shouldn't we check for error here and extract the Transaction instead of converting the whole result to a Nock Noun?

agureev: Solved by moving the function inside a `with`

lukaszcz:
I will also need `CairoResource.Transaction.prove_delta` in the Nock stdlib.

PR #1931, opened 2025-03-08
remove intent and otehr confusing protocols

### LINKED ISSUE #1870: Get Rid of Intent protocol, Intents, DumbIntents, etc.
customer: @mariari 
performer:
deadline:
estimated:
started:
actual:
completed:
confirmed:
dependencies:
- [ ]
supersedes:
-
---

Basic transparent transactions should be enough for debugging while the interface just complicates solver setups etc.

m1dnight:
Fixes #1870


mariari:
Really we should just rename the protocol, and call it a day, this does a little bit more than that

PR #1943, opened 2025-03-11
Make dependencies between apps explicit.

m1dnight:
This PR changes the dependencies in the separate apps. The dependencies are explicit now.

To figure out these dependencies I had to some convoluted stuff. I lost the bash script, but the gist is as follows.

1. Move all apps/*/lib/ into apps/*/lib/*/
2. `mix xref graph --only-direct --format dot` will not output graph nodes that have a prefix that uniquely identifies the app it belongs to
3. Generate graph 
4. Look at graph, figure out which app depends on which 
5. Explicitly define this in mix.exs in each app.

### DIFF
--- a/xref_graph.dot
+++ b/xref_graph.dot
@@ -1,0 +1,852 @@
+digraph "xref graph" {
+  "lib/anoma/cairo_resource/action.ex"
+  "lib/anoma/cairo_resource/action.ex" -> "lib/anoma/cairo_resource/compliance_instance.ex"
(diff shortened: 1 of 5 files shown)


mariari:
According to my tool, this PR is incomplete



is the current state as of base.

You are missing event_broker--->anoma_lib dependency

mariari:
Going further, here is the current state checking out at your PR



Namely event_broker being dependent upon anoma_lib

xref_graph.dot
```diff
@@ -0,0 +1,852 @@
+digraph "xref graph" {
```
mariari, line 1: Did you mean to push the dot file?

PR #1878, opened 2025-02-13
APIs for polynomial-term monad

rokopt:
These commits introduce the APIs (all of which are universal morphisms) for the polynomial-term monad introduced in #1859 (which as of now is still waiting for merge).  This is part of the work for #1846 ; I'm submitting this first because it's all internal implementation while the remainder will be external use, so I'd expect it to be more easily reviewable in separate chunks.

PR #1859, opened 2025-02-06
Initial polynomial-functor representation and specification of Nock

rokopt:
For #1844 .

Introduce:

- A generic type of open terms of any polynomial type, implemented as a functor (type constructor) and its free monad and initial algebra
- A representation of a simple form of non-dependent polynomial functors
- A representation of Nock terms (nouns) as a polynomial
- Conversions between the polynomial and Elixir-native (`Noun.t()`) formats
- Examples/tests for the above

This produces an open "Nock" term -- a Nock syntax with metavariables -- and as such will be used in a follow-on PR to support #1846 .  It also serves as the beginning of a type system including recursive and corecursive types (see #1845 and #1849 , which will also be built upon this in future PRs).

PR #1948, opened 2025-03-12
Introduce Extended Nock module and initial instructions

rokopt:
This PR builds on #1859 and #1878 to introduce the notion of Extended Nock:  an instruction set made of polynomial terms which include `:atom` and `:cell`, which directly embed Nock constructs, and also add further constructs for internal use.

It is layered as follows:  Elixir macros -> S-expressions -> Polynomial terms -> Elixir Nock.

The S-expression layer is effectively a macro layer of its own specifically for `:atom` and `:cell`, the Nock terms -- it allows polynomial terms to be represented using (proper) Elixir lists.  The Elixir-macros section isn't even used initially; it's just an empty module ignored by code coverage, but each Elixir macro that we do decide to make should be a thin veneer on the S-expressions, which themselves are a compact internal convenience for the real AST of extended Nock, which is the polynomial terms.

The point of the term architecture, in turn, is to allow a Nock base gradually to be extended upward, one construct or a small number of constructs at a time, towards becoming Anoma Level.  I propose we thereby build one seamless language, rather than a higher-level language which compiles to a lower-level one.

This PR introduces simple macros with readable names for Nock instructions plus a couple for core creation and invocation.  I think I should stop at this point and get things reviewed, but the next few will be targeted at replacing our hard-coded Hoon-built code, especially `resource-machine.hoon`, and thus will include multi-arm cores, layered libraries, and molds.

There is also a lot of refactoring that I want to do to make the `compile_algebra` more object-oriented and to allow multiple layers of the extended-Nock language, so that a higher layer can produce middle-layer constructs for further processing, rather than having to go directly to Nock terms when that might be repeating work from simpler macros, but again, I think this is probably a good checkpoint at which to stop this PR for review and discussion.  I will continue with both the new constructs and the refactoring in one or more further branches off this one.


rokopt:
I'm going to close this PR because the branch from which it requests to merge has been superseded by https://github.com/anoma/anoma/tree/terence/inductive-inductive-types .  The latter branch includes all the changes from the one in this PR as well as more enhancements and refactorings and responses to code-review comments.  Once there is a PR from that branch, it will do everything proposed in this PR (and more).

PR #1967, opened 2025-03-20
Zero Delta for TRM V2

agureev:
2 not 0 as jam of empty map not empty map


agureev:
Bad diff for our porposes, will include into a separate branch

PR #1976, opened 2025-03-25
Client tables

m1dnight:
This PR puts the creation of client storage table in the TAbles module with the other table infrastructure code.


agureev (approved):
Looks good to me, thank you for taking the time abstracting this!

agureev:
One thing I would request: please change the name of Anoma.Node.Tables or move it to the Node application

That makes me think it is located in the Node application

PR #1986, opened 2025-03-28
Merkle Tree Stores Indices

agureev:
Merkle tree now has a field with a map from commitments to indices. Needed for specs-compliance and improved user experience.

### DIFF
--- a/apps/anoma_lib/lib/examples/ecommitment_tree.ex
+++ b/apps/anoma_lib/lib/examples/ecommitment_tree.ex
@@ -120,7 +120,25 @@ defmodule Examples.ECommitmentTree do
       refute CommitmentTree.Proof.verify(spec, wrong, anchor, hash)
     end
 
+    for hash <- hashes do
+      prove = CommitmentTree.prove(ct, hash)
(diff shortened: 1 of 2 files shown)


apps/anoma_lib/lib/examples/ecommitment_tree.ex
```diff
@@ -120,9 +120,25 @@ defmodule Examples.ECommitmentTree do
...
 
+    for hash <- hashes do
+      prove = CommitmentTree.prove(ct, hash)
+      assert CommitmentTree.Proof.verify(spec, prove, anchor, hash)
+    end
+
     ct
   end
 
+  @spec babylon_mnesia_ct(CommitmentTree.Spec.t()) :: CommitmentTree.t()
```
XuyangSong: The function spec is incorrect. Please fix it before merging.

XuyangSong (approved):
lgtm

apps/anoma_lib/lib/examples/ecommitment_tree.ex
agureev: Done, thank you for noticing!

PR #1992, opened 2025-04-01
Protocol for Bool does not exist

m1dnight:
We had an implementation for `Noun.Nounable` for `Bool`, but that type does not exist. It is actually Atom. 

Migrating to 1.19 brought this to my attention. 

This PR fixes that.

### DIFF
--- a/apps/anoma_lib/lib/anoma/rm/transparent/proving_systems/rlps.ex
+++ b/apps/anoma_lib/lib/anoma/rm/transparent/proving_systems/rlps.ex
@@ -33,7 +33,7 @@ defmodule Anoma.RM.Transparent.ProvingSystem.RLPS.Instance do
 
   @spec from_noun(Noun.t()) :: {:ok, t()} | :error
   def from_noun([tag, flag, consumed, created | app_data]) do
-    with {:ok, bool} <- Noun.Nounable.Bool.from_noun(flag),
+    with {:ok, bool} <- Noun.Nounable.Atom.from_noun(flag),
          {:ok, list_consumed} <- Noun.Nounable.List.from_noun(consumed),
          {:ok, list_created} <- Noun.Nounable.List.from_noun(created),
          {:ok, list_app_data} <- Noun.Nounable.List.from_noun(app_data) do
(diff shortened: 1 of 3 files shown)


apps/anoma_lib/lib/anoma/rm/transparent/proving_systems/rlps.ex
```diff
@@ -58,7 +58,7 @@ defmodule Anoma.RM.Transparent.ProvingSystem.RLPS.Instance do
   def to_noun(instance) do
     [
       instance.tag,
-      Noun.Nounable.Bool.to_noun(instance.flag),
+      Noun.Nounable.Atom.to_noun(instance.flag),
```
agureev, line 61: If you are using a proper protocol it is probably best to just use Nounable itself without extra qualifiers.

```suggestion
      Noun.Nounable.to_noun(instance.flag),
```

apps/anoma_lib/lib/noun/nounable.ex
```diff
@@ -81,9 +60,21 @@ defimpl Nounable, for: BitString do
 end
 
 defimpl Nounable, for: Atom do
+  import Noun
+
   def to_noun(true), do: 0
   def to_noun(false), do: 1
   def to_noun(atom), do: Atom.to_string(atom)
```
agureev, line 67: Delete this line, let it fail on a general atom. Not the ideal option but honestly probably the best we can do, i.e. assuming we never actually compute non-boolean atoms (maybe add in the `nil` compilation to 0 or something as Ray suggested).

```suggestion
```

agureev (changes requested):
So the semantics of the `from_noun` and `to_noun` for the same thing should be that they are inverses.

Pushing the boolean logic to atom as you did there does not work for the evident reason that currently `:atom |> to_noun |> from_noun` currently produces `:error`

Now the problem also is that if you introduce the evident additional match for the `from_noun` that checks whether the inut is an atom, then transforms to binary and takes it to atom, it is also a bit stupid as it then can produce extremely stupid atoms. This can be mitigated but honestly why bother.

The better approach is to assume as Ray suggested that we only transform a very limited number of atoms, specifically booleans.

I think this is the nicer thing to do.

Either that, or make the inverse approach but it's going to be pretty nasty to make properly.

PR #1997, opened 2025-04-03
Livebook for eventbroker load testing

m1dnight:
This PR adds a Livebook to run some rudimentary load tests on the eventbroker and includes pretty charts.


mariari:
Since this is mostly docs I accept, it's not ideal, as we can do it in the same image, but this is fine.

PR #1998, opened 2025-04-03
Sharding Implementation

AHartNtkn:
Implements shards as a new backend. Core functionality is implemented in the transaction directory within anoma_node. The main files to look at are.
- shards.ex, which contains the core implementation of shards as a genserver.
- backend.ex, which was modified with a new shard_storage backend option that integrates with shards.
- ordering.ex, which was modified to perform the watermark tracking operations necessary for shards.

Still significant but less substantial are the files
- node/supervisor.ex
- transaction/supervisor.ex
- shard_supervisor.ex
- shard_router.ex
which add the ability to provide a simple schema at node startup which is used to spawn the initial shards and associate them with keys in the distributed KV store.

Additional examples were added. Specifically
- e_shard.ex, which tests the shards themselves
- e_shard_supervisor.ex, which tests shard spawning at node startup
- e_shard_backend.ex, which tests actually executing transactions using the shard_storage backend

### DIFF
--- a/apps/anoma_lib/lib/examples/enock.ex
+++ b/apps/anoma_lib/lib/examples/enock.ex
@@ -2765,4 +2765,154 @@ defmodule Examples.ENock do
 
     code
   end
+
+  @spec a_int() :: integer()
+  @spec b_int() :: integer()
+  @spec c_int() :: integer()
+  def a_int(), do: Noun.atom_binary_to_integer("a")
+  def b_int(), do: Noun.atom_binary_to_integer("b")
+  def c_int(), do: Noun.atom_binary_to_integer("c")
+
+  @spec abc_list() :: Noun.t()
+  def abc_list() do
+    [[0 | a_int()], [0 | b_int()], [1 | c_int()] | 0]
+  end
+
+  ####################################################################
+  ##                     Shard Backend Examples                     ##
+  ####################################################################
+
+  # Write Program Generator (writes a constant value to a key)
+  def write_code_gen(key_int, value) do
+    reservations = [[1 | key_int] | 0]
+    writes_logic = [1 | [[key_int | value] | 0]]
+    [[0 | 3] | [reservations | [writes_logic | [0 | 0]]]]
+  end
+
+  @doc """
+  Nock program for reading "a", "b", adding, and writing sum to "c". Used in EShardBackend tests.
+  """
+  @spec read_ab_write_c_sum() :: Noun.t()
+  def read_ab_write_c_sum() do
+    key_a = a_int()
+    key_b = b_int()
+    key_c = c_int()
+
+    reservations = abc_list()
+
+    scryA = [12 | [[1 | 0] | [1 | key_a]]]
+    scryB = [12 | [[1 | 0] | [1 | key_b]]]
+    addFormula = [4 | [4 | [4 | [4 | [0 | 2]]]]]
+    sumFormula = [7 | [[scryA | scryB] | addFormula]]
+    writes_logic = [[[1 | key_c] | sumFormula] | [1 | 0]]
+
+    [[0 | 3] | [reservations | [writes_logic | [0 | 0]]]]
+  end
+
+  @doc """
+  Nock program for copying value from "a" to "b". Used in EShardBackend tests.
+  """
+  @spec copy_a_to_b() :: Noun.t()
+  def copy_a_to_b() do
+    key_a = a_int()
+    key_b = b_int()
+
+    reservations = [[0 | key_a], [1 | key_b] | 0]
+    scryA = [12 | [[1 | 0] | [1 | key_a]]]
+    writes_logic = [[[1 | key_b] | scryA] | [1 | 0]]
+
+    [[0 | 3] | [reservations | [writes_logic | [0 | 0]]]]
+  end
+
+  @doc """
+  Nock program that reserves write on 'a', then crashes. Used in EShardBackend tests.
+  """
+  @spec crash_after_reserve_a() :: Noun.t()
+  def crash_after_reserve_a() do
+    reservations = [[1 | a_int()] | 0]
+    # Invalid Nock [0 | [0 | 0]] for the writes stage causes a crash
+    [[0 | 3] | [reservations | [0 | [0 | 0]]]]
+  end
+
+  @doc """
+  I test a Nock program for writing 3 to "a".
+
+  I verify stage 1 (reservation/code split) and stage 2 (execution).
+  """
+  @spec test_nock_program_tx1() :: :ok
+  def test_nock_program_tx1() do
+    # Setup
+    dummy_tx_id = "test_tx_id_1"
+    tx_code = write_code_gen(a_int(), 3)
+    expected_reservations = [[1 | a_int()] | 0]
+    expected_writes = [[a_int() | 3] | 0]
+
+    # Stage 1: Execute the core using the backend's formula to extract reservations
+    {:ok, stage1_result} = Nock.nock(tx_code, [9, 2, 0 | 1], %Nock{})
+
+    # Extract components from stage 1 result
+    [reservations | stage2_code] = stage1_result
+    assert reservations == expected_reservations
+
+    # Stage 2: Execute using the exact sequence from Backends.vm_execute_stage2
+    env_tx1 = %Nock{}
+
+    # Step 1: Apply formula [10, [6, 1 | id], 0 | 1] - this inserts the tx_id into the code
+    {:ok, ordered_tx1} =
+      Nock.nock(stage2_code, [10, [6, 1 | dummy_tx_id], 0 | 1], env_tx1)
+
+    # Step 2: Apply formula [9, 2, 0 | 1] - this executes the transaction
+    {:ok, result1} = Nock.nock(ordered_tx1, [9, 2, 0 | 1], env_tx1)
+
+    # This is the expected format that backends.ex receives from vm_execute_stage2
+    assert result1 == expected_writes
+
+    :ok
+  end
+
+  @doc """
+  I test a Nock program for the sequence (read a, read b, add a and b, write result to c).
+
+  I verify stage 1 (reservation/code split) and stage 2 (execution with mock scry).
+  """
+  @spec test_nock_program_tx3() :: :ok
+  def test_nock_program_tx3() do
+    # Setup
+    dummy_tx_id = "test_tx_id_3"
+    tx_code = read_ab_write_c_sum()
+    expected_reservations = abc_list()
+    expected_writes = [[c_int() | 7] | 0]
+
+    # Stage 1: Execute the core using the backend's formula
+    {:ok, [reservations | stage2_code]} =
+      Nock.nock(tx_code, [9, 2, 0 | 1], %Nock{})
+
+    # Extract the components
+    assert reservations == expected_reservations
+
+    # Create a mock scry function that simulates reading values from storage
+    a_val = a_int()
+    b_val = b_int()
+
+    mock_scry_tx3 = fn
+      ^a_val -> {:ok, 3}
+      ^b_val -> {:ok, 4}
+      _ -> :error
+    end
+
+    env_tx3 = %Nock{scry_function: mock_scry_tx3}
+
+    # Stage 2: Execute using the exact sequence from Backends.vm_execute_stage2
+    # Step 1: Apply formula [10, [6, 1 | id], 0 | 1] - inserts tx_id
+    {:ok, ordered_tx3} =
+      Nock.nock(stage2_code, [10, [6, 1 | dummy_tx_id], 0 | 1], env_tx3)
+
+    # Step 2: Apply formula [9, 2, 0 | 1] - executes transaction
+    {:ok, result3} = Nock.nock(ordered_tx3, [9, 2, 0 | 1], env_tx3)
+
+    # Expected sum and result format
+    assert result3 == expected_writes
+
+    :ok
+  end
 end
--- a/apps/anoma_node/lib/examples/e_shard.ex
+++ b/apps/anoma_node/lib/examples/e_shard.ex
@@ -1,0 +1,952 @@
+defmodule Anoma.Node.Examples.EShard do
+  @moduledoc """
+  I contain examples on how to interact with the Shard module.
+  """
+
+  alias Anoma.Node.Registry
+  alias Anoma.Node.Transaction.Shard
+  alias Anoma.Node.Examples.ENode
+
+  import ExUnit.Assertions
+
+  @doc """
+  I start a node with shards "a", "b", "c", initializing "a" and "c"
+  with specific values, and verify the initial state.
+  """
+  @spec spawn_node_with_initial_state() :: {:ok, ENode.t()}
+  def spawn_node_with_initial_state() do
+    # Generate a unique node_id for test isolation
+    unique_suffix = System.unique_integer([:positive, :monotonic])
+    node_id = "shard_test_node_#{unique_suffix}"
+
+    # Define the schema: keys "a", "b", "c". "a" and "c" have initial values.
+    schema = [{"a", 5}, "b", {"c", 15}]
+    shard_config = [strategy: :one_per_key, schema: schema]
+    opts = [node_id: node_id, transaction: [shards: shard_config]]
+
+    # Start the node
+    enode = ENode.start_node(opts)
+    assert %ENode{node_id: ^node_id} = enode
+
+    # Get shard PIDs
+    pid_a = Registry.whereis(node_id, Shard, :a)
+    pid_b = Registry.whereis(node_id, Shard, :b)
+    pid_c = Registry.whereis(node_id, Shard, :c)
+
+    assert is_pid(pid_a), "Shard 'a' PID not found."
+    assert is_pid(pid_b), "Shard 'b' PID not found."
+    assert is_pid(pid_c), "Shard 'c' PID not found."
+
+    # Get initial states
+    state_a = Shard.debug_get_state(pid_a)
+    state_b = Shard.debug_get_state(pid_b)
+    state_c = Shard.debug_get_state(pid_c)
+
+    # Verify initial states at height 0
+    assert Map.get(state_a.kv, "a", %{})[0].value == 5,
+           "Shard 'a' initial value mismatch at height 0"
+
+    assert Map.get(state_c.kv, "c", %{})[0].value == 15,
+           "Shard 'c' initial value mismatch at height 0"
+
+    # Shard "b" should have no entry for key "b" at height 0
+    b_key_map = Map.get(state_b.kv, "b", %{})
+
+    refute Map.has_key?(b_key_map, 0),
+           "Shard 'b' should not have an initial value at height 0"
+
+    {:ok, enode}
+  end
+
+  @doc """
+  I start a Shard with a predefined initial state and verify that
+  reading the initial state (at height 1) returns the correct values.
+  """
+  @spec start_and_initial_state() :: {:ok, ENode.t()}
+  def start_and_initial_state() do
+    {:ok, enode} = spawn_node_with_initial_state()
+    %ENode{node_id: node_id} = enode
+
+    shard_a_via = Registry.via(node_id, Shard, :a)
+    shard_b_via = Registry.via(node_id, Shard, :b)
+    shard_c_via = Registry.via(node_id, Shard, :c)
+
+    # --- Simulate Watermark Advancement prior to acquiring reservations ---
+    Shard.advance_write_watermark(shard_c_via, "c", 5)
+
+    # --- Acquire Reservations First ---
+    assert :ok == Shard.reserve(shard_a_via, "a", 1, :read)
+    assert :ok == Shard.reserve(shard_b_via, "b", 5, :read)
+    assert :ok == Shard.reserve(shard_c_via, "c", 4, :read)
+
+    # --- Simulate Watermark Advancement after acquiring reservations ---
+    Shard.advance_write_watermark(shard_a_via, "a", 1)
+    Shard.advance_write_watermark(shard_b_via, "b", 10)
+
+    # --- Test Reads (Now that watermarks allow immediate resolution) ---
+
+    # Test key "a"
+    assert Shard.read(shard_a_via, "a", 1) == {:ok, 5}
+
+    # Test key "b"
+    assert Shard.read(shard_b_via, "b", 5) == :absent
+
+    # Test key "c"
+    assert Shard.read(shard_c_via, "c", 4) == {:ok, 15}
+
+    {:ok, enode}
+  end
+
+  @doc """
+  I test a scenario where a read is requested before the watermark allows,
+  then the watermark advances, and the read completes.
+  """
+  @spec queued_read() :: {:ok, ENode.t()}
+  def queued_read() do
+    {:ok, enode} = spawn_node_with_initial_state()
+    %ENode{node_id: node_id} = enode
+
+    shard_via = Registry.via(node_id, Shard, :a)
+
+    key = "a"
+    height = 7
+
+    # 1. Acquire Reservation
+    assert :ok == Shard.reserve(shard_via, key, height, :read)
+
+    # 2. Start the read in a separate task (it will block)
+    read_task =
+      Task.async(fn ->
+        Shard.read(shard_via, key, height)
+      end)
+
+    # 2.5 Verify the read task is initially blocked
+    assert Task.yield(read_task, 100) == nil,
+           "Read task should be blocked before watermark advances"
+
+    # 3. Advance the watermark *after* the read call is blocked
+    Shard.advance_write_watermark(shard_via, key, height + 1)
+
+    # 4. Wait for the result from the task (should unblock now)
+    result = Task.await(read_task, 1000)
+
+    # 5. Assert the result
+    assert result == {:ok, 5}
+
+    {:ok, enode}
+  end
+
+  @doc """
+  I test a variation of queued read where a write reservation is acquired and
+  a write is performed *after* the read is queued but *before* the read resolves,
+  affecting the read's outcome.
+  """
+  @spec queued_read_with_intermediate_write() :: {:ok, ENode.t()}
+  def queued_read_with_intermediate_write() do
+    {:ok, enode} = spawn_node_with_initial_state()
+    %ENode{node_id: node_id} = enode
+
+    shard_via = Registry.via(node_id, Shard, :a)
+
+    key = "a"
+    read_height = 7
+    # Height for the intermediate write
+    write_height = 5
+    # Value for the intermediate write
+    write_value = 10
+
+    # 1. Acquire Read Reservation for the future read
+    assert :ok == Shard.reserve(shard_via, key, read_height, :read)
+
+    # 2. Start the read in a separate task (it will block)
+    read_task =
+      Task.async(fn ->
+        Shard.read(shard_via, key, read_height)
+      end)
+
+    # 2.5 Verify the read task is initially blocked
+    assert Task.yield(read_task, 100) == nil,
+           "Read task should be blocked initially"
+
+    # 3. Acquire Write Reservation for an intermediate height BEFORE watermark advances
+    assert :ok == Shard.reserve(shard_via, key, write_height, :write)
+
+    # 4. Advance the watermark AFTER the read call is blocked, enabling read resolution
+    # WM >= read_height
+    Shard.advance_write_watermark(shard_via, key, read_height + 1)
+
+    # 5. Perform the Write AFTER watermark advanced but potentially before read task resumes
+    # This write should be visible to the resolving read at height 7.
+    assert :ok ==
+             Shard.write(
+               shard_via,
+               key,
+               write_value,
+               write_height
+             )
+
+    # 6. Await the result from the read task (should unblock due to WM)
+    result = Task.await(read_task, 1000)
+
+    # 7. Assert the result
+    # The read at height 7 should resolve to the latest write strictly below 7.
+    # The write at height 5 (value 10) occurred before the read resolved.
+    # The initial value at 0 is 5.
+    # Therefore, the latest write < 7 is the one at height 5.
+    assert result == {:ok, write_value}
+
+    {:ok, enode}
+  end
+
+  @doc """
+  I test a scenario where a read is requested, but the watermark never
+  advances, causing the read to time out.
+  """
+  @spec read_timeout() :: {:ok, ENode.t()}
+  def read_timeout() do
+    {:ok, enode} = spawn_node_with_initial_state()
+    %ENode{node_id: node_id} = enode
+
+    shard_via = Registry.via(node_id, Shard, :a)
+
+    key = "a"
+    height = 5
+
+    # 1. Acquire Reservation
+    assert :ok == Shard.reserve(shard_via, key, height, :read)
+
+    # 2. Start the read in a separate task (it will block)
+    read_task =
+      Task.async(fn ->
+        # Note: The GenServer.call within Shard.read uses :infinity,
+        # so this task itself won't timeout internally. The timeout
+        # comes from Task.await below.
+        Shard.read(shard_via, key, height)
+      end)
+
+    # 2.5 Verify the read task is initially blocked
+    assert Task.yield(read_task, 100) == nil,
+           "Read task should be blocked before trying to await with timeout"
+
+    # 3. DO NOT advance the watermark
+
+    # 4. Await the result with a short timeout
+    # We expect this to exit with reason :timeout
+    try do
+      Task.await(read_task, 100)
+      # If await succeeds, the test fails
+      flunk("Task.await should have timed out and exited, but it returned.")
+    catch
+      :exit, reason ->
+        assert reason == :timeout or match?({:timeout, _}, reason)
+    end
+
+    if Process.alive?(read_task.pid),
+      do: Task.shutdown(read_task, :brutal_kill)
+
+    {:ok, enode}
+  end
+
+  @doc """
+  I test a scenario with two pending reads at different heights.
+  An intermediate watermark advance unblocks only the lower-height read,
+  while the higher-height read eventually times out.
+  """
+  @spec partial_read_unblocking_with_timeout() :: {:ok, ENode.t()}
+  def partial_read_unblocking_with_timeout() do
+    {:ok, enode} = spawn_node_with_initial_state()
+    %ENode{node_id: node_id} = enode
+
+    shard_via = Registry.via(node_id, Shard, :a)
+
+    key = "a"
+
+    read_height_ok = 5
+    read_height_timeout = 15
+    watermark_height = 10
+
+    # 1. Acquire Reservations
+    assert :ok == Shard.reserve(shard_via, key, read_height_ok, :read)
+
+    assert :ok == Shard.reserve(shard_via, key, read_height_timeout, :read)
+
+    # 2. Start Read Tasks (both will block initially)
+    read_task_ok =
+      Task.async(fn ->
+        Shard.read(shard_via, key, read_height_ok)
+      end)
+
+    read_task_timeout =
+      Task.async(fn ->
+        Shard.read(shard_via, key, read_height_timeout)
+      end)
+
+    # 2.5 Verify both tasks are initially blocked
+    assert Task.yield(read_task_ok, 100) == nil,
+           "Read task (ok) should be blocked initially"
+
+    assert Task.yield(read_task_timeout, 100) == nil,
+           "Read task (timeout) should be blocked initially"
+
+    # 3. Advance Watermark partially (enough for height 5, not for 15)
+    Shard.advance_write_watermark(shard_via, key, watermark_height)
+
+    # 4. Await the read that should succeed
+    result_ok = Task.await(read_task_ok, 1000)
+
+    # Read at height 5 resolves based on latest write < 5, which is initial value 5 at height 0
+    assert result_ok == {:ok, 5}
+
+    # 5. Await the read that should time out
+    try do
+      Task.await(read_task_timeout, 100)
+
+      flunk(
+        "Task for height #{read_height_timeout} should have timed out, but it returned."
+      )
+    catch
+      :exit, reason ->
+        assert reason == :timeout or match?({:timeout, _}, reason)
+    end
+
+    if Process.alive?(read_task_timeout.pid),
+      do: Task.shutdown(read_task_timeout, :brutal_kill)
+
+    {:ok, enode}
+  end
+
+  @doc """
+  I test a more complex scenario involving multiple writes, reads, and
+  write watermark advancements.
+  """
+  @spec complex_write_and_read_scenario() :: {:ok, ENode.t()}
+  def complex_write_and_read_scenario() do
+    {:ok, enode} = spawn_node_with_initial_state()
+    %ENode{node_id: node_id} = enode
+
+    shard_via = Registry.via(node_id, Shard, :a)
+
+    key = "a"
+
+    # --- Acquire Write Reservations ---
+    assert :ok == Shard.reserve(shard_via, key, 5, :write)
+    assert :ok == Shard.reserve(shard_via, key, 6, :write)
+    assert :ok == Shard.reserve(shard_via, key, 10, :write)
+
+    # --- Acquire Read Reservations ---
+    assert :ok == Shard.reserve(shard_via, key, 1, :read)
+    assert :ok == Shard.reserve(shard_via, key, 4, :read)
+    assert :ok == Shard.reserve(shard_via, key, 5, :read)
+    assert :ok == Shard.reserve(shard_via, key, 6, :read)
+    assert :ok == Shard.reserve(shard_via, key, 7, :read)
+    assert :ok == Shard.reserve(shard_via, key, 9, :read)
+    assert :ok == Shard.reserve(shard_via, key, 10, :read)
+    assert :ok == Shard.reserve(shard_via, key, 11, :read)
+
+    # --- Perform Writes ---
+    assert :ok == Shard.write(shard_via, key, 7, 5)
+    assert :ok == Shard.write(shard_via, key, 2, 6)
+    assert :ok == Shard.write(shard_via, key, 8, 10)
+
+    # --- Simulate Watermark Advancements ---
+    # Reads 1, 4, 5 depend on initial state (implied WM >= 1)
+    Shard.advance_write_watermark(shard_via, key, 1)
+
+    # Read 6 needs to see write at 5
+    Shard.advance_write_watermark(shard_via, key, 6)
+
+    # Reads 7, 9, 10 need to see write at 6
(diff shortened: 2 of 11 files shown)


apps/anoma_node/lib/examples/e_shard_backend.ex
```diff
@@ -0,0 +1,738 @@
...
+    key_b_int = Noun.atom_binary_to_integer("b")
+    key_c_int = Noun.atom_binary_to_integer("c")
+    dummy_tx_id = "test_tx_id_3"
+
+    # Tx3: Read "a", Read "b", Add, Write result to "c"
+    tx3_reservations = [
+      [0 | key_a_int],
+      [0 | key_b_int]
+      | [1 | key_c_int]
+    ]
```
mariari: All of these should be moved to examples like so

```elixir

defmodule Examples.ENock do

  ####################################################################
  ##                           Noun fun                             ##
  ####################################################################

  @spec a_int() :: integer()
  @spec b_int() :: integer()
  @spec c_int() :: integer()
  def a_int(), do: Noun.atom_binary_to_integer("a")
  def b_int(), do: Noun.atom_binary_to_integer("b")
  def c_int(), do: Noun.atom_binary_to_integer("c")


  @spec abc_list() :: Noun.t()
  def abc_list() do
    [[0 | a_int()], [0 | b_int()] | [1 | c_int()]]
  end
```

Now the code above can be

```elixir
@spec test_nock_program_tx3() :: :ok
  def test_nock_program_tx3() do
    dummy_tx_id = "test_tx_id_3"
    tx3_reservation = Example.ENock.abc_list()
    ...
```

I notice these variables in particular are used in a lot of places and can be reused instead of copying

apps/anoma_node/lib/examples/e_shard_backend.ex
```diff
@@ -0,0 +1,738 @@
...
+
+    # 1. Setup Node with Shards
+    # No initial values, which won't matter.
+    schema = ["a", "b", "c"]
+    shard_config = [strategy: :one_per_key, schema: schema]
+    opts = [node_id: node_id, transaction: [shards: shard_config]]
+    enode = ENode.start_node(opts)
+    assert %ENode{node_id: ^node_id} = enode
+    # Allow time for supervisors and shards to start
+    Process.sleep(200)
```
mariari: We should not be using sleep in examples, this means we likely aren't emitting an event or having something that should be synchronous but isn't

apps/anoma_node/lib/examples/e_shard_backend.ex
```diff
@@ -0,0 +1,738 @@
...
+    tx1_writes = [key_a_int | 3]
+    tx1_code = [[0 | 3] | [tx1_reservations | [[1 | tx1_writes] | [0 | 0]]]]
+
+    # Tx2: Write 4 to "b"
+    tx2_reservations = [1 | key_b_int]
+    tx2_writes = [key_b_int | 4]
+    tx2_code = [[0 | 3] | [tx2_reservations | [[1 | tx2_writes] | [0 | 0]]]]
+
+    # Tx3: Read "a", Read "b", Add, Write result to "c"
+    tx3_reservations = [[0 | key_a_int], [0 | key_b_int] | [1 | key_c_int]]
```
mariari: Example reuasge... all the tx_* can be set as examples in enock and brought over

apps/anoma_node/lib/node/transaction/shard_router.ex
```diff
@@ -0,0 +1,121 @@
...
+
+  I make a synchronous call to my GenServer process to perform the lookup
+  in the ETS table (`:shard_key_map`).
+  I return `{:ok, shard_label}` if the key is found, otherwise `:error`.
+  """
+  @spec get_shard_label(key :: key_t()) ::
+          {:ok, atom()} | :error
+  def get_shard_label(key) when is_binary(key) do
+    GenServer.call(__MODULE__, {:get_shard_label, key})
+  end
```
mariari: This should be multihomed, this should take a `node_id` as it's first argument and find the shard there

apps/anoma_node/lib/examples/e_shard_supervisor.ex
```diff
@@ -0,0 +1,78 @@
...
+    assert GenServer.call(via_router, {:get_shard_label, "b"}) ==
+             {:ok, :b},
+           "Router lookup for 'b' failed"
+
+    assert GenServer.call(via_router, {:get_shard_label, "c"}) ==
+             {:ok, :c},
+           "Router lookup for 'c' failed"
+
+    assert GenServer.call(via_router, {:get_shard_label, "d"}) == :error,
+           "Router lookup for unknown key 'd' should return :error"
```
mariari: We should just use `ShardRouter.get_shard_label` instead, we should not be doing raw `GenServer.call`'s here

apps/anoma_node/lib/examples/e_shard.ex
```diff
@@ -0,0 +1,865 @@
...
+    # Test key "a"
+    assert Shard.read(shard_via, "a", 0) == {:ok, 5}
+
+    # Test key "b"
+    assert Shard.read(shard_via, "b", 5) == :absent
+
+    # Test key "c"
+    assert Shard.read(shard_via, "c", 4) == {:ok, 15}
+
+    :ok
```
mariari: We should be returning the shard here, as it could bootstrap other examples as you reuse it in examples to grow more and more complex examples.

Think about the multi-stage setup, instead of doing it in 1 "test" instead do it in a series of examples you can reuse that give you the shard at certain states to assert properties about

apps/anoma_node/lib/node/transaction/shard.ex
```diff
@@ -0,0 +1,767 @@
...
+      watermarks: watermarks,
+      pending_reads: %{}
+    }
+
+    {:ok, state}
+  end
+
+  ############################################################
+  #                   Genserver Callbacks                    #
+  ############################################################
```
mariari: In general we should do no processing in the `handle_call/cast/info` instead we should dispatch it to functions like `handle_dispatch` that does the logic for the specific cases you care about.

This is mainly so dialyzer can help typecheck the bodies and we can add type signatures to all the functions here

apps/anoma_node/lib/node/transaction/shard.ex
```diff
@@ -0,0 +1,767 @@
...
+
+  # --- Reservation Handling ---
+  @impl true
+  def handle_call({:reserve, key, height, type}, _from, state) do
```
mariari, line 296: This function is far too big and not abstracted I believe we can handle it like this

```elixir
def handle_reserve(key, height, type, state) when type in [:write, :read_write] do
    key_watermarks = Map.get(state.watermarks, key, %{read: -1, write: -1})
    with :ok <- reserve_under_watermark_check(key_watermarks),
         {:ok, details} <- details(type, state.kv, key, height) do
      ...
    else
      error -> error
    end
end
```

depending on the specific abstractions `handle_reserve` can remove the `when` clause. 

- The `reserve_under_watermark_check` can do the appropriate checking in the head as pattern matching and return the resposne `{:error, :reason}` which gets returned.

- The `details` hides away basically hiding away the `details_after_write` `details_after_read` and the tuple return you've done with a basic enum that if it fails goes to the error case right away which would be sufficient

Overall I think this function can be cut down and become a lot more readable.

a big benefit of doing it like this is that we can move all casing on `:read`, `:read_write` and `:write` into the heads of functions instead of bodies making it easier to add new cases later as we don't have to read the bodies of the functions to figure out where we should add the case.

apps/anoma_node/lib/node/transaction/shard.ex
```diff
@@ -0,0 +1,767 @@
...
+
+        # 3. Determine essential heights to keep below each active read reservation
+        essential_below_reservations =
+          Enum.reduce(read_reservation_heights, MapSet.new(), fn h_rr,
+                                                                 acc_set ->
+            MapSet.union(
+              acc_set,
+              find_essential_heights_below(h_rr, key_height_map)
+            )
+          end)
```
mariari:
```suggestion
          read_reservation_heights
          |> Enum.map(&find_essential_heights_below(&1, key_height_map))
          |> Enum.reduce(MapSet.union/2)
```

basically we can turn this into a clean map reduce, making the reduce properly associative and the map explicit, this makes the code easier to read and reason about

mariari (changes requested):
I have not done a proper review yet but:

1. The examples aren't really examples, we should make them examples and reuse components
2. we should never use sleep
3. We should abstract genserver bodies into their own functions and make sure functions are abstracted from there as 100 lines where variables can live over the whole scope is a bit hard to reason about
4. we should multi-home everything

apps/anoma_node/lib/node/transaction/shard_supervisor.ex
```diff
@@ -0,0 +1,201 @@
+defmodule Anoma.Node.Transaction.ShardSupervisor do
```
agureev, line 1: Is this worth having as an extra process?

Why not have the ShardRouter combine these logics?

apps/anoma_node/lib/node/transaction/shard_supervisor.ex
```diff
@@ -0,0 +1,201 @@
...
+  @type shard_args_t :: [
+          id: key_t(),
+          initial_kv: %{key_t() => initial_value_t()}
+        ]
+
+  ############################################################
+  #                       Constants                          #
+  ############################################################
+
+  @ets_table_name :shard_key_map
```
agureev: Should be parameterized by the node id for multihoming.

apps/anoma_node/lib/node/transaction/shard_supervisor.ex
```diff
@@ -0,0 +1,201 @@
...
+          {[], %{}}
+
+        {_invalid_strategy, _} ->
+          {[], %{}}
+      end
+
+    # Create and populate ETS table if shards were generated
+    if map_size(key_to_name_map) > 0 do
+      ets_table =
+        :ets.new(@ets_table_name, [
```
agureev: I think we support Mnesia first an foremostly for actual backends. Probably want to keep it that way here as well. Thoughts @juped @m1dnight?

apps/anoma_node/lib/node/transaction/shard_router.ex
```diff
@@ -0,0 +1,122 @@
...
+
+  @typedoc "I am the arguments passed to start_link. Requires node_id."
+  @type args_t :: [node_id: String.t()]
+
+  ############################################################
+  #                       Constants                          #
+  ############################################################
+
+  # The ETS table is created and owned by the ShardSupervisor, but we read from it.
+  @ets_table_name :shard_key_map
```
agureev: Same comment as for supervisor.

agureev: Also if you keep both the supervisor and router you can abstract constants as functions. See @jet_registry in Nock.Jets.Mugs

apps/anoma_node/lib/node/transaction/shard.ex
```diff
@@ -0,0 +1,732 @@
...
+            {:reply, {:ok, %{read: new_read_ref, write: new_write_ref}},
+             state}
+          end
+        end
+    end
+  end
+
+  # --- Write Handling ---
+  @impl true
+  def handle_call({:write, key, value, height, write_ref}, _from, state) do
```
agureev: Are you planning on introducing table-backing for the results?

apps/anoma_node/lib/node/transaction/shard.ex
```diff
@@ -0,0 +1,732 @@
...
+  and initialize its KV state based on `initial_kv` options.
+  """
+  @spec start_link(keyword()) :: GenServer.on_start()
+  def start_link(opts) do
+    # id: shard_id, node_id: node_id, initial_kv: %{key => val}
+    GenServer.start_link(__MODULE__, opts)
+  end
+
+  @doc """
+  I am the lock function for the Shard module.
```
agureev: Can you elaborate in the documentation what locks mean practically maybe?

apps/anoma_node/lib/examples/e_shard.ex
```diff
@@ -0,0 +1,746 @@
...
+
+    # 2. Start the read in a separate task (it will block)
+    read_task =
+      Task.async(fn ->
+        Shard.read(shard_via, key, height, read_ref)
+      end)
+
+    # Give the task a tiny moment to start and make the call
+    Process.sleep(50)
+
```
agureev: And here the expected semantics is that the task is not done yet, is that correct?

apps/anoma_node/lib/examples/e_shard.ex
```diff
@@ -0,0 +1,746 @@
...
+  """
+  @spec test_queued_read_with_intermediate_write() :: ENode.t()
+  def test_queued_read_with_intermediate_write() do
+    enode = ENode.start_node()
+    node_id = enode.node_id
+    shard_id = :test_shard_queued_write
+    shard_via = Registry.via(node_id, Shard, shard_id)
+
+    key = "a"
+    initial_value = 5
```
agureev: Maybe use some other type than integer for the values? Makes it a bit hard to read and differentiate values with heights.

apps/anoma_node/lib/node/transaction/shard.ex
```diff
@@ -0,0 +1,732 @@
...
+          read_lock_ref: reference() | nil,
+          # The ref if write-locked, nil otherwise
+          write_lock_ref: reference() | nil
+        }
+
+  ############################################################
+  #                         State                            #
+  ############################################################
+
+  typedstruct enforce: true do
```
agureev, line 78: General question: given that we do not have a notion of "current height" in the shards (if I understand the semantics correctly), what do read-only transactions ask to read at?

agureev: I understand that we can probably do it through Ordering but how is that handled in the specs?

apps/anoma_node/lib/node/transaction/shard_router.ex
```diff
@@ -0,0 +1,122 @@
+defmodule Anoma.Node.Transaction.ShardRouter do
```
agureev: Alternatively to my original take re merging Supervisor and Router: I think the router can be moved to Ordering maybe? Either way when transactions e.g. read they should ask to read at their id first, so they will first route the request to Ordering.

apps/anoma_node/lib/node/transaction/backends.ex
```diff
@@ -36,6 +37,7 @@ defmodule Anoma.Node.Transaction.Backends do
           | :debug_bloblike
           | :transparent_resource
           | :cairo_resource
+          | :shard_storage
```
agureev: General question: Ideally we would like for shards to be something that replaces current Storage, is that correct? i.e. we don't want it to be the backend in the sense of Backends module, which just lists basically some forms of allowed RM protocols.

What is blocking currently that sort of intergration? Keyspace reveal at the Mempool level?

apps/anoma_node/lib/node/transaction/shard_router.ex
```diff
@@ -115,8 +105,17 @@ defmodule Anoma.Node.Transaction.ShardRouter do
...
-        # Reply with error if ETS fails unexpectedly
         {:reply, :error, state}
     end
   end
+
+  @impl true
+  @doc """
+  I clean up the ETS table when I (the router) terminate.
+  """
+  @spec terminate(reason :: any(), state :: :no_state) :: :ok
```
agureev: I don't think it is the best idea to have this sort of cleanup, we should instead probably backup the table with shard names.

apps/anoma_node/lib/examples/e_shard_backend.ex
```diff
@@ -0,0 +1,738 @@
...
+  defp log_shard_states(node_id, stage_name) do
+    pid_a = Registry.whereis(node_id, Shard, :a)
+    pid_b = Registry.whereis(node_id, Shard, :b)
+    pid_c = Registry.whereis(node_id, Shard, :c)
+
+    state_a = if is_pid(pid_a), do: :sys.get_state(pid_a), else: :not_found
+    state_b = if is_pid(pid_b), do: :sys.get_state(pid_b), else: :not_found
+    state_c = if is_pid(pid_c), do: :sys.get_state(pid_c), else: :not_found
+
+    Logger.debug("-- Shard States [#{stage_name}] --")
```
agureev: Why not use the EventBroker here?

apps/anoma_node/lib/node/transaction/backends.ex
```diff
@@ -143,52 +145,196 @@ defmodule Anoma.Node.Transaction.Backends do
...
+          backend()
+        ) :: :ok
+  defp execute_shard_storage(node_id, tx_code, id, backend) do
+    # For shard storage, first extract reservations
+    case vm_execute_stage1(tx_code) do
+      {:ok, {stage_2_tx, reservations}} ->
+        case parse_reservations(reservations) do
+          {:ok, parsed_reservations} ->
+            # Request reservations from Ordering
+            case Ordering.request_reservations(
```
agureev: Could you explain a bit what is happening here?

What I would assume is that you would delegate to Ordering to forward the read to the appropriate time once that has been assigned by the consensus. 

Instead we block here on getting the height. Why prefer this over the model with one less blocking call?

I may be just stupid here, so I would appreciate elaboration.

apps/anoma_node/lib/node/transaction/backends.ex
```diff
@@ -143,52 +145,196 @@ defmodule Anoma.Node.Transaction.Backends do
...
+                      # Try to store the result using shard logic
+                      case shard_storage_logic(
+                             node_id,
+                             id,
+                             vm_res,
+                             height,
+                             shard_pids
+                           ) do
+                        {:ok, backend_res} ->
+                          Ordering.transaction_completed(
```
agureev: Why do we need to make this a separate function? Looks like it can be an event instead.

But I guess it will become clearer once we properly integrating this with the RMs.

apps/anoma_node/lib/node/transaction/backends.ex
```diff
@@ -208,6 +354,101 @@ defmodule Anoma.Node.Transaction.Backends do
...
+    do_parse_reservations(reservations, [])
+  end
+
+  # Helper for parsing reservation lists recursively
+  @spec do_parse_reservations(Noun.t(), [{:read | :write, binary()}]) ::
+          {:ok, [{:read | :write, binary()}]} | {:error, atom()}
+  defp do_parse_reservations(noun, acc) do
+    case noun do
+      # --- Standard list terminators ---
+      0 ->
```
agureev: We have several more elegant ways of doing this casing such as `is_noun_zero` which can be also used as a guard and `Noun.equal?`

agureev (commented):
Thanks for all your work! Left some questions

apps/anoma_node/lib/examples/e_shard_backend.ex
AHartNtkn: As it turns out, the sleep calls in e_shard and e_shard_supervisor were doing nothing, so I got rid of them. Some of the ones in e_shard_backend.ex were effectively waiting for a Mempool.BlockEvent. I replaced those with an `assert_receive(... %Mempool.BlockEvent` call. Is that fine?

apps/anoma_node/lib/examples/e_shard.ex
AHartNtkn: Yes.

apps/anoma_node/lib/node/transaction/shard_supervisor.ex
AHartNtkn: Either you'd have a supervisor that acts like a genserver (or tries to), or you'd have a genserver that acts like a supervisor (or tries to). The impression I got from the conversation we had last week was that it wouldn't be possible to get a supervisor to do basic things since it's not a genserver, so we needed the genserver under it.

apps/anoma_node/lib/examples/e_shard.ex
agureev: Would be nice to test it some way, e.g. await a response in some try environment expecting it to fail.

apps/anoma_node/lib/node/transaction/shard_supervisor.ex
agureev: Right, yes, but what we talked about during the architecture discussion is that Elixir supervisors cannot do much other than supervise. That is, the supervisor you launch via `Supervisor.init(all_children, strategy: :one_for_one)`. Th GenServer launching the supervisor can still do... well, anything a GenServer can.

So you indeed use ShardSupervisor to launch the appropriate Elixir supervisor. However, I still a bit confused why cannot ShardRouter do that?

Is there some specific shutdown strategy you are considering here which I am not seeing?

apps/anoma_node/lib/node/transaction/shard.ex
AHartNtkn: All shard transactions have to have a specific height, regardless of if they only read. Read-only transactions are assigned a height, and they read from that height. There isn't a distinction made in the specs.

apps/anoma_node/lib/node/transaction/shard_supervisor.ex
agureev: Right, my bad, I was reading the code incorrectly. Indeed, this is fine.

apps/anoma_node/lib/node/transaction/backends.ex
AHartNtkn: > Ideally we would like for shards to be something that replaces current Storage, is that correct?

Who are you asking? If you're asking me, I don't know, because I'm not an owner of the project and am not privy to the long-term desires for it.

> What is blocking currently that sort of intergration?

As far as I can tell, the only outstanding issue is that Mempool calls storage directly when calling handle_execution_event. If that can be replaced with a call to a shard, then I think you would be able to delete Storage.

apps/anoma_node/lib/node/transaction/shard.ex
agureev: But who assigns them this height? The user, right?

If so, what is the best way a user can know the timestamps at which they should query the system?

apps/anoma_node/lib/node/transaction/shard_router.ex
AHartNtkn: I changed this as part of a multihoming refactor for this table.

apps/anoma_node/lib/examples/e_shard_backend.ex
AHartNtkn: For what? Logging? Getting state? Are you suggesting a modification to the Shard processes to publish their state as events through the EventBroker and have another process subscribe?

apps/anoma_node/lib/node/transaction/shard.ex
AHartNtkn: I wasn't, but I can work on that if you'd like.

apps/anoma_node/lib/node/transaction/backends.ex
AHartNtkn: Yes, I think this could be turned into an event. The event system is still strange to me.

apps/anoma_node/lib/node/transaction/backends.ex
AHartNtkn: Hmm... I'll think about it.

apps/anoma_node/lib/node/transaction/backends.ex
AHartNtkn: They have now been replaced with events.

apps/anoma_node/lib/node/transaction/shard_router.ex
AHartNtkn: Yes, that does make sense. Ordering is the only thing that calls ShardRouter.

apps/anoma_node/lib/node/transaction/shard.ex
AHartNtkn: In specs, the user submits a transaction to the mempool worker, and the worker assigns a height.

apps/anoma_node/lib/node/transaction/shard.ex
agureev: It need not be in this PR, but generally we should be backing the writes similarly to how we do with current storage. So something for future improvement.

apps/anoma_node/lib/node/transaction/shard.ex
agureev: Interesting. It makes sense. 

The mempool then probably just grabs some latest timestamp of that specific key available at the particular shard by querying the Shard Router. @juped do you think this makes sense?

apps/anoma_node/lib/node/transaction/backends.ex
agureev: Right, regarding the final result: I would say we indeed should aim for it to replace the current Storage module.

The idea I think is something like this for current setup I think: spawn on startup shards

1. A shard for Transparent Resource Machine (keys of form `/anoma/transparent/root`, `/anoma/transparent/nullifiers`) etc
2. A shard for Cairo Resource Machine (keys of form `/anoma/cairo/root`, ...)
3. probably blobs?

> As far as I can tell, the only outstanding issue is that Mempool calls storage directly when calling handle_execution_event. If that can be replaced with a call to a shard, then I think you would be able to delete Storage.

Right, so I think we should aim towards that. 

The way to do it I think is to:

1) Setup mnesia/rocksdb backing of shard key-values and 
2) provide exposition of keys at Mempool level

does that make sense? part 2 I am working on right this moment, should be done by the end of the week

apps/anoma_node/lib/examples/e_shard_backend.ex
agureev: No, no, I just meant that the debug messages could instead be placed through the eventbroker. This isn't important though, we can keep it this way, it is merely aesthetic choice.

apps/anoma_node/lib/node/transaction/backends.ex
AHartNtkn: Okay, after thinking it over the weekend, I've come to the conclusion that this block cannot be removed; at least not completely. A lack of height will always block an execution, one way or another, either during scrying, or at the end when writes need to happen. However, what you pointed out is blocking earlier than needed. I refactored things so that reservations are requested asynchronously, and that only becomes blocking when a scry or end write needs it. So it's still blocking, but it blocks as late as possible. This was kind of a nightmare to implement, but I found integrating the functionality of shard_router into ordering (thus deleting shard_router) made it simpler.

agureev: So right, I absolutely agree that the block cannot be removed completely. 

But I do think it may be worth moving the block to "as late as possible", i.e. during scry for one simple reason that if we have a transaction function which does not scry, then it should not worry about its height at all until it needs to write somewhere or read for RM reasons.

e.g. I submit a transaction candidate that does no scrying but e.g. does a lot of minor computations which were made to e.g. compress the transaction like jamming and such. then the beneficial thing would be for it to just do the computations before it gets ordered

this way we may save quite some time on block creation

anyway, I don't think this is a must, but it is worth considering

apps/anoma_node/lib/node/transaction/backends.ex
mariari: Yeah, this is a big issue for this work being truly integrated

apps/anoma_lib/lib/examples/enock.ex
```diff
@@ -2765,4 +2765,155 @@ defmodule Examples.ENock do
...
+      Nock.nock(stage2_code, [10, [6, 1 | dummy_tx_id], 0 | 1], env_tx1)
+
+    # Step 2: Apply formula [9, 2, 0 | 1] - this executes the transaction
+    {:ok, result1} = Nock.nock(ordered_tx1, [9, 2, 0 | 1], env_tx1)
+
+    # This is the expected format that backends.ex receives from vm_execute_stage2
+    assert result1 == expected_writes
+
+    :ok
+  end
```
mariari: Shouldn't we return the `%Nock{}` datastructure.

In fact you can use this setup for further examples that build upon this

apps/anoma_lib/lib/examples/enock.ex
```diff
@@ -2765,4 +2765,155 @@ defmodule Examples.ENock do
...
+
+  @doc """
+  I test a Nock program for the sequence (read a, read b, add a and b, write result to c).
+
+  I verify stage 1 (reservation/code split) and stage 2 (execution with mock scry).
+  """
+  @spec test_nock_program_tx3() :: :ok
+  def test_nock_program_tx3() do
```
mariari, line 2879: Ditto here, we should not return `:ok` when we have actually interesting data we can return, a user should be able to take this and play with it to develop further examples or learn the API or how the internals of the structure work

apps/anoma_node/lib/examples/e_shard.ex
```diff
@@ -0,0 +1,868 @@
...
+    # Test key "a"
+    assert Shard.read(shard_via, "a", 0) == {:ok, 5}
+
+    # Test key "b"
+    assert Shard.read(shard_via, "b", 5) == :absent
+
+    # Test key "c"
+    assert Shard.read(shard_via, "c", 4) == {:ok, 15}
+
+    :sys.get_state(shard_pid)
```
mariari: just return the shard_pid. Is the shard not the interesting data? You can use the pid itself for further things.

Is there even an API for working on the state? Isn't this an internal detail of the system that the user can't do anything with?

apps/anoma_node/lib/examples/e_shard.ex
```diff
@@ -0,0 +1,868 @@
...
+  end
+
+  @doc """
+  I test a scenario where a read is requested before the watermark allows,
+  then the watermark advances, and the read completes.
+  """
+  @spec test_queued_read() :: Shard.t()
+  def test_queued_read() do
```
mariari: Ditto here

apps/anoma_node/lib/examples/e_shard.ex
```diff
@@ -0,0 +1,868 @@
+defmodule Anoma.Node.Examples.EShard do
+  @moduledoc """
+  I contain examples on how to interact with the Shard module.
+  """
+
+  alias Anoma.Node.Registry
+  alias Anoma.Node.Transaction.Shard
+
+  import ExUnit.Assertions
+
```
mariari, line 11: Could we not have an example here that bootstraps the rest... you are always using the same setup and then use it in all your examples?

```elixir
  @spec initial_unread_shard_ac() :: {String.t(), pid()}
  def initial_unread_shard_ac do
    node_id = Anoma.Node.Examples.ENode.random_node_id()
    shard_id = :test_shard_1
    shard_via = Registry.via(node_id, Shard, shard_id)

    initial_kv = %{
      "a" => 5,
      "c" => 15
    }

    # Start the shard
    {:ok, shard_pid} =
      Shard.start_link(node_id: node_id, id: shard_id, initial_kv: initial_kv)

    {node_id shard_pid}
  end
```

I still don't know why we make a node_id here rather than having an actual node we spawn and do things that way, but this can remove the setup logic for most of your examples, if you aren't happy add extra shard with the public API. This should save a lot of repetitive examples

apps/anoma_node/lib/examples/e_shard.ex
```diff
@@ -0,0 +1,868 @@
+defmodule Anoma.Node.Examples.EShard do
+  @moduledoc """
+  I contain examples on how to interact with the Shard module.
+  """
+
```
mariari, line 5: In general we can remove the word `test_` from all names, we quality import things, test is a redundant word here

apps/anoma_node/lib/examples/e_shard_backend.ex
```diff
@@ -0,0 +1,718 @@
...
+  require Logger
+
+  import ExUnit.Assertions
+
+  @dialyzer :no_improper_lists
+
+  # --- Helper Functions ---
+
+  # Helper to wait for BlockEvent
+  defp wait_for_block(node_id, expected_round) do
```
mariari: Please add a type signature to this

apps/anoma_node/lib/examples/e_shard_backend.ex
```diff
@@ -0,0 +1,718 @@
...
+
+        if predicate_fun.(state) do
+          :ok
+        else
+          Process.sleep(check_interval)
+          # Recurse
+          fun.(fun)
+        end
+      end
+    end
```
mariari: We should not use `:sys.get_state`. This should be used very specifically and this isn't a valid use case.

Further can we remove the `Process.sleep` if you are waiting for something, why not subscribe to an event that you emit to do this. We should not sleep here

apps/anoma_node/lib/examples/e_shard_backend.ex
```diff
@@ -0,0 +1,718 @@
...
+    assert is_pid(pid_a),
+           "Shard 'a' PID not found. ShardSupervisor might not have started it correctly."
+
+    assert is_pid(pid_b), "Shard 'b' PID not found."
+    assert is_pid(pid_c), "Shard 'c' PID not found."
+
+    # Check final state using :sys.get_state - Tx1 (h=1), Tx2 (h=2), Tx3 (h=3)
+    state_a = :sys.get_state(pid_a)
+    state_b = :sys.get_state(pid_b)
+    state_c = :sys.get_state(pid_c)
```
mariari: We are abusing `get_state` we should not use it this much unless it's for meta operations, and I don't believe this counts, can we check this another way?

apps/anoma_node/lib/examples/e_shard_backend.ex
```diff
@@ -0,0 +1,718 @@
...
+    assert state_b.watermarks["b"].read == 3,
+           "Shard 'b' read watermark should be 3 after concurrent execution"
+
+    # Only Tx3 wrote to 'c'
+    assert state_c.watermarks["c"].write == 3,
+           "Shard 'c' write watermark should be 3 after concurrent execution"
+
+    # 6. Cleanup
+    :ok = ENode.stop_node(enode)
+    :ok
```
mariari: Why are we cleaning up? Why not let the node live as an argument to the example, see transaction examples that have the optional argument to shut down.

Basically we wish to carry on with this in further examples that you or others can write

apps/anoma_node/lib/node/transaction/ordering.ex
```diff
@@ -55,7 +64,7 @@ defmodule Anoma.Node.Transaction.Ordering do
     I am the type of the Ordering Enigine.
 
     I contain the Node for which the Ordering is launched, the upcoming
-    hight as well as a map from transaction IDs to their global order.
+    height as well as a map from transaction IDs to their global order.
```
mariari: This should be a different PR. Please don't fix stuff in the same PR this leads to horrid conflicts that we must avoid

apps/anoma_node/lib/node/transaction/ordering.ex
```diff
@@ -444,4 +1150,358 @@ defmodule Anoma.Node.Transaction.Ordering do
...
+          tx_id,
+          fn ->
+            blocking_shard_operation(
+              type,
+              state.node_id,
+              tx_id,
+              key,
+              value,
+              from
+            )
```
mariari: Does the formatter really do this?

apps/anoma_node/lib/node/transaction/ordering.ex
```diff
@@ -444,4 +1150,358 @@ defmodule Anoma.Node.Transaction.Ordering do
...
+         type,
+         tx_id,
+         key,
+         # nil for reads
+         value,
+         from,
+         shard_pid,
+         height,
+         node_id
+       ) do
```
mariari: This seems like a lot of arguments, maybe we can cut it down by some structs perhaps? Can those natrually occur?

apps/anoma_node/lib/node/transaction/shard.ex
```diff
@@ -0,0 +1,823 @@
...
+        essential_below_watermark =
+          find_essential_heights_below(read_watermark, key_height_map)
+
+        # 3. Determine essential heights to keep below each active read reservation
+        essential_below_reservations =
+          read_reservation_heights
+          |> Enum.map(&find_essential_heights_below(&1, key_height_map))
+          |> Enum.reduce(MapSet.new(), fn set1, set2 ->
+            MapSet.union(set1, set2)
+          end)
```
mariari:
```suggestion
          |> Enum.reduce(MapSet.new(), &MapSet.union/2)
```

apps/anoma_node/lib/node/transaction/ordering.ex
```diff
@@ -357,20 +579,483 @@ defmodule Anoma.Node.Transaction.Ordering do
...
+                updated_key_wms
+              )
+        }
+      else
+        # No change
+        state_after_update
+      end
+
+    # Return final state for this key
+    state_after_update
```
mariari: Maybe we can flip the case, so the shorter case comes first, so it's easier to see what block relates to what, I have a hard time intuitively seeing the relation when the shorter case comes second

apps/anoma_node/lib/node/transaction/ordering.ex
```diff
@@ -357,20 +579,483 @@ defmodule Anoma.Node.Transaction.Ordering do
...
+              Map.put(
+                state_after_update.watermark_state,
+                key,
+                updated_key_wms
+              )
+        }
+      else
+        # No change
+        state_after_update
+      end
```
mariari: ditto here.... wait isn't this the same logic... this should be abstracted, this is all too manual

apps/anoma_node/lib/node/transaction/ordering.ex
```diff
@@ -357,20 +579,483 @@ defmodule Anoma.Node.Transaction.Ordering do
...
+  end
+
+  # Recalculates watermarks for a single key and sends updates to shards if they advance.
+  @spec recalculate_and_update_watermarks_for_key(binary(), t()) :: t()
+  defp recalculate_and_update_watermarks_for_key(key, state) do
+    current_wms =
+      Map.get(state.watermark_state, key, %{
+        read_watermark: -1,
+        write_watermark: -1
+      })
```
mariari: The default read_watermark and write_watermark leaks everywhere this should be properly abstracted and not leak as what if we add other watermarks? Do we have to chase all of these?

mariari (changes requested):
So far I've only done a surface reading.

The biggest conceptual issues I see are as follows:

1. Examples aren't being used to the fullest, a lot of room is had to cut down the amount of code and make more useful examples that grow on each other
2. A lot of code is sorta repeated through the modules I've read, it's hard to pinpoint exactly what, but abstraction can be used to better encapsulate all the code. It's hard to read a sea of code
3. The git history isn't useful, I gave up trying to read it commit by commit as many parts are rewritten within the same PR meaning that the history will be annoying to merge and reason about.
4. The code isn't integrated, this is the biggest blocker for me wanting to merge this as shards aren't a different backend, it completely replaces our storage system. Bolting on more code to the codebase seems questionable and we should find ways to cut a lot of code not add a lot of code in this process.

apps/anoma_node/lib/node/transaction/shard_supervisor.ex
```diff
@@ -0,0 +1,256 @@
...
+          Supervisor.init([], strategy: :one_for_one)
+
+        {:error, _reason} ->
+          {:stop, :ets_table_creation_failed}
+      end
+    rescue
+      e ->
+        # Reraise the exception to ensure supervisor termination/restart according to strategy
+        reraise e, __STACKTRACE__
+    end
```
mariari: What's the point of the `try do`?

Just run the code if it crashes it'll already cause the supervisor above this to fail

apps/anoma_node/lib/node/transaction/shard_supervisor.ex
```diff
@@ -0,0 +1,256 @@
...
+        {:ok, _ets_tid} ->
+          # Simply initialize the supervisor with the child specs
+          Supervisor.init(shard_child_specs, strategy: :one_for_one)
+
+        :no_shards ->
+          # No shards configured or generated, start no children
+          Supervisor.init([], strategy: :one_for_one)
+
+        {:error, _reason} ->
+          {:stop, :ets_table_creation_failed}
```
mariari: What does this stop do that just crashing in `create_and_populate_ets_table` doesn't do?

I'm unsure of the semantic difference between `:stop` and `:shutdown` from a supervisor @m1dnight any thoughts?

apps/anoma_node/lib/node/transaction/shard_supervisor.ex
```diff
@@ -0,0 +1,256 @@
...
+            id: shard_id,
+            initial_kv: %{}
+          ]
+
+          child_spec = %{
+            id: shard_id,
+            start: {Shard, :start_link, [shard_args]}
+          }
+
+          {[child_spec | specs_acc], Map.put(map_acc, key, shard_id)}
```
mariari: These two cases are the same, abstract it.

If you find yourself repeating yourself it's a good sign that it can be done better with some minimal abstraction ontop

apps/anoma_node/lib/node/transaction/shard_supervisor.ex
```diff
@@ -0,0 +1,256 @@
...
+  (`{:via, Registry, {Anoma.Node, {Shard, key}}}`) of the `Shard` process
+  responsible for that key. The actual lookup of keys is handled by the
+  `Anoma.Node.Transaction.Ordering`.
+
+  ### Key Concepts
+
+  - **Supervisor Args:** Keyword list including `:strategy` and `:schema`.
+  - **Strategy:** Determines how shards are created (e.g., `:one_per_key`).
+  - **Schema:** Defines the initial keys and their starting values.
+  - **ETS Table:** `:shard_key_map` for key -> shard name lookup (used by Anoma.Node.Transaction.Ordering).
```
mariari: Why are we using ETS tables, and not using our `Anoma.Node.Tables` module? We need support for rocksdb, as ETS tables have massive flaw and we should be operate where we can turn them automatically into rocksdb tables with mnesia as the interface

apps/anoma_node/lib/node/transaction/shard_supervisor.ex
```diff
@@ -0,0 +1,256 @@
...
+        # Reraise the exception to ensure supervisor termination/restart according to strategy
+        reraise e, __STACKTRACE__
+    end
+  end
+
+  ############################################################
+  #                    Private Helpers                       #
+  ############################################################
+
+  # Processes the schema based on the strategy to generate shard child specs
```
mariari: I believe most of these can go away. The supervisor module shouldn't have this much code, most of it seems to do with tables which should be moved to the tables module and use mnesia anyways

apps/anoma_node/lib/node/transaction/ordering.ex
AHartNtkn: Yes.

apps/anoma_node/lib/examples/e_shard.ex
AHartNtkn: All of those examples were written to test shards before they were actually spawned by nodes.

mariari:
This work can be found at Attic/shard-integration.

The work will instead be merged as #2060, in which I've rerolled the entire PR

PR #2004, opened 2025-04-07
TRM Byte Array Nock Representation

agureev:
Make relevant TRM data byte-array representable on Nock level.

Namely nullifier key commitment, nullifier key, and nonce.


agureev:
@lukaszcz once Cairo integration with Node is done we can do the same for the CairoRM Nock representations.

PR #2038, opened 2025-04-25
Config files

m1dnight:
This PR adds config files for dev/test/prod. Jeremy noticed that the ports being used under test were the same as the ones in dev. I took the opportunity to create the proper config scaffolding now. 

The idea is that configs under dev/test are generated at compiletime, and not at runtime. This makes sense because you run the dev/test in the repo itself, which causes a compile cycle every time anyway.


mariari:
Needs to be rerolled

PR #2040, opened 2025-04-28
Ordering State Requests

agureev:
Previously Ordering launched blocking tasks on each request that was
subscribed to the OrderingEvent

Now, it simply stores the requests and launches them if stacked upon
order retrieval.

Currently expects each transaction worker to have a unique unanswered
request at a time.


agureev:
Closed in favor of in-shard integration

PR #2049, opened 2025-05-06
Add Kudos application as an example to the client

m1dnight:
This adds the Kudos examples from anoma-apps in the client as examples.


mariari:
Closing

PR #2050, opened 2025-05-06
BugFix: Cairo Transactions Not Assumed to Have Hash Fields

agureev:
This assumption after field removal made transactions fail inappropriately.


XuyangSong (approved):
Thank you for fixing it

PR #2052, opened 2025-05-07
JSON encoding for events

m1dnight:
Add JSON encoding for all events.


mariari:
This topic has one evil merge when testing it against v0.34.0. Note vm_error in backends/events for myself

PR #2055, opened 2025-05-07
Remove lost and found `IO.puts`

m1dnight:
There was a lost and now found IO.puts that I had to get rid of.


m1dnight:
wrong branch.

PR #2058, opened 2025-05-07
Add a `wrap` flag to the mempool to wrap transactions in a transaction candidate before submitting

m1dnight:
Solvers in particular don't know how to create a transaction candidate from a transaction binary without having a full nock infrastructure available. 

I updated the mempool api to pass in a flag "wrap" that will wrap the transaction in a transaction candidate before submitting it.


agureev (approved):
lgtm 

maybe just add a test to see that the wrapper actually evaluates correctly? had to test it by hand it make sure

PR #2060, opened 2025-05-07
Shard Integration

agureev:
Integrate shards as a Storage replacement


mariari:
I still need to split out all the commits and fix a bug.

Namely some writes in the backends may fail, but the entire TX doesn't roll back. This is serious enough where I care

See

#2179 

Also for something I won't fix before putting it in a RC:

#2178

PR #2061, opened 2025-05-08
Replace Client with the LocalDomain

mariari:


mariari:
Saved under Attic/mariari/local-domain for prosperity.

We should remember the functionality that we lost here to make it whole again

PR #2064, opened 2025-05-09
Accept integers in Cairo workflow

lukaszcz:
Some fields in the JSONs accepted only hexadecimal string representations of numbers. They should also accept JSON integers.


mariari:
I'll have to rebase this PR on where it really belongs, as in the current shape it's not mergable

mariari:
I've rerolled it

PR #2069, opened 2025-05-14
docs: replace broken HexDocs links with GitHub sources for `compile_protoc` and `anoma_lib`

rnkrtt:
Hi, I’ve updated the documentation by swapping the outdated HexDocs URLs for `compile_protoc` and `anoma_lib` with direct links to their GitHub directories. This eliminates 404s and lets contributors reach the source code instantly—no functional changes, docs only.


agureev:
Hi! The documentation is correct in an of itself: when the docs will be published, they will be availiable through hex. They currently are not.

It would be semantically incorrect to put a github link there.

Issue #1511, opened 2024-10-31
Expose the client CLI from the Anoma release binary


murisi:
customer:
performer:
deadline:
estimated:
started:
actual:
completed:
confirmed:
dependencies:
- [ ]
supersedes:
-
---

Currently the Anoma release binary only exposes subcommands like `start` and `remote` for starting and connecting to Anoma nodes. It is desired for the Anoma release binary to somehow expose and handle the client subcommands defined in the `anoma_client` escript. So the end result should be a release binary with node operation subcommands as well as client subcommands that trigger operations on nodes.

PR #2070, opened 2025-05-16
Murisi/release anoma client rebased

murisi:
An attempt to address #1511 . Added an Anoma client executable script to the release archive that works by using the Erlang emulator to directly execute BEAM files in the release. From the release directory root, this script can be executed with `./bin/anomac --listen-port 4001 --node-host localhost --node-port 8181` for instance. This approach to releasing the Anoma client was chosen for the following reasons:
* The BEAM files for each Elixir module are not duplicated. If the binary produced `mix do --app anoma_client escript.build` was used instead, then BEAM files would be duplicated since the output of `escript.build` is essentially an archive of BEAM files headed by a shebang line.
* The source for the Anoma client does not have to be copied into the release archive. Doing this would be undesirable because (1) the logic that is in the BEAM files is still indirectly duplicated, and (2) it is probably more efficient to run the Anoma client from compiled artifacts than the source code directly.
* The `anomac` command invokes `erl` through the `anoma` executable script produced by `mix release` in order to avoid hardcoding the path to the `erl` binary (which might change between `mix` releases) and to ensure that the Anoma client is run with the same settings/configuration and the Anoma node.
* The `anomac` command boots into `releases/0.25.0/start` instead of `releases/0.25.0/start_clean` in order to ensure that `Anoma.Client.Application` starts the supervisor `Anoma.Client.ConnectionSupervisor` before the client CLI script starts. This is necessary because the client CLI script depends upon `Anoma.Client.ConnectionSupervisor`.

### DIFF
--- a/.github/workflows/build_release.yaml
+++ b/.github/workflows/build_release.yaml
@@ -33,8 +33,7 @@ jobs:
       - uses: actions/checkout@v4
         with:
           fetch-depth: 50
-      - uses: erlef/setup-beam@v1.18.2
-        if: startsWith(runner.os, 'Linux') || startsWith(runner.os, 'Windows')
+      - uses: erlef/setup-beam@6e1d5a5aa63e46e2579c0a01718a1376fb24354f
         with:
           otp-version: ${{ inputs.otp-version }}
           elixir-version: ${{ inputs.elixir-version }}
@@ -45,7 +44,7 @@ jobs:
       - name: Install system dependencies
         shell: bash
         if: startsWith(runner.os, 'macOS')
-        run: brew install elixir protobuf
+        run: brew install protobuf
       - name: Install system dependencies
         shell: pwsh
         if: startsWith(runner.os, 'Windows')
--- a/.github/workflows/build_release.yaml
+++ b/.github/workflows/build_release.yaml
@@ -63,6 +63,10 @@ jobs:
         run: |
           export PATH="$PATH:$HOME/.mix/escripts"
           mix release
+      - name: Copy shared libraries
+        if: startsWith(runner.os, 'macOS')
+        shell: bash
+        run: ./scripts/copy-dylibs.zsh _build/dev/rel/anoma
       - name: Build Anoma release
         if: startsWith(runner.os, 'Windows')
         shell: pwsh
--- a/.github/workflows/build_release.yaml
+++ b/.github/workflows/build_release.yaml
@@ -66,7 +66,9 @@ jobs:
       - name: Copy shared libraries
         if: startsWith(runner.os, 'macOS')
         shell: bash
-        run: ./scripts/copy-dylibs.zsh _build/dev/rel/anoma
+        run: |
+           ./scripts/copy-dylibs.zsh _build/dev/rel/anoma
+           cp ./scripts/anomac _build/dev/rel/anoma/bin/
       - name: Copy shared libraries and client
         if: startsWith(runner.os, 'Linux')
         shell: bash
--- a/.github/workflows/build_release.yaml
+++ b/.github/workflows/build_release.yaml
@@ -89,14 +89,18 @@ jobs:
         shell: pwsh
         run: |
           cd _build/dev/rel/
-          tar czf anoma.tar.gz anoma
-          shasum -a 1 anoma.tar.gz > anoma.tar.gz.sha1sum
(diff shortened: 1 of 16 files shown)


murisi:
See an example of the release generated by the CI here: https://github.com/murisi/anoma/releases/tag/v41 . The linked binaries seem to be working, but more testing on different platforms and OSs would be appreciated. This is what I have tested so far:
* `anoma-v41-Linux-X64.tar.gz`:
  * On a clean x86_64 Linux machine, download the archive, extract it
  * Running `./anoma/bin/anoma start` or `./anoma/bin/anoma start_iex` should both work
  * Running the examples `Anoma.Node.Examples.EShieldedTransaction.complex_cairo_transaction()` and `Examples.ECrypto.blood_l_signed()` should work. Which ones don't work?
  * Starting a node with `node = Anoma.Node.Examples.ENode.start_node()` and then a client with `Anoma.Client.Examples.EClient.create_example_client()` should work.
    * I've only tested the `/intents` query at `http://localhost:4000/spec/swaggerui . It works returning the empty list.
    * I've tested starting another client using a variation of `ELIXIR_ERL_OPTIONS="-sname client@ec2-instance" CLIENT_HTTP_PORT=4001 CLIENT_GRPC_PORT=40052 NODE_GRPC_PORT=50052 ./anoma/bin/anomac --node-host localhost --node-port 50051 --node-id "79FAAA27633F8EF4316ADC7DDE80E5F3307E52A5C111EF0E785F8DEC16E554C3"`
      * `sname` and `NODE_GRPC_PORT` are currently required because I have not yet worked on preventing the GenSevers from being started when the `anomac` is invoked
* `anoma-v41-macOS-X64.tar.gz`: Same invocation instructions as `anoma-v41-Linux-X64.tar.gz`, but now on a x86_64 Mac OS system. Passes the same tests as `anoma-v41-Linux-X64.tar.gz`
* `anoma-v41-Windows-X64.tar.gz`:
  * DLL dependencies currently have to be manually copied into `erts-15.2.3/bin/` for `bin/anoma` to work and not display NIF errors
  * An `bin/anomac` (like the one for Linux and Mac OS) script has not yet been made.
  * On a clean x86_64 Windows machine, download the archive, extract it
  * Running `./anoma/bin/anoma start` or `./anoma/bin/anoma start_iex` should both work
  * Running the examples `Anoma.Node.Examples.EShieldedTransaction.complex_cairo_transaction()` and `Examples.ECrypto.blood_l_signed()` should work. Which ones don't work?
  * Starting a node with `node = Anoma.Node.Examples.ENode.start_node()` and then a client with `Anoma.Client.Examples.EClient.create_example_client()` should work.
    * I've only tested the `/intents` query at `http://localhost:4000/spec/swaggerui . It works returning the empty list.

murisi:
> DLL dependencies currently have to be manually copied into erts-15.2.3/bin/ for bin/anoma to work and not display NIF errors

The CI now automatically finds and copies Anoma for Windows' DLL dependencies into the release archive. Therefore DLL dependencies no longer have to be manually copied into erts-15.2.3/bin/ for `bin/anoma.bat` to work and not display NIF errors. So on a recent Windows 64-bit machine (without developer tools installed), it should now be sufficient to just download the Anoma archive and run the `bin/anoma.bat` binary.

Additionally, the Docker build also works. To run it follow the steps below:
* Run `docker pull ghcr.io/murisi/anoma:v48` to download the Anoma image
* Running `docker run -it --network host 92676abaea4c start` or `docker run -it --network host 92676abaea4c start_iex` should both work
* Running the examples `Anoma.Node.Examples.EShieldedTransaction.complex_cairo_transaction()` and `Examples.ECrypto.blood_l_signed()` should work. Which ones don't work?
* Starting a node with `node = Anoma.Node.Examples.ENode.start_node()` and then a client with `Anoma.Client.Examples.EClient.create_example_client()` should work.
  * I've only tested the /intents query at `http://localhost:4000/spec/swaggerui . It works returning the empty list.

@juped @mariari @m1dnight Please do let me know whether or not you are able to run the generated binaries on your machines. Please also let me know if using the binaries is confusing or is not in line with how Anoma is supposed/envisioned to be run. Thanks!

The above binary and Docker image were built in this CI job: https://github.com/murisi/anoma/actions/runs/15753634641 . I am now trying to see if the CI steps can be demonstrated in the https://github.com/anoma/anoma-ci-test repo, but am running up against some environment protections for `github-pages` during the `publish-docs` step...

PR #2102, opened 2025-07-04
Ray/naive sparse rebased

murisi:
Opened up a PR for the proposed Sparse Merkle Tree implementation at https://github.com/anoma/anoma/tree/ray/naive-sparse to enable reviewing. Since the branch is based on an empty worktree, this PR targets an empty branch so that all the additions can be seen in the PR. Should the PR be approved, it would have to be rebased onto https://github.com/anoma/anoma/tree/base .


lib/sparse_merkle_tree/ets.ex
```diff
@@ -0,0 +1,255 @@
...
+
+  @doc """
+  I hash a leaf value and insert it into a sparse merkle tree, returning
+  an updated tree.
+  """
+  @spec insert(t(), binary()) :: t()
+  def insert(tree, leaf) do
+    digest = hash(leaf)
+
+    :ok = put_digest(tree.table, digest)
```
murisi, line 88: Does the Merkle tree in this PR support inserting multiple leaves before rehashing? Probably this functionality would be useful when a Merkle tree is not being built incrementally, and is instead being constructed from a list of digests. Also, this use case is supported in https://github.com/anoma/anoma/pull/1655 through the `rehash` and `put` functions.

lib/examples/e_sparse_merkle_tree/generic.ex
```diff
@@ -0,0 +1,166 @@
...
+
+      def dont_prove_abc_absent_in_abc_def() do
+        :error = prove_absent(abc_def_tree(), "abc")
+      end
+
+      def dont_prove_abc_absent_in_abc() do
+        :error = prove_absent(abc_tree(), "abc")
+      end
+
+      def big_tree(exponent \\ 8) do
```
murisi, line 129: This implementation seems to have lower performance than https://github.com/anoma/anoma/pull/1655 . Specifically, running `big_tree(16)` takes 68s (native) and 46s (ETS) here in contrast to https://github.com/anoma/anoma/pull/1655 (after bug fix) where it takes 14s. In https://github.com/anoma/anoma/tree/ray/sparse-merkle-tree the bottleneck seemed to be in `Map.put` (as the leaf count gets higher). Could there be a similar issue here?

lib/sparse_merkle_tree/ets.ex
```diff
@@ -0,0 +1,255 @@
...
+
+  @spec prove(t(), binary(), (digest_map(), bitstring() -> Hash.t())) ::
+          {:ok, Proof.t()} | :error
+  defp prove(tree, leaf, fun) do
+    # todo: prove more efficiently without copying
+    temp_table = :ets.new(__MODULE__, [])
+    :ets.insert(temp_table, :ets.tab2list(tree.table))
+
+    {proof, <<>>} =
+      for _ <- 256..0//-1, reduce: {[], hash(leaf)} do
```
murisi, line 157: Can this design be extended to support Sparse Merkle Tree depths less than 256 bits? I guess such an extension would be desirable for the Plasma architecture since we need to support both 256-bit and 64-bit Sparse Merkle trees, right? Also, would it be possible to do this extension in a way that has hash conflict detection (so that leaf proving and verification fail if the leaf's digest is off by even 1 bit)?

lib/sparse_merkle_tree/ets.ex
```diff
@@ -0,0 +1,255 @@
...
+
+  @derive {Inspect, only: [:table]}
+  typedstruct enforce: true do
+    @typedoc """
+    I am the struct representing a naive SHA-256 sparse merkle tree.
+
+    I am mutable, operations update the underlying ETS table.
+    """
+
+    field(:table, :ets.tid())
```
murisi, line 34: Is the usage of ETS tables necessary given that https://github.com/anoma/anoma/pull/1655 is performant without them? This question is rather aesthetic, but it seems that https://github.com/anoma/anoma/pull/1655 uses strictly less Elixir/Erlang features in its implementation and its performance seems to be no worse than this one.

lib/sparse_merkle_tree/ets.ex
```diff
@@ -0,0 +1,255 @@
...
+
+    :ok
+  end
+
+  @spec compute_digest(:ets.tid(), Hash.t()) :: Hash.t()
+  defp compute_digest(_table, _bits = <<_::256>>) do
+    @present_hash
+  end
+
+  @spec compute_digest(:ets.tid(), bitstring()) :: Hash.t()
```
murisi, line 224: Just a very minor note: `compute_digest` could probably get away with calling `get_digest` once. This could be achieved if we were to supply it with the hash of one of its children computed from the previous call to `compute_digest`. But probably doing this wouldn't have a noticeable performance impact since this is probably not the bottleneck in the first place. Also see https://github.com/anoma/anoma/blob/4e526d421ed4febb4861c74cd63e6c8f98a548b4/apps/anoma_lib/lib/sparse_merkle_tree.ex#L160 where experiments along this line were carried out.

lib/sparse_merkle_tree/naive.ex
```diff
@@ -0,0 +1,255 @@
...
+
+  @derive {Inspect, only: [:root]}
+  typedstruct enforce: true do
+    @typedoc """
+    I am the struct representing a naive SHA-256 sparse merkle tree.
+
+    I am immutable, all operations return a new value.
+    """
+
+    field(:leaves, MapSet.t(Hash.t()), default: MapSet.new())
```
murisi, line 35: In the naive Sparse Merkle Tree implementation, the `leaves` and/or the `root` fields might be redundant. I.e. maybe the presence of a leaf could be checked by seeing whether the path is present in the `digests` `Map`? Also, maybe the root hash could be obtained by getting the digest corresponding to the empty path in `digests`? See https://github.com/anoma/anoma/blob/4e526d421ed4febb4861c74cd63e6c8f98a548b4/apps/anoma_lib/lib/sparse_merkle_tree.ex#L59-L68 for an attempt at doing this.

lib/examples/e_sparse_merkle_tree/generic.ex
```diff
@@ -0,0 +1,166 @@
...
+    implementation = Keyword.fetch!(opts, :implementation)
+
+    quote do
+      import SparseMerkleTree
+      import SparseMerkleTree.Proof
+
+      def empty_tree() do
+        tree = unquote(implementation).new()
+
+        <<40, 126, 172, 2, 225, 203, 200, 220, 129, 46, 78, 65, 80, 109,
```
murisi, line 12: Would it be possible to replace the constant bitstrings in `generic.ex` with the formulas used to generate them? Doing this might make it easier to check whether the test is correct. Also, would it be possible to add some tests checking that a Merkle tree generated by the library fully matches a handmade `Map` whose entries are calls to `hash(... <> ...)`? See https://github.com/anoma/anoma/blob/4e526d421ed4febb4861c74cd63e6c8f98a548b4/apps/anoma_lib/lib/examples/e_sparse_merkle_tree.ex#L247-L255 for some tests constructed along these lines.

murisi (commented):
Thanks for working on this @juped . I haven't seen any bugs yet, but it seems there are two general issues that might need to be addressed:
* Both Merkle tree implementations seem to be noticeably less performant than https://github.com/anoma/anoma/pull/1655 (after the two bug fixes) (at least for the operations that I've tried).
* Some useful features like non-256 bit tree depths and batch leaf insertion/delayed tree hashing (found in https://github.com/anoma/anoma/pull/1655 ) are missing here.

lib/sparse_merkle_tree/ets.ex
juped: This could be useful in the interface, yeah.

lib/examples/e_sparse_merkle_tree/generic.ex
juped: That implementation is algorithmically better, no contest; it's more or less the standard algorithm described in https://eprint.iacr.org/2016/683.pdf (I think, it's been a while since I reviewed it). Naive isn't very good (that's why "naive"), it's just the baseline results other implementations have to match. My design is an interface and test cases, the implementations are just placeholders (that give correct results inefficiently).

lib/sparse_merkle_tree/ets.ex
juped: Depth equaling hash length (because the hash is the path) is more or less what makes it a sparse merkle tree; the usual merkle tree is non-sparse and history-dependent (e.g. we can still use the poseidon CommitmentTree from elsewhere as a commitment tree). So if you want it to be smaller, you replace your hash with, say, truncated sha256 or something. This interface is idempotent on double inserts, though perhaps it should instead fail (because ultimately, we want to avoid double-insertion into a nullifier set).

lib/sparse_merkle_tree/ets.ex
juped: Some implementation strategies want to operate in-place; using ETS is a rough approximation (though it's still generically map-backed and you won't get the algorithmic benefits). The main benefit of a mutable backing store is not copying large terms between processes, just copying the reference over.

lib/sparse_merkle_tree/naive.ex
juped: Yes, that's true. (It's basically for debug printing to the shell.)

lib/examples/e_sparse_merkle_tree/generic.ex
juped: It makes sense to e.g. call default_digests for the correct empty tree root; the equivalent of hash(hash("f" <> "f") <> hash("t" <> "f")) up the tree could be a test helper, perhaps.

lib/sparse_merkle_tree/ets.ex
murisi: Thanks for explaining @juped ! The sparse Merkle tree implementation from Eiger seems to support tree heights less than (or equal to) the hash length. (So if the hash length is `l` and the tree height is `d`, then a proof produced by Eiger's implementation will be a list of length less than (or equal to) `d` where each element is a hash of exactly length `l`.) See https://github.com/anoma/anoma/pull/2099/files#diff-e10f531a529a2eebee8899481efe6a33d2060388a5391b1b04309a3959e7f269R65 . It also seems to be sparse, history independent, and idempotent on double inserts, but I'm not sure and would appreciate your opinion here. Would you be able to review the implementation at https://github.com/anoma/anoma/pull/2099 please?

mariari:
Good bookmark

PR #2106, opened 2025-07-10
Update documentation, small fix to live docs

l4e21:
A [live doc is broken](https://anoma.github.io/anoma/writing-documents.html), also I had an issue I ran into while running wx and realised this needed (but not necessarily anoma specific) package could be useful to specify in the readme.


tuntunkhan990:
wow

PR #2129, opened 2025-08-06
Removed Unintended Toggle

dinitheth:
# Fix: Removed Unintended Toggle in Solver’s `:disable` Call

## Summary

- Removed unintended toggle behavior in the solver’s `:disable` call.
- The function now **consistently sets the solver state to disabled**, rather than re-enabling it on repeated calls.

---
### Improvements & Recommendations
- Add an enable/1 API and write comprehensive tests for both enable and disable behaviors.
- Introduce automated integration tests for solver events and state transitions.
- Consider documenting the solver’s configuration options in user-facing guides or README to improve developer experience.

### DIFF
--- a/apps/anoma_node/lib/node/intents/solver.ex
+++ b/apps/anoma_node/lib/node/intents/solver.ex
@@ -111,7 +111,7 @@ defmodule Anoma.Node.Intents.Solver do
   end
 
   def handle_call(:disable, _from, state) do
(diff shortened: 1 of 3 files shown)


mariari:
Odd, why are you merging trash pr's into this work which is unrelated. I'd understand if it was related to your topic in any way

PR #2133, opened 2025-08-08
Fix: Handle fallback submit case, normalize `valid?/1`, and improve intent safety

Nepali123:
This PR fixes minor issues in Anoma.Node.Intents.Solver:

Adds logging and :ok return in fallback submit/2 clause to avoid silent failures.
Normalizes valid?/1 to return {:error, :invalid} instead of false.
Adds a 5-second timeout in submit/2 receive to prevent blocking.
Wraps Intent.nullifiers/1 and Intent.commitments/1 results in MapSet.new/1 for safety in unsolved_reject/2.

cc @mariari @agureev

### DIFF
--- a/apps/anoma_node/lib/node/intents/solver.ex
+++ b/apps/anoma_node/lib/node/intents/solver.ex
@@ -130,10 +130,6 @@ defmodule Anoma.Node.Intents.Solver do
     state
   end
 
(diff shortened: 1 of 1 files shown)


mariari:
Predicates return booleans, all the other changes are odd

PR #2134, opened 2025-08-08
Fix stop_node/1 to safely lookup and stop node process by PID

Nepali123:
This patch corrects a critical bug in the Anoma.Supervisor.stop_node/1 function.

Previously, stop_node/1 incorrectly passed a :via tuple directly to Supervisor.stop/1, which expects a PID. This would cause runtime errors.
The fix involves first resolving the registered process PID via Anoma.Node.Registry.whereis/3.
If no process is found, it returns {:error, :not_found}.
Otherwise, it safely calls Supervisor.stop/1 with the PID to stop the node supervisor process.
This change ensures safe and reliable stopping of node supervisors.

cc @mariari @agureev

### DIFF
--- a/apps/anoma_node/lib/supervisor.ex
+++ b/apps/anoma_node/lib/supervisor.ex
@@ -73,11 +73,15 @@ defmodule Anoma.Supervisor do
   @doc """
   Given a node id, I stop that node completely.
   """
-  @spec stop_node(String.t()) :: :ok
+  @spec stop_node(String.t()) :: :ok | {:error, :not_found}
   def stop_node(node_id) do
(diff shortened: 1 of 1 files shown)


mariari:
You are incorrect

```elixir
iex(anoma@YU-NO)23> node = Anoma.Node.Examples.ENode.start_node
%Anoma.Node.Examples.ENode{
  pid: #PID<0.1036.0>,
  node_id: "B5826C71D7A53AA8883C4A6AD6E730ACCA61286DC40B6FF60C893111773F049F"
}
iex(anoma@YU-NO)24> Anoma.Supervisor.stop_node(node.node_id)
:ok

```

Also none of this is critical

PR #2135, opened 2025-08-08
Fix typo in handle_info return reason (:disonnect → :disconnect)

Nepali123:
This PR fixes a typo in the handle_info/2 callback of Anoma.Client.Node.GRPCProxy.
The atom :disonnect was not a meaningful or intentional value, and should be :disconnect.

cc @mariari @agureev


mariari:
The reason given doesn't particularly matter to the supervisor who shuts it down since it's not one of the answers that affect restarts. I'll let @m1dnight decide what to do with this particular value

m1dnight:
It's a typo, so a fix is never a bad idea. Thanks for this!

Nepali123:
@m1dnight The failing `ReplayTest` seems unrelated to this typo fix. All other checks passed, could we force merge this?

mariari:
Read our git processes:

https://anoma.github.io/anoma/git.html

and wait for a release

PR #2145, opened 2025-08-26
fix: secure binary deserialization and handle malformed input

Nepali123:
This PR makes `:binary_to_term` usage in PubSub and IntraNode safe by enabling `[:safe]` and adding error handling for invalid binaries. This prevents crashes and security risks from malformed inputs without changing functionality.


Nepali123:
CC @mariari @agureev @m1dnight

m1dnight:
This is not necessary. If the publish fails, it fails. We leave the try/catch for the Java people.

PR #2185, opened 2026-02-10
Simplify Shard.Supervisor: flatten control flow, deduplicate helpers

mariari:
Simplification of the shard supervisor code, written entirely by Claude.

### DIFF
--- a/apps/anoma_node/lib/node/transaction/shard/supervisor.ex
+++ b/apps/anoma_node/lib/node/transaction/shard/supervisor.ex
@@ -3,31 +3,20 @@ defmodule Anoma.Node.Transaction.Shard.Supervisor do
   I am the dynamic supervisor for `Anoma.Node.Transaction.Shard` processes.
 
   I start with no children. Children (shards) are started dynamically either
-  during initial node setup based on a provided schema or later via the
-  `start_shard/3` function. I maintain a Mnesia table
-  (`Anoma.Tables.table_shard_key_map/1`) mapping keys to the registered name
-  (`{:via, Registry, {Anoma.Node, {Shard, key}}}`) of the `Shard` process
-  responsible for that key. The actual lookup of keys is handled by the
-  `Anoma.Node.Transaction.Ordering`.
+  during initial node setup based on a provided schema or later via
+  `start_shard/3`.
 
-  ### Key Concepts
-
-  - **Supervisor Args:** Keyword list including `:node_id`, and optionally `:strategy` and `:schema` for *initial* setup.
-  - **Strategy:** Determines how initial shards are created (e.g., `:one_per_key`). Only used during initial setup.
-  - **Schema:** Defines the initial keys and their starting values. Only used during initial setup.
-  - **Mnesia Table:** `table_shard_key_map/1` for key -> shard name lookup (used by Anoma.Node.Transaction.Ordering).
-  - **Dynamic Starting:** Use `start_shard/3` to add new shards after initial setup.
+  I record key-to-shard mappings in Mnesia
+  (`Anoma.Tables.table_shard_key_map/1`) for future replay support.
+  These mappings are not currently read in production.
 
   ### Public API
 
   - `start_link/1`: I start the supervisor.
-  - `start_shard/3`: I dynamically start a new shard for a given key and optionally value.
-  - `get_shard_key_map/1`: I return the name of the Mnesia table for the shard key map for the given node ID.
+  - `start_shard/3`: I dynamically start a new shard for a given key.
   """
 
   use DynamicSupervisor
-  use EventBroker.DefFilter
-  use TypedStruct
 
   alias Anoma.Node.Registry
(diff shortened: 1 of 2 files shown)


apps/anoma_node/lib/node/transaction/shard/supervisor.ex
```diff
@@ -39,199 +28,62 @@ defmodule Anoma.Node.Transaction.Shard.Supervisor do
...
-        ) ::
+  @spec start_shard(String.t(), key(), initial_value() | nil) ::
           DynamicSupervisor.on_start_child()
-          | {:error, :mnesia_update_failed, any()}
   def start_shard(node_id, key, initial_value \\ nil) do
-    supervisor_name = Registry.via(node_id, __MODULE__)
-    shard_id = Enum.reduce(key, "", &(&2 <> &1)) |> String.to_atom()
+    shard_id = shard_id(key)
 
-    # Check if already registered before attempting start
```
agureev, line 194: Why remove this comment?

apps/anoma_node/lib/node/transaction/shard/supervisor.ex
```diff
@@ -240,144 +92,71 @@ defmodule Anoma.Node.Transaction.Shard.Supervisor do
...
-    mnesia_tx = fn ->
-      :mnesia.write({table_name, key, shard_id})
-    end
-
-    case :mnesia.transaction(mnesia_tx) do
-      {:atomic, :ok} ->
-        :ok
+  # Best-effort Mnesia write for future replay support.
+  defp record_shard_mapping(node_id, key, shard_id) do
+    table = Tables.table_shard_key_map(node_id)
```
agureev, line 164: Maybe we can record it as some top-level var? Cheap but I wonder if we can get rid of this recompute given that the table is constant.

apps/anoma_node/lib/node/transaction/supervisor.ex
```diff
@@ -20,7 +20,7 @@ defmodule Anoma.Node.Transaction.Supervisor do
           node_id: String.t(),
           mempool: Mempool.args_t(),
           ordering: Ordering.args_t(),
-          shards: Shard.Supervisor.args_t() | nil
+          shards: list(Shard.Supervisor.startup_options()) | nil
```
agureev, line 23: I wonder if we should instead prefer to have a map in the future rather than a keyword-list with possibly repeating elements

apps/anoma_node/lib/node/transaction/shard/supervisor.ex
```diff
@@ -240,144 +92,71 @@ defmodule Anoma.Node.Transaction.Shard.Supervisor do
...
-  # Cases where strategy/schema are nil or invalid for initial setup
-  defp process_schema(_node_id, _strategy, _schema) do
-    {[], %{}}
+    case {strategy, schema} do
+      {:one_per_key, entries} when is_list(entries) ->
+        Enum.flat_map(entries, fn
+          {key, value} when is_list(key) ->
+            shard_id = shard_id(key)
+            record_shard_mapping(node_id, key, shard_id)
+            [child_spec(node_id, shard_id, %{key => value})]
```
agureev: This can be abstracted I think

Something to do in another PR maybe

apps/anoma_node/lib/node/transaction/shard/supervisor.ex
```diff
@@ -240,144 +92,71 @@ defmodule Anoma.Node.Transaction.Shard.Supervisor do
...
-         initial_kv
-       ) do
-    shard_id = Enum.reduce(key, "", &(&2 <> &1)) |> String.to_atom()
-
-    shard_args = [
-      node_id: node_id,
-      id: shard_id,
-      initial_kv: initial_kv
-    ]
+  defp shard_id(key) do
```
agureev: Does not respect the function order of the style guide

apps/anoma_node/lib/node/transaction/shard/supervisor.ex
```diff
@@ -240,144 +92,66 @@ defmodule Anoma.Node.Transaction.Shard.Supervisor do
...
```
agureev: Specs needed

agureev (changes requested):
Looks good!

Couple of comments:

1) Specs needed
2) In-code comments would be welcome
3) Function order per the style guide needed

apps/anoma_node/lib/node/transaction/shard/supervisor.ex
mariari: I assume it was overzealous, I saw it do it in another topic and I told it to cut it out. I think my newest rules I put in should remove this behaviour in the future

apps/anoma_node/lib/node/transaction/supervisor.ex
mariari: these are the arguments as this is how startup options are passed normally IIRC. I think it's fine per elixir standard. I could be wrong though

